PR #6, opened 2021-07-01
Add tests for API extraction logic

shadaj:
Also introduces GitHub Actions configuration


shadaj:
It seems that new actions don't run in PRs, so we'll find out if the configuration is correct after merging.

Manishearth (approved):

Issue #163, opened 2022-06-03
`Method::params_held_by_output` doesn't catch all cases


QnnOkabayashi:
Bug introduced in #159 

Currently, `Method::params_held_by_output` uses a naive strategy for finding params by traversing the type tree and seeing if any lifetimes share the same name as something in the output. This introduces a bug where if a method introduces lifetime `'b: 'a`, and a param uses `'b` and the output uses `'a` and borrows from the param by lifetime coercion, it won't be detected because it has a different name. As an example, consider the following:
```rust
#[diplomat::opaque]
pub struct Wrapper<'a>(&'a MyOpaqueType);

impl<'a> Wrapper<'a> {
    pub fn new<'b: 'a>(inner: &'b MyOpaqueType) -> Box<Wrapper<'a>> {
        Box::new(Self(inner))
    }
}
```
Here, the implementation thinks that `&'b MyOpaqueType` doesn't share any lifetimes with `Box<Wrapper<'a>>` because we don't consider the fact that `'b: 'a`, so the method would return an empty vec which is incorrect.

This could be really bad because in languages like JS, we depend on this information being correct for holding lifetime guards, so a faulty implementation could cause use-after-free bugs.

There are two approaches we can take:
1. Disallow bounds on lifetimes (easy)
2. Somehow build a lifetime resolution system by using the `LifetimeDef` types that we hold in the ast (hard)

PR #164, opened 2022-06-03
JS lifetime guards

### LINKED ISSUE #12: Handle lifetime tracking in GC targets

QnnOkabayashi:
This PR fixes #12 by adding lifetime guards to the params that the output is lifetime bound to.

This PR won't by itself allow lifetime magic yet. Other critical steps include resolving #163, and addressing #145 which would involve a lot of validity checks that don't exist yet to make this safe.


QnnOkabayashi:
I'm pretty sure this isn't blocked by #145 so this is ready for review.

PR #202, opened 2022-07-07
Private lifetime edges

### LINKED ISSUE #182: JS private fields for lifetime guards
When JS has an opaque struct with a lifetime, it stores the value that it borrows from as a field like `this.__x_lifetime_guard`. We should consider some form of private stashing so these fields can't be deleted by downstream users, potentially causing the GC to free an object and invalidate a lifetime bound.

I think the best way to go about this is to have a [private list field](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields#examples), and just push edges to the list.

QnnOkabayashi:
Fixes #182

### DIFF
--- a/example/js/api.mjs
+++ b/example/js/api.mjs
@@ -9,15 +9,17 @@ const ICU4XDataProvider_box_destroy_registry = new FinalizationRegistry(underlyi
 });
 
 export class ICU4XDataProvider {
-  constructor(underlying, owned) {
+  #lifetimeEdges = [];
+  constructor(underlying, owned, edges) {
     this.underlying = underlying;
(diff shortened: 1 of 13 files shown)


QnnOkabayashi:
One of the "gotchas" here is that non-opaque structs, even if they contain lifetimes, do not store lifetime edges. Instead, they attach them to the fields that use those lifetimes. This also means that if you have a `&'a Opaque` as a field, where `Opaque` doesn't have an lifetimes itself, it will still have to store an edge to the owner of the `'a` lifetime. This means that regardless of having any borrows, all opaque types must have a `#lifetimeEdges` private field.

Manishearth (approved):

PR #207, opened 2022-07-13
Improve TSDoc

QnnOkabayashi:
This PR brings a couple small fixes building off of #206.

* Added `@throws` doc comments to methods that return `Result`.

* Added numeric primitive aliases to clarify APIs. These don't help typeck other than specifying `number` (or `string` for chars), but they do clarify intent for downstream users. An alternative would be to add a doc comment saying `@param x - Rust u8`.

* Use TypeScript string enums to allow for per-variant documentation.

* Redesigned the import resolution code to have an `Imports` type that just figured out all the imports at once for a particular type's `.mjs` file and `.d.ts` file.

* Added `tsconfig.tsbuildinfo` to .gitignore


tool/src/js/structs.rs
```diff
@@ -478,7 +502,13 @@ pub fn gen_ts_type<W: fmt::Write>(
             ast::PrimitiveType::bool => write!(out, "Uint8Array")?,
             ast::PrimitiveType::char => write!(out, "Uint32Array")?,
         },
-        ast::TypeName::Unit => out.write_str("void")?,
+        ast::TypeName::Unit => {
+            if is_return_type {
+                out.write_str("void")?;
+            } else {
+                out.write_str("{}")?;
```
sffc, line 509: Question: where does this `{}` get used?

sffc (approved):

tool/src/js/structs.rs
QnnOkabayashi: Methods that return `Result<T, ()>` use this because when we write the `@throws` thing, I wanted to specify the content of the `FFIError`. This lead to it generating `@throws FFIError<void>`, while the JS code is actually throwing `FFIError<{}>`. Should I change this to a more explicit enum, or at least change the name to `is_ffierror_value` and invert the usage?

sffc: You should be able to set `error_value` to `undefined` and then treat it as an `FFIError<void>`

PR #214, opened 2022-07-25
Add `diplomat.config.js`

QnnOkabayashi:
Add support for a `diplomat.config.js`, which allows for library authors to specify the path to their `.wasm` file and an initialization function that will get run right after the `wasm` object is instantiated. I've also updated the documentation in `docs/npm_packaging.md`.

Here's an example:
```js
function init(wasm) {
    wasm.some_logger_init();
}

export default {
    wasm_path: new URL('./api/diplomat_example.wasm', import.meta.url),
    init,
};
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,13 +59,13 @@ jobs:
       with:
         node-version: 14.17.0
     - name: Build feature tests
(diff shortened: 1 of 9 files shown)


Manishearth (approved):

sffc (commented):
lgtm

Issue #218, opened 2022-08-07
`usize` is missing in TypeScript


sffc:
I got the following compile error when running `tsc` in the ICU4X lib directory

```
../icu4x/lib/ICU4XBidiInfo.d.ts:16:22 - error TS2304: Cannot find name 'usize'.

16   paragraph_count(): usize;
                        ~~~~~
```

I worked around it by adding `usize` to diplomat-runtime.d.ts. Is it missing? @QnnOkabayashi

QnnOkabayashi:
Fixed in https://github.com/rust-diplomat/diplomat/pull/219

PR #219, opened 2022-08-08
Add `usize` and `isize` aliases to `runtime.d.ts`

QnnOkabayashi:
Fixes the issue where functions signatures in JS/TS files would use `usize` and `isize`, but would import the `u32` and `i32` type aliases. Now we have type aliases `usize` and `isize` for number, and those are imported instead.


sffc (commented):
LGTM

QnnOkabayashi:
Fixes https://github.com/rust-diplomat/diplomat/issues/218

Issue #226, opened 2022-08-23
Impl `serde` for HIR types


QnnOkabayashi:
This is the first step for adding snapshot testing.

PR #253, opened 2022-09-03
HIR snapshots

QnnOkabayashi:
Small PR to add the first iteration of snapshots, which are just debug snapshots. Eventually we'll want to do this better with serde and converting the IDs into type names, but this is good for now for catching bugs in the implementation.

For example, I caught a copy-pasta error that didn't allow out structs to contain owned opaques. We should write a lot more tests, but it seems from my super naive testing that lifetime elision works :)

### DIFF
--- a/core/src/ast/modules.rs
+++ b/core/src/ast/modules.rs
@@ -43,7 +43,7 @@ impl Module {
         rust_links
     }
 
-    fn insert_all_types(&self, in_path: Path, out: &mut Env) {
+    pub fn insert_all_types(&self, in_path: Path, out: &mut Env) {
         let mut mod_symbols = ModuleEnv::default();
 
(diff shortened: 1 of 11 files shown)


QnnOkabayashi:
I had to hack around in the `Env` stuff which I think is really not good, open to suggestions on how to make the `do_test!` less evil

Manishearth (approved):
Looks good. Would eventually probably want nicer output displays (even just some manual Debug impls on the leaf types like TypeLifetime would be pretty great), but this is fine for now.

QnnOkabayashi:
Yup, the goal of this PR was to _have something_, and it's already proven useful. Next steps are to write more tests and eventually do serde: https://github.com/rust-diplomat/diplomat/issues/226.

PR #258, opened 2022-09-17
Replace `structopt` with `clap@3`.

wtgtybhertgeghgtwtg:
Running [`cargo audit`](https://crates.io/crates/cargo-audit) on a project depending on this repository gives a warning that the transitive dependency `ansi_term` [is unmaintained ](https://rustsec.org/advisories/RUSTSEC-2021-0139).  This comes from the `structopt` dependency by way of `clap@2`.  However, `structopt` [is in maintenance mode](https://github.com/TeXitoi/structopt/tree/v0.3.26#maintenance) since its main feature is included in `clap@3`, so that dependency on `clap@2` will not be updated.  As it is suggested by `structopt` to move to `clap@3`, and `clap@3` replaced the `ansi_term` dependency, I am filing this PR to do so.

Please let me know if I have missed anything, or if this change is unwanted.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -27,15 +27,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ansi_term"
-version = "0.11.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ee49baf6cb617b853aa8d93bf420db2383fab46d314482ca2803b40d5fde979b"
-dependencies = [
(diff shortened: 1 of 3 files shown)


Manishearth (approved):
This is fine, but also that's weird, nobody should be depending on the diplomat_tool crate

wtgtybhertgeghgtwtg:
Thank you for merging.

Issue #225, opened 2022-08-23
Solving variance with `Opaque` and `OpaqueMut`


QnnOkabayashi:
# Problem
With support for lifetimes, Diplomat is now vulnerable to dangling references. This can be caused when we modify the contents of an opaque that we have a mutable reference to, since we have no way of ensuring that something else in FFI land isn't holding a reference into it. The classic example of this is holding a reference into a `Vec`, and then clearing it in another function call, resulting in a dangling reference.

# Solution
When working with FFI, we don't have access to the full scope that types are passed around in. However, we do have full access to the declarations of types. Since only opaques can be passed by reference, we can mark them either as "mutable" or "immutable" in their declarations. Something like `#[diplomat::opaque]` vs `#[diplomat::opaque_mut]`?

Here are the key differences I imagine:
1. Only mutable opaques can be passed as mutable references in arguments, but both can be passed immutably.
2. You can only "return" borrows from immutable opaques, where "return" means either in the return position or in "out" parameters.
    * This also means that when a mutable opaque is passed to a method as a ref (cannot be passed by value), regardless of whether the reference is mutable or immutable, the lifetime of the reference and any lifetimes that the opaque is generic over _must not_ appear anywhere else in the method signature. This ensures that nothing borrows from it. This would also mean we would have to check that no lifetimes can coerce into them. The easiest solution would be to enforce anonymous lifetimes, and check that any output lifetime elision isn't tied to it. I'm working on elision inference right now and this shouldn't be too difficult to add later.

### Pros
* This seems like the "Rusty" solution, and allows for maximum flexibility.
### Cons
* Many changes to the HIR, particularly splitting opaques into `Opaque` and `OpaqueMut`.

# Alternatives
Just ban mutability altogether. The downside of this is that updating non-node based data structures becomes impractical, and using node based data structures in Rust is hell.

Manishearth:
Oh, this is a neat idea. I like it!

Manishearth:
Actually, would we have to split Opaque and OpaqueMut as types? We could start off with a bool that's checked and later splti types if we really need (I'm wary of having a giant pile of HIR types that are almost the same but not quite)

Manishearth:
An annoying thing about those rules is that iterators that return opaque references won't work. That seems useful to have.

QnnOkabayashi:
> An annoying thing about those rules is that iterators that return opaque references won't work. That seems useful to have.

[...]

PR #263, opened 2022-09-27
Attr parsing for `#[diplomat::opaque_mut]`

QnnOkabayashi:
This is the first PR towards https://github.com/rust-diplomat/diplomat/issues/225.

The first step towards adding opaque_mut is supporting parsing it from the AST. The way we parsed our custom attributes from the AST was a bit of a mess, so this PR improves that mechanism, and then uses the improved version to also support parsing `#[diplomat::opaque_mut]`. It then uses this information to set a new flag field in `ast::OpaqueStruct` when constructed, which dictates whether or not it's an opaque_mut.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -238,7 +238,7 @@ dependencies = [
 
 [[package]]
 name = "diplomat"
-version = "0.3.0"
+version = "0.4.0"
 dependencies = [
  "diplomat_core",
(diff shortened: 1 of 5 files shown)


Manishearth (approved):
praise: quite straightforward!

(didn't realize we already had output_only!)

PR #357, opened 2023-11-07
Dart: contiguous enums

robertbastian:


Manishearth (approved):
seems fine though this might be cleaned up further by unconditionally generating a `toFFI()`/`fromFFI()` and calling it everywhere else

robertbastian:
The `asFfi`/`fromFfi` expressions are quite short, and sometimes unused so I think inlining them is better.

PR #379, opened 2023-11-29
String fixes

robertbastian:
Using `char16_t`/`u16string_view`, and adding UB warning to inline C++ docs.


hsivonen:
Thanks. Looks good to me as a non-owner.

sffc (commented):
Praise: `char16_t` is the way to go. Don't touch `wchar` with a 10-foot pole.

PR #384, opened 2023-12-01
Fix JS slice borrowing

robertbastian:
#383

### DIFF
--- a/tool/src/js/conversions.rs
+++ b/tool/src/js/conversions.rs
@@ -541,7 +541,8 @@ impl SliceKind {
             SliceKind::Str16 => write!(f, "diplomatRuntime.readString16(wasm, {ptr}, {size})"),
             SliceKind::Primitive(prim) => match prim {
                 JsPrimitive::Number(num) => {
-                    write!(f, "new {num}Array(wasm.memory.buffer, ptr, size)")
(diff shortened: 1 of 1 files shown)


tool/src/js/conversions.rs
```diff
@@ -541,7 +541,8 @@ impl SliceKind {
             SliceKind::Str16 => write!(f, "diplomatRuntime.readString16(wasm, {ptr}, {size})"),
             SliceKind::Primitive(prim) => match prim {
                 JsPrimitive::Number(num) => {
-                    write!(f, "new {num}Array(wasm.memory.buffer, ptr, size)")
+                    // TODO(#383): can we borrow this?
+                    write!(f, "{num}Array.from(new {num}Array(wasm.memory.buffer, ptr, size))")
```
sffc, line 545: Yeah this is confusing and not well documented, but according to the spec, when you create a TypedArray from an ArrayBuffer, you borrow the ArrayBuffer, but when you create it from any other collection, even another TypedArray of the same type, you allocate your own data.

https://tc39.es/ecma262/multipage/indexed-collections.html#sec-typedarray

sffc (approved):

Manishearth (approved):

PR #394, opened 2023-12-19
Fix typedef in result

### LINKED ISSUE #393: Typedef in `.h` shadows class in `.hpp`
In Google C++, each type's header needs to be explicitly included. This leads to an issue with Diplomat's current output. For example:

```cpp
#ifndef diplomat_result_box_ICU4XBidi_ICU4XError_H
#define diplomat_result_box_ICU4XBidi_ICU4XError_H
#include <stdio.h>
#include <stdint.h>
#include <stddef.h>
#include <stdbool.h>
#include "diplomat_runtime.h"

typedef struct ICU4XBidi ICU4XBidi;
#include "ICU4XError.h"
#ifdef __cplusplus
namespace capi {
extern "C" {
#endif
typedef struct diplomat_result_box_ICU4XBidi_ICU4XError {
    union {
        ICU4XBidi* ok;
        ICU4XError err;
    };
    bool is_ok;
} diplomat_result_box_ICU4XBidi_ICU4XError;
#ifdef __cplusplus
} // extern "C"
} // namespace capi
#endif
#endif
```

Note that the `typedef struct ICU4XBidi ICU4XBidi;` is defined outside the `namespace capi`. Wherever this file is included, there is some weird shadowing going on, and our linter insists on importing this file, which then breaks everything as the struct is not a class.

robertbastian:
Fixes #393

Including the `.h` instead of defining a typedef correctly puts the typedef into the `capi` namespace in C++.

### DIFF
--- a/example/c/include/diplomat_result_box_ICU4XFixedDecimalFormatter_void.h
+++ b/example/c/include/diplomat_result_box_ICU4XFixedDecimalFormatter_void.h
@@ -6,7 +6,7 @@
 #include <stdbool.h>
 #include "diplomat_runtime.h"
 
-typedef struct ICU4XFixedDecimalFormatter ICU4XFixedDecimalFormatter;
+#include "ICU4XFixedDecimalFormatter.h"
 #ifdef __cplusplus
 namespace capi {
 extern "C" {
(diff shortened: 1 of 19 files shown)


Manishearth (approved):
I thought we did this to avoid cyclic imports when methods need imports. But if this works, this works!

Manishearth (approved):

robertbastian:
The `gen` task is broken on main, looks like we cannot use `/tmp` anymore. Merging.

PR #400, opened 2024-01-05
Dart native assets

robertbastian:


Manishearth (approved):
I don't fully understand this but seems fine

PR #431, opened 2024-02-21
Improving lowering error types

robertbastian:
Lowering currently uses `Option`s for three different purposes, which is confusing:
* signaling failure
* unit return type
* absent lifetime env

I changed the first case to `Result<T, ()>`, and the last case to non-optional.


robertbastian:
The second commit replaces `Option<SuccessType>` by a new `SuccessType::Unit` variant, which simplifies logic and is less error prone.

sffc (approved):
Splitting None and Err seems good

PR #465, opened 2024-03-28
Use Clap 4.2

sffc:
If we don't need the features in 4.5, we should use 4.2 which is what we have in ICU4X. 4.5 is not compatible with MSRV


Manishearth (approved):

sffc:
The CI failure doesn't seem to be related; should I ignore it?

robertbastian:
It's fixed on main, you need to merge

PR #467, opened 2024-04-08
Add indexing special method and clean up Dart

robertbastian:
This is the Diplomat side of https://github.com/unicode-org/icu4x/pull/4783

### DIFF
--- a/tool/src/dart/mod.rs
+++ b/tool/src/dart/mod.rs
@@ -505,44 +505,34 @@ impl<'a, 'cx> TyGenContext<'a, 'cx> {
 
         let params = param_decls_dart.join(", ");
 
-        let declaration = if let Some(special) = &method.attrs.special_method {
-            match &special {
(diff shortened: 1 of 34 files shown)


tool/src/dart/mod.rs
```diff
@@ -544,37 +528,21 @@ impl<'a, 'cx> TyGenContext<'a, 'cx> {
...
                     format!("{return_ty} get iterator")
                 }
+                SpecialMethod::Indexer => {
+                    format!("{return_ty} operator []({params})")
+                }
                 _ => unimplemented!("Found unknown special method type {special:?}"),
             }
-        } else if method.param_self.is_none() {
-            // Static field
-            if params.is_empty()
```
robertbastian, line 551: I removed this case because I realised that in order to make something a static singleton, it needs to be immutable, and it's hard to figure that out.

Manishearth (approved):

PR #474, opened 2024-05-13
WASM.mjs template fix for Node/Browser detection

ambiguousname:
`fetch === 'undefined'` doesn't work for my setup when running ava in the ICU4X ffi/capi folder. fetch is defined for the Browser as well as Node.JS.

`typeof process` hopefully works better to detect if it's Node.JS instead of Browser modules? But let me know if this breaks anything for anyone else!

### DIFF
--- a/tool/src/js/wasm.mjs
+++ b/tool/src/js/wasm.mjs
@@ -26,7 +26,7 @@ const imports = {
   }
 }
 
-if (typeof fetch === 'undefined') { // Node
+if (typeof process !== 'undefined') { // Node
   const fs = await import("fs");
   const wasmFile = new Uint8Array(fs.readFileSync(cfg['wasm_path']));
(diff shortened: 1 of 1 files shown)


robertbastian (commented):
What is your setup? Iirc you need an older version of node for this code. I'd like to understand exactly which node versions support which checks before merging anything. We should also update the node version used in CI.

sffc:
So in theory `fetch` should work for both filesystem and http URLs, but in practice it doesn't seem to work very well for filesystem URLs. Reverting back to using the `fs` module in Node for loading the WASM seems like a good idea.

I'm okay dropping support for older Nodes. 18 or 20 can be our min version.

ambiguousname:
This fix is included in #487, so I'm closing this in favor of those broader changes.

PR #493, opened 2024-06-09
Fix and Test JS UTF-16 code

sffc:
This code must have never been run before by anyone, because it simply didn't work. This impacts `segment_utf16` and any other functions that take utf16. I added a test.

### DIFF
--- a/feature_tests/js/package.json
+++ b/feature_tests/js/package.json
@@ -19,6 +19,7 @@
       "esm"
     ],
     "nodeArguments": [
+      "--experimental-wasm-bigint"
     ]
   },
   "scripts": {
--- a/feature_tests/js/package.json
+++ b/feature_tests/js/package.json
@@ -19,7 +19,6 @@
       "esm"
     ],
     "nodeArguments": [
-      "--experimental-wasm-bigint"
     ]
   },
   "scripts": {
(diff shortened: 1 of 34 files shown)


feature_tests/js/package.json
```diff
@@ -19,7 +19,6 @@
       "esm"
     ],
     "nodeArguments": [
-      "--experimental-wasm-bigint"
```
sffc, line 22: I just noticed that this change made it into this PR; it is needed for Node 14 compatibility. I can move it to another PR if desired

sffc:
The actual change is tool/src/js/runtime.mjs

Manishearth (approved):

feature_tests/src/structs.rs
```diff
@@ -51,13 +51,11 @@ pub mod ffi {
             Box::new(Opaque("".into()))
         }
 
-        #[diplomat::attr(supports = constructors, constructor)]
```
robertbastian: could you add a lowering check that enforces a single constructor?

feature_tests/js/package.json
robertbastian: Have we changed CI to node 16 already? It should be in that PR

sffc: Okay I removed this change from this PR so we don't have to figure it out here

feature_tests/src/structs.rs
sffc: Out of scope, I think

robertbastian: then file an issue please

robertbastian (approved):

PR #497, opened 2024-06-13
kotlin-benchmark - basic benchmark

jcrist1:
This is a first attempt to address #496. I adapted the existing `example` test for kotlin into a benchmark using `jmh`. For comparison I implemented a test with criterion for the example package. On my machine I get 

### kotlin (jmh)
```
Benchmark                                       Mode  Cnt     Score      Error  Units
ICU4XFixedDecimalFormatterBench.benchDecimal    avgt    4  1031,388 ±  180,639  ns/op
ICU4XFixedDecimalFormatterBench.benchFormat     avgt    4  4526,571 ±   56,495  ns/op
ICU4XFixedDecimalFormatterBench.benchFormatter  avgt    4  7386,616 ± 1595,937  ns/op
ICU4XFixedDecimalFormatterBench.benchLocale     avgt    4  2271,330 ±  379,326  ns/op
ICU4XFixedDecimalFormatterBench.benchOptions    avgt    4  1550,664 ± 1361,251  ns/op
ICU4XFixedDecimalFormatterBench.benchProvider   avgt    4  2409,987 ±  212,799  ns/op
```

### rust (criterion)
```
create locale           time:   [40.487 ns 40.679 ns 40.929 ns]
Found 3 outliers among 100 measurements (3.00%)
  3 (3.00%) high severe

create provider         time:   [282.74 ns 283.94 ns 285.33 ns]
Found 7 outliers among 100 measurements (7.00%)
  7 (7.00%) high mild

create options          time:   [1.1152 ns 1.1170 ns 1.1190 ns]
Found 18 outliers among 100 measurements (18.00%)
  4 (4.00%) low severe
  2 (2.00%) low mild
  8 (8.00%) high mild
  4 (4.00%) high severe

create decimal          time:   [23.563 ns 23.612 ns 23.665 ns]
Found 19 outliers among 100 measurements (19.00%)
  3 (3.00%) high mild
  16 (16.00%) high severe

create formatter        time:   [219.83 ns 220.05 ns 220.33 ns]
Found 17 outliers among 100 measurements (17.00%)
  7 (7.00%) high mild
  10 (10.00%) high severe

format                  time:   [60.580 ns 60.625 ns 60.676 ns]
Found 10 outliers among 100 measurements (10.00%)
  5 (5.00%) high mild
  5 (5.00%) high severe
```

excluding the options which are just a pure stack structure and hence super fast in rust which is about 1000x faster, most of the kotlin wrappers take 10-70x longer. I can easily benchmark other things as well. The `format` bench in rust is annoying cause I needed to expose the `write` create and destroy methods. Open to other ideas there. Maybe I should just do something that works with strings.


runtime/src/lib.rs
```diff
@@ -10,6 +10,7 @@ mod wasm_glue;
 
 mod write;
 pub use write::DiplomatWrite;
+pub use write::{diplomat_buffer_write_create, diplomat_buffer_write_destroy};
```
jcrist1, line 13: this feels wrong. I would propose just adding a method to the example lib that uses a String to write to.

jcrist1:
n.b. This is my first time using either `jmh` or `criterion`. So I might be missing some basic benchmarking best practices.

Manishearth (approved):
Wow, thank you so much for working on this!!

jcrist1:
Also I'm having a weird issue where the gradle command fails when I run it from a clean state, but succeeds if I run the command first in IntelliJ 🤔 .

jcrist1:
okay gradle seems to skip the bench if a result file already exists in `diplomat/example/kotlin/somelib/build/results/jmh/results.txt`

PR #498, opened 2024-06-14
Use process.getBuiltinModule when available; sniff by globalThis.process

sffc:
Latest iteration on https://github.com/rust-diplomat/diplomat/issues/283

### DIFF
--- a/tool/src/js/wasm.mjs
+++ b/tool/src/js/wasm.mjs
@@ -26,12 +26,20 @@ const imports = {
   }
 }
 
-if (typeof fetch === 'undefined') { // Node
-  const fs = await import("fs");
+if (globalThis.process?.getBuiltinModule) {
+  // Node (>=22)
+  const fs = globalThis.process.getBuiltinModule('fs');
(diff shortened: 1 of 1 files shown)


Manishearth (approved):

Manishearth (approved):

sffc:
I guess there's nothing holding up merging this

example/js/api/diplomat-wasm.mjs
```diff
@@ -26,12 +26,20 @@ const imports = {
   }
 }
 
-if (typeof process !== 'undefined') { // Node
-  const fs = await import("fs");
+if (globalThis.process?.getBuiltinModule) {
+  // Node (>=22)
+  const fs = globalThis.process.getBuiltinModule('fs');
```
robertbastian, line 31: is this more efficient than `await import('fs')`? This seems to be the only functional change

sffc: My understanding is it is the same as `import('fs')` except it is synchronous and only works for built-in modules.

PR #499, opened 2024-06-14
panama-backend - basic integration with jextract

jcrist1:
I've implemented a basic integration with jextract for a pure java backend. This is just a baseline to start building on, that generates the C2 backend, and uses that with jextract to create the java interface to the native code. There's no code generation yet for ergonomic interfaces. I feel like enums and structs will be relatively straightforward, so once again I will start off building the wrappers for opaque type and cleanup. Also I don't actually no any java yet 🤪, but no time like the present to learn.


tool/src/java/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+        std::fs::create_dir_all(parent)?;
+        let mut out_file = File::create(&out_path)?;
+        out_file.write_all(text.as_bytes())?;
+    }
+
+    let lib_path = tmp_path.join(format!("{LIBRARY}.h"));
+
+    let mut lib_file = File::create(&lib_path)?;
+    for include in include_files {
+        writeln!(lib_file, "#include \"{include}\"")?;
```
jcrist1, line 58: jextract suggests loading all headers into a single library file. This gets transformed into an enormous monster of java

tool/src/lib.rs
```diff
@@ -76,6 +78,30 @@ pub fn gen(
...
+            let mut attr_validator = hir::BasicAttributeValidator::new("true");
+            attr_validator.support.renaming = true;
+            attr_validator.support.disabling = true;
+            attr_validator.support.iterators = true;
+            attr_validator.support.iterables = true;
+            attr_validator.support.indexing = true;
+            attr_validator.support.constructors = true;
+            attr_validator.support.named_constructors = true;
+            attr_validator.support.memory_sharing = true;
+            attr_validator.support.accessors = true;
```
jcrist1, line 91: Just enabled everything for generating some code to start off with. Will adjust later

jcrist1:
I now manually implemented an opaque type with cleanup as well. This will be the model for the template. I also added a benchmark where I just instantiate the opaque struct because I was quite curious what the cost of GC was. With GC on the benchmark throughput more than halved. going from ~2.7M/s to 1.2 M/s. A slow down from about 350ns to ~800ns, which is just interesting.

tool/src/java/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+const JAVA_DIR: &str = "src/main/java/";
+
+pub fn run(
+    tcx: &TypeContext,
+    _conf_path: Option<&Path>,
+    out_folder: &Path,
+) -> std::io::Result<FileMap> {
+    let files = FileMap::default();
+    let mut context = c2::CContext::new(tcx, files, false);
+    context.run();
```
Manishearth, line 25: Note that after https://github.com/rust-diplomat/diplomat/pull/495, the C2 backend can now be directly invoked per-type in a cleaner fashion, if you decide that's more useful

feature_tests/java/somelib/src/main/java/dev/diplomattest/somelib/Opaque.java
```diff
@@ -0,0 +1,41 @@
...
+
+    public long pointer() {
+        return internal.address();
+    }
+
+    public void delete() {
+        somelib_h.Opaque_destroy(this.internal);
+    }
+
+    public void assertStruct(MyStruct struct) {
```
Manishearth, line 38: question: wait, jextract generated this from the C header files? How did it know that these C header files correspond to a class with methods?

Manishearth (commented):
this is neat but i'm kind of surprised it works??

feature_tests/java/somelib/src/main/java/dev/diplomattest/somelib/Opaque.java
jcrist1: Sadly no, that would be amazing 😃. That’s just a hand rolled wrapper for a more ergonomic type. I’m using this to build the templating for opaque types

jcrist1: Everything in the `ntv` package was generated using jextract from the c backend.

Manishearth: Aha! Okay, carry on!

feature_tests/java/somelib/src/main/java/dev/diplomattest/somelib/Opaque.java
```diff
@@ -0,0 +1,67 @@
...
+import dev.diplomattest.somelib.ntv.*;
+
+
+import java.lang.foreign.Arena;
+import java.lang.foreign.MemorySegment;
+import java.lang.ref.Cleaner;
+import java.nio.charset.StandardCharsets;
+
+
+public class Opaque {
```
jcrist1, line 12: Although the tool doesn't actually generate code, the class body has now been been copied from an insta snapshot.

tool/src/java/snapshots/diplomat_tool__java__test__opaque.snap
```diff
@@ -0,0 +1,69 @@
...
+package dev.diplomattest.somelib;
+
+import dev.diplomattest.somelib.ntv.*;
+
+
+import java.lang.foreign.MemorySegment;
+import java.lang.ref.Cleaner;
+
+
+public class Opaque {
```
jcrist1, line 14: Here is the snapshot from where we copied the body

tool/src/java/mod.rs
```diff
@@ -0,0 +1,963 @@
...
+                // public double[] asBoxedSLice() {
+                //     try (var arena = Arena.ofConfined()) {
+                //         var nativeVal = somelib_h.Float64Vec_as_boxed_slice(arena, internal);
+                //         var data = dev.diplomattest.somelib.ntv.DiplomatF64View.data(nativeVal);
+                //         var len = dev.diplomattest.somelib.ntv.DiplomatF64View.len(nativeVal);
+                //         var returnVal = data.asSlice(0, len * JAVA_DOUBLE.byteSize()).toArray(JAVA_DOUBLE);
+                //         Native.free(data.address());
+                //         return returnVal;
+                //     }
+                // }
```
jcrist1, line 334: Here's what a boxed slice would look like using jna. It should be the only place where we'd need to use it.

tool/src/java/mod.rs
jcrist1: I'd prefer to implement that as a followup.

jcrist1:
Okay @Manishearth I think I'm ready to open this up for review. There are still some issues that would be nice to have, but maybe aren't dealbreakers
* I think this could do with a lot more testing, but I'd kind of like to solicit some ideas. Maybe if someone has a concrete use case they can propose some additional test cases (@brainstorm you said you were interested in the Panama backend do you have any examples you'd like to specifically see?). 
* I'm also a little embarrassed to admit that I can't figure out how to run the tests from the command line (as a reminder my background is in scala, so I'm still pretty new to plain java) so that's still open. 
* Similar to the kotlin backend, I'm not entirely convinced I have the borrows correct. 
* Finally in order to get this working on CI the CI instance would need jextract. I'm open openning up a PR somewhere to add it, but don't know where to start as I've not worked with Github Actions before.

Also it never feels good opening a PR that looks like this: 
Screenshot 2024-08-14 at 21 23 33
😅

tool/src/java/mod.rs
```diff
@@ -0,0 +1,1461 @@
...
+        };
+
+        (
+            format!("{}.java", o.name).into(),
+            opaque_tpl.render().expect("Failed to render opaque type"),
+        )
+    }
+}
+
+trait PostFix: Sized {
```
Manishearth, line 1019: nit: i'd avoid these helpers, just use Ok()/Err() and`.into()` for Cow

Manishearth (commented):
bit concerned at the PR size: none of the other backends are this big in generated code. unclear what makes this one so big

and yes, we should only land this if we can have CI generating it. otherwise we should land it without the checked-in generated code.

If jextract is annoying to install I'm also potentially in favor of having this backend be developed in a separate repo on the diplomat org with its own CI, this way regular Diplomat contributors don't have to install that tool

it appears like it's not too bad to install, we could add a `cargo make download-jextract` that downloads it to a known location

example/java/ICU4X4J/src/main/java/dev/diplomattest/somelib/ntv/FILE.java
```diff
@@ -0,0 +1,47 @@
+// Generated by jextract
```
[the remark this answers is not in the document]
jcrist1: I think this is one of the major drivers for the PR size. It should be somewhat doable. I can limit the intermediate C code generation to just the files that we need but I think I also need to generate some of the shared C code (e.g. the slices/views). I might need to tweak the jextract call though.

tool/src/java/README.md
```diff
@@ -0,0 +1,37 @@
+* slices appear to be passed as memory segments and lengths and not as slice structs
+* structs and opaque are just memsegs which seem to be basically pointers
+* some functions on Opaque types don't get generated call methods, only something to create an "invoker"
+which must be called separately
+
+
+## Invoker functions
```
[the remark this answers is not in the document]
jcrist1: I'm pretty sure it's unnecessary. Something changed in the c codegen, and the resulting functions no longer need to be instantiated as "invokers"

tool/templates/java/Struct.java.jinja
```diff
@@ -0,0 +1,51 @@
...
+import static java.lang.foreign.ValueLayout.*;
+import java.nio.charset.StandardCharsets;
+import java.util.stream.Stream;
+
+public class {{type_name}} {
+    {% for field in fields -%}
+    {{field.ty}} {{field.name}};
+    {% endfor %}
+
+    MemorySegment internal;
```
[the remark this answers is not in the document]
jcrist1: That helps. It should be possible to mostly use confined arenas for structs then.

jcrist1:
This will probably sit for a little bit again while I'll be looking into the kotlin ci setup, as that will be a good intro to github actions so I can set something up later for this.

jcrist1:
I added ci tasks for generating the java code and testing the java code. However I'm having some issues with the test task it doesn't actually do this https://github.com/rust-diplomat/diplomat/actions/runs/10654429057/job/29530731634?pr=499#step:5:67
```
-------------------------------------------------------
 T E S T S
-------------------------------------------------------

Results :

Tests run: 0, Failures: 0, Errors: 0, Skipped: 0
```
When I run `cargo make test-java` locally it actually runs tests.
```
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running dev.diplomattest.somelib.Float64VecTest
WARNING: A restricted method in java.lang.foreign.SymbolLookup has been called
WARNING: java.lang.foreign.SymbolLookup::libraryLookup has been called by dev.diplomattest.somelib.ntv.somelib_h in an unnamed module
WARNING: Use --enable-native-access=ALL-UNNAMED to avoid a warning for callers in this module
WARNING: Restricted methods will be blocked in a future release unless native access is enabled

[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.157 s -- in dev.diplomattest.somelib.Float64VecTest
[INFO] Running dev.diplomattest.somelib.OpaqueTest
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.017 s -- in dev.diplomattest.somelib.OpaqueTest
[INFO] Running dev.diplomattest.somelib.MyStringTest
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.003 s -- in dev.diplomattest.somelib.MyStringTest
[INFO]
[INFO] Results:
[INFO]
[INFO] Tests run: 3, Failures: 0, Errors: 0, Skipped: 0
[INFO]
```
So I need to to investigate a bit more, but am honestly a bit frustrated with Java right now, and if anyone who knows something about maven and java can help I'd definitely appreciate it. 
`</screaming into void>`

example/java/ICU4X4J/src/main/java/dev/diplomattest/somelib/ntv/FILE.java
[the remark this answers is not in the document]
jcrist1: For now deleting the example code for java

tool/templates/java/Lib.java.jinja
```diff
@@ -0,0 +1,218 @@
...
+import java.lang.ref.Cleaner;
+import java.nio.charset.StandardCharsets;
+import java.util.function.Consumer;
+import java.util.Arrays;
+
+import static java.lang.foreign.ValueLayout.*;
+
+public class Lib {
+
+    public sealed interface Result<T, E> {
```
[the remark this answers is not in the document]
jcrist1: We don't support fallible returns yet, so these types were very premature, but god to know.

tool/src/java/mod.rs
jcrist1: removed

feature_tests/java/somelib/src/jmh/java/dev/diplomattest/somelib/OpaqueBench.java
```diff
@@ -0,0 +1,28 @@
+package dev.diplomattest.somelib;
+
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.infra.Blackhole;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.RunnerException;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+
+public class OpaqueBench {
```
jcrist1, line 10: I don't have a shell command to actually run this yet. I only ran it from IntelliJ, but want to keep it as it'll be nice to have to compare to the kotlin backend once

feature_tests/java/somelib/src/main/java/dev/diplomattest/somelib/Opaque.java
jcrist1: ☝️ no longer valid 😝

feature_tests/java/somelib/src/main/java/dev/diplomattest/somelib/ntv/FILE.java
jcrist1, line 1: A lot of files seem to be generated by default. It seems to be possible to pass in every single function and struct to jextract to specify which ones should be used to generate code, but I can't see a way to only forbid these standard constructions which are causing a lot of noise. I'd like to move that investigation to a followup.

feature_tests/src/slices.rs
```diff
@@ -17,6 +17,7 @@ mod ffi {
             Box::new(Self(v.to_string()))
         }
 
+        #[diplomat::attr(supports = memory_sharing, disable)]
```
jcrist1, line 20: @Manishearth I had an issue with passing owned strings in java as then I needed to pass an arena around which felt contradictory to the idea of an owned parameter.

runtime/src/lib.rs
```diff
@@ -33,6 +33,7 @@ pub type DiplomatStr = [u8];
 pub type DiplomatStr16 = [u16];
 
 /// Like [`u8`], but interpreted explicitly as a raw byte as opposed to a numerical value.
+///
```
jcrist1, line 36: I think this came from a clippy fix, but I'm not sure

tool/src/c/mod.rs
```diff
@@ -38,15 +38,15 @@ pub(crate) fn attr_support() -> BackendAttrSupport {
     a
 }
 
+#[derive(askama::Template)]
+#[template(path = "c/runtime.h.jinja", escape = "none")]
+pub(crate) struct Runtime;
```
jcrist1, line 43: need access to the c runtime for the java backend so that jextract can process it.

tool/src/java/mod.rs
```diff
@@ -0,0 +1,1667 @@
...
+    errors: &'a ErrorStore<'cx, String>,
+}
+
+#[derive(Debug, Clone)]
+struct Conversion<'cx> {
+    conversion: Cow<'cx, str>,
+    converted_value: Cow<'cx, str>,
+}
+
+mod arena {
```
jcrist1, line 363: I wanted param conversions to not have to be aware of the context wherein they are done, and so introduced a kind of inversion of control, where if a conversion needs an arena it returns a variant that can only be accessed by providing an arena.

tool/src/java/mod.rs
```diff
@@ -0,0 +1,1667 @@
...
+}
+
+enum AllocateConversion<'cx, F: Fn(arena::Arena<'cx>) -> Conversion<'cx>> {
+    NotAllocating(Conversion<'cx>),
+    Allocating(arena::AllocationConversion<'cx, F>),
+}
+
+impl<'a, 'cx> TyGenContext<'a, 'cx> {
+    #[allow(unused)]
+    fn java_to_native<T: TyPosition>(
```
jcrist1, line 413: As per your comment last review @Manishearth I tried to harmonise the param conversion to just a language conversion, so it can be reused in e.g. the struct conversion. It feels a lot cleaner but I admit the control inversion above is a little weird.

jcrist1:
Okay @Manishearth I think this could do with another review. I substantially redid the generic native <-> java conversion coded. Also added the ci steps. I'm still missing a ci step that actually runs the tests, but I'd like to wrap this up, and focus on smaller pieces at a time.

tool/src/java/mod.rs
jcrist1: I actually implemented that. Biggest problem is that I need the C runtime too, which I made pub (crate)

example/java/ICU4X4J/src/main/java/dev/diplomattest/somelib/ntv/FILE.java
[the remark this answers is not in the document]
jcrist1: I left a comment in the PR on the corresponding file for `feature`. This is automatically created by jextract and I couldn't quickly see how to keep it from being generated. Apparently you can declare which methods and structs to create, but I couldn't see how to keep this standard stuff from being created.

tool/src/java/mod.rs
```diff
@@ -0,0 +1,1461 @@
...
+        let fields = s
+            .fields
+            .iter()
+            .map(|field| {
+                let name = self.formatter.fmt_field_name(field);
+                let native_name = field.name.as_str();
+                let native_val =
+                    format!("{domain}.{lib_name}.ntv.{type_name}.{native_name}(structSegment)");
+                let field_val = match &field.ty {
+                    hir::Type::Enum(ref enum_def) => format!(
```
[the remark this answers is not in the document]
jcrist1: I substantially redid this, and left some comments below about it.

jcrist1:
Screenshot 2024-09-13 at 15 35 47
🎉 🎉 🎉 🎉 🎉 🎉

Manishearth (commented):
oops, never posted this

feature_tests/java/somelib/src/main/java/dev/diplomattest/somelib/ntv/NestedBorrowedFields.java
```diff
@@ -0,0 +1,219 @@
+// Generated by jextract
```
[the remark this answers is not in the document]
Manishearth: this comment was supposed to be on the threading files below this

Manishearth:
Alright, so my biggest concern with this is making the JDK a part of the `gen` step. "Run tests for a different backend" isn't a common task for Diplomat development, but "regen test code for all backends" is (since it's common enough to add stuff to the `feature_tests` crate), and this means all Diplomat devs need to set up JDK, something that can often be very finicky.

I would _like_ for the Java backend to be a part of this repo, however. We have a couple routes in front of us, and I'd like to know what you think:

 1. Diplomat devs get used to setting up JDK, potentially with us merging PRs that temporarily break the java gen task (we can keep it a separate CI job) with the understanding that one of us with a JDK setup will fix it. I can commit to doing so, though I'd rather not.
      - Downside: This is going to be really annoying for other Diplomat devs. I'm open to this but I'd rather not go down this route.
 3. This backend is added in-tree, however we do not check in generated code for it, and CI runs `gen-java` before running tests.
     - Downside: we lose out on Java diffs
4. This backend is added in-tree, and we check in regular generated code but not the jextracted code, and CI runs jextract (or perhaps some command like `diplomat-tool --backend jextractwrapper`)
   - Downside: Kind of weird, we still lose out on diffs but not the important ones
5. This backend is maintained as `rust-diplomat/java` as a separate crate, and has full CI there.
   - Downside: Diplomat changes to the HIR don't automatically get nicely propagated there, it's possible for the backends to diverge more in features. Right now even if I don't maintain the Kotlin backend whenever I change something it is relatively easy for me to ensure that Kotlin is kept up to date most of the time.
 6. We manually generate bindings using JEP 454 APIs rather than using `jextract`
   - Downside: this is a bunch of extra work


The last one is my preferred end state, honestly: when we were considering Panama we largely intended for the JEP 454 `java.lang.foreign` (etc) APIs to be used, but not `jextract`. I think `jextract` is a great way to bootstrap this project, though.


A hybrid route I find somewhat appealing is to take option 4 or 5 above ("don't check in the full bindings"), and over time move to a diplomat that doesn't need jextract. A particularly cool option would be for diplomat to generate an `ntv/` folder that is API compatible with jextract, so we can run tests swapping out the diplomat-generated ntv with the jextracted ntv.


@jcrist1 what do you think? Ultimately you're the one doing the work here, so I don't want to dictate anything, but I want to try and resolve the tradeoff with the impact on other backends, and try to understand your own preferences here.

jcrist1:
@Manishearth thanks for the feedback. I'll probably just implement a filter to remove the extraneous generated files from jextract ... depending on the conclusion of what to do with this backend

> ... need to set up JDK, something that can often be very finicky.

write once run... something something ... I forget the rest 😛 

Honestly the jextract thing bothers me from a rust devx perspective as well as an end user perspective. I agree that generating the JEP 454 APIs will probably be the way to go longer term, but I gotta be honest, I'm not super enthusiastic about doing that myself, not right now at least. While jextract generates _a lot_ of superfluous stuff and is a bit of a pain to setup, it also seems like the easiest way to get functional bindings. 

So for that moving a java backend to its own repo in the diplomat project is a nice compromise. We could get a working backend, and slowly transition to diplomat generated JEP 454 code. Having a repo in the rust-diplomat project would give the panama backend a bit more legitimacy (and hopefully direct other contributors to it) while signalling that the support level is lower and also allowing a more customised dev-setup. We could definitely open an issue around setting up the JEP 454 generation but it's not the most motivating kind of coding for me so progress there will be slow. I would like to pin the diplomat dependency to a specific commit and maybe set up an alert for when the latest main breaks it. That way people can still have a working tool, but I could get notified to address incompatible changes to the diplomat dependency so maintenance wouldn't fall on your shoulders.

jcrist1:
@Manishearth If we're agreed I can try and set up a standalone bin project tomorrow, if you can create the  `rust-diplomat/java` repo

Manishearth:
@jcrist1 here you go: https://github.com/rust-diplomat/diplomat-java

PR #503, opened 2024-06-21
Improve CI, bump Node version

robertbastian:
* Use easier `cargo-make` installer
* Test node version matrix for the three current node versions (18, 20, 22, up from 14)
  * Fix node detection in `wasm.mjs` 
* Split native backends into separate jobs
* Rename `test-wasm*` to `test-js*`, as that's what the backend is called


sffc (commented):
lgtm

PR #506, opened 2024-06-21
Typedef for result structs

robertbastian:
Otherwise they always have to be referred to by `struct Foo_result`


sffc (approved):
@robertbastian showed me all the places that need to change to use `struct`. It's enough that I think the typedef is justified

Issue #504, opened 2024-06-21
Typescript backend is not tested and does not build


robertbastian:

ambiguousname:
JS AST is no longer supported, and this is fixed per #550, closing accordingly.

PR #509, opened 2024-06-22
Typescript Test Fixes

ambiguousname:
Allows building and testing typescript files through cargo make

The only issue is that the JS backend has some issues with adding imports from diplomatRuntime in feature_tests\js\api\Comparable.d.ts and feature_tests\js\api\Opaque.d.ts.

I added a quick fix in fe49266f1ec10e7cf9ad054aa5a6eda298f2c0a5, but the issue needs to be fixed inside the itself backend.

Fix for #504


ambiguousname:
A better version of this is done in #550, so I'm closing this in favor of that. The JS portion should be update to follow #550's example as well.

PR #550, opened 2024-07-16
JS2 Backend Typescript fixes

ambiguousname:
Introduces a number of fixes to ensure typescript tests compile and run.

As a result, this also includes fixes for the typescript portion of the JS2 backend.

### DIFF
--- a/tool/src/js2/formatter.rs
+++ b/tool/src/js2/formatter.rs
@@ -206,7 +206,7 @@ impl<'tcx> JSFormatter<'tcx> {
 
     pub fn fmt_primitive_list_type(&self, primitive: hir::PrimitiveType) -> &'static str {
         match primitive {
-            hir::PrimitiveType::Bool => "Array<bool>",
+            hir::PrimitiveType::Bool => "Array<boolean>",
             hir::PrimitiveType::Char => "Array<char>",
             hir::PrimitiveType::Byte => "Uint8Array",
             hir::PrimitiveType::Int(_)
@@ -220,7 +220,7 @@ impl<'tcx> JSFormatter<'tcx> {
 
     pub fn fmt_primitive_list_view(&self, primitive: hir::PrimitiveType) -> &'static str {
         match primitive {
-            hir::PrimitiveType::Bool => "bool",
+            hir::PrimitiveType::Bool => "boolean",
             hir::PrimitiveType::Char => "u16",
             hir::PrimitiveType::Byte => "u8",
             hir::PrimitiveType::Int(hir::IntType::I8) => "i8",
--- a/tool/src/js2/formatter.rs
+++ b/tool/src/js2/formatter.rs
@@ -150,13 +150,11 @@ impl<'tcx> JSFormatter<'tcx> {
             match primitive {
                 hir::PrimitiveType::Bool => "boolean",
                 hir::PrimitiveType::Char => "char",
+                hir::PrimitiveType::Int(hir::IntType::I64) | hir::PrimitiveType::Int(hir::IntType::U64) | hir::PrimitiveType::Int128(_) => "bigint",
                 hir::PrimitiveType::Int(_)
                 | hir::PrimitiveType::IntSize(_)
                 | hir::PrimitiveType::Byte
                 | hir::PrimitiveType::Float(_) => "number",
-                hir::PrimitiveType::Int128(_) => {
-                    panic!("Javascript backend does not currently support BigInt.")
-                }
             }
         } else {
             match primitive {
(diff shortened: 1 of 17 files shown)


ambiguousname:
This is a partial fix for #509 and #504 (this does not adjust the original JS backend, just the JS2 backend)

Manishearth (approved):

example/js2/api/FixedDecimal.d.ts
robertbastian, line 1: nit: these are not marked as generated files

tool/src/js2/formatter.rs
```diff
@@ -150,13 +150,13 @@ impl<'tcx> JSFormatter<'tcx> {
             match primitive {
                 hir::PrimitiveType::Bool => "boolean",
                 hir::PrimitiveType::Char => "char",
+                hir::PrimitiveType::Int(hir::IntType::I64)
+                | hir::PrimitiveType::Int(hir::IntType::U64)
```
robertbastian, line 154:
```suggestion
                hir::PrimitiveType::Int(hir::IntType::I64 | hir::IntType::U64)
```

tool/src/js2/formatter.rs
```diff
@@ -235,9 +235,8 @@ impl<'tcx> JSFormatter<'tcx> {
             hir::PrimitiveType::IntSize(hir::IntSizeType::Usize) => "usize",
             hir::PrimitiveType::Float(hir::FloatType::F32) => "f32",
             hir::PrimitiveType::Float(hir::FloatType::F64) => "f64",
-            hir::PrimitiveType::Int128(_) => {
-                panic!("Javascript backend does not currently support BigInt")
-            }
+            hir::PrimitiveType::Int128(hir::Int128Type::I128) => "i128",
+            hir::PrimitiveType::Int128(hir::Int128Type::U128) => "u128",
```
robertbastian, line 239: I don't think this is supported, `DiplomatBuf.slice` doesn't currently handle these values, because there's no `Uint128Array`

tool/templates/js2/method.js.jinja
```diff
@@ -5,7 +5,7 @@
     {%- if typescript %}: {{ param.ty -}}{%- endif -%}
     {%- if !loop.last -%}, {% endif -%}
 {%- endfor -%}
-) {%- if typescript %}{{ return_type }};{% else %} {
+) {%- if typescript %}{% if return_type != ": void" %}{{ return_type }}{% endif %};{% else %} {
```
robertbastian, line 8: nit: in order to have less logic in the template, the return type can be set to `""` in Rust code

tool/src/js2/formatter.rs
```diff
@@ -206,21 +206,21 @@ impl<'tcx> JSFormatter<'tcx> {
...
-            hir::PrimitiveType::Int128(_) => {
-                panic!("Javascript backend does not currently support BigInt.")
-            }
         }
     }
 
     pub fn fmt_primitive_list_view(&self, primitive: hir::PrimitiveType) -> &'static str {
         match primitive {
-            hir::PrimitiveType::Bool => "bool",
+            hir::PrimitiveType::Bool => "boolean",
```
robertbastian, line 223: The `DiplomatBuf.slice` constructor uses `bool` and needs to be updated

tool/templates/js2/method.js.jinja
robertbastian, line 8: Actually I don't think this is correct. When I apply this, I get the warning

```
<method> implicitly has an 'any' return type, but a better type may be inferred from usage. ts(7050)
```

However, this *does* seem to be required for setters, where without I get the error

```
A 'set' accessor cannot have a return type annotation. ts(1095)
```

robertbastian (changes requested):

tool/src/js2/formatter.rs
robertbastian: done

tool/src/js2/formatter.rs
robertbastian: done

tool/templates/js2/method.js.jinja
robertbastian: done

tool/src/js2/formatter.rs
robertbastian: done

tool/templates/js2/method.js.jinja
robertbastian: done

robertbastian (approved):
I applied some fixes.

ambiguousname (commented):
LGTM!

PR #552, opened 2024-07-17
Callback support in the AST/HIR/macro, and codegen for C

emarteca:
This PR addresses [@146](https://github.com/rust-diplomat/diplomat/issues/146): adding callback support

---

We add callback support from the AST, through the HIR, and to the macro. This PR adds support for the C backend.

We describe the details of our design our design doc [linked here](https://docs.google.com/document/d/1sWeYZY182RQDac3UKtiWyHc4opXap4c2t4jlTECkLgA/edit?tab=t.0), and in the discussion on this PR!


Manishearth:
cc @jcrist1

example/simple_testing/src/lib.rs
```diff
@@ -0,0 +1,137 @@
+// #[diplomat::bridge]
+// mod ffi {
+
+fn test_rust_fn(f: impl Fn(i32) -> i32) {
```
Manishearth, line 4: nit: Diplomat doesn't really do standalone functions (yet). I'd recommend writing this to look more like a static function on some type so that it's easier to follow along

example/simple_testing/src/lib.rs
```diff
@@ -0,0 +1,137 @@
...
+        (*fw).get_arg0()
+    }
+}
+
+// function that actually calls the callback
+// it's basically a heavily instrumented version of the original Rust function that takes
+// a callback as an argument.
+// it wraps the arguments to the callback in the arg wrapper struct, and calls the callback
+#[no_mangle]
+pub unsafe extern "C" fn GEND_BRIDGE_test_run_fn(cb: &DiplomatCallback) {
```
Manishearth, line 132: in Diplomat, this function would be called `nameoftype_nameofmethod`. The name initially confused me, I thought this was the trampoline.

example/simple_testing/kotlin/src/main/kotlin/dev/diplomattest/simple_testing/Main.kt
```diff
@@ -0,0 +1,113 @@
...
+}
+
+class FFIArgsWrapper_f internal constructor (
+    internal val handle: Pointer,
+    // These ensure that anything that is borrowed is kept alive and not cleaned
+    // up by the garbage collector.
+    internal val selfEdges: List<Any>
+)  {
+
+    internal class FFIArgsWrapper_fCleaner(val handle: Pointer, val lib: FFIArgsWrapper_f_Lib) : Runnable {
```
Manishearth, line 22: does Kotlin ever need to call a dtor on this type?

example/simple_testing/kotlin/src/main/kotlin/dev/diplomattest/simple_testing/Main.kt
```diff
@@ -0,0 +1,113 @@
...
+        }
+    }
+
+    companion object {
+        internal val libClass: Class<FFIArgsWrapper_f_Lib> = FFIArgsWrapper_f_Lib::class.java
+        internal val lib: FFIArgsWrapper_f_Lib = Native.load("somelib", libClass, 
+            Collections.singletonMap(Library.OPTION_ALLOW_OBJECTS, true))
+
+        fun get_arg0_from_ffiargswrapper_f_pointer(handle: Pointer): Int {
+            return lib.get_arg0_from_ffiargswrapper_f_pointer(handle);
```
Manishearth, line 34: nit: we already have the ability to do this in pure Kotlin using JNA

jcrist1:
I had a look at the code, and am still trying to wrap my head around it a bit, (also the design doc and @146). Although I think I get why we would need to support passing runtimes / context to rust in general (i.e. what we're doing with the JNIEnv here), it looks like JNA supports callbacks without passing extra context via the `com.sun.jna.Callback` class (see https://github.com/java-native-access/jna/blob/master/www/CallbacksAndClosures.md). I created a minimal PoC in a dedicated branch [here](https://github.com/rust-diplomat/diplomat/compare/main...jcrist1:diplomat:kotlin-jna-callback?expand=1). Could we instead use `jna.Callback` for the kotlin backend?

emarteca:
> I had a look at the code, and am still trying to wrap my head around it a bit, (also the design doc and @146). Although I think I get why we would need to support passing runtimes / context to rust in general (i.e. what we're doing with the JNIEnv here), it looks like JNA supports callbacks without passing extra context via the `com.sun.jna.Callback` class (see https://github.com/java-native-access/jna/blob/master/www/CallbacksAndClosures.md). I created a minimal PoC in a dedicated branch [here](https://github.com/rust-diplomat/diplomat/compare/main...jcrist1:diplomat:kotlin-jna-callback?expand=1). Could we instead use `jna.Callback` for the kotlin backend?

Thank you for taking a look and for building this example! We originally thought the JNA Callback wouldn't fit in a model that generalized to other languages, but looking at your example it seems like we might be wrong. We're going to try and adapt our example to use it now!

We'll also need to think about if this strategy generalizes to callbacks that are not functions, such as closures.

Also, using the JNA Callback means that the lifetime of the callback would need to be managed by the foreign language. Not necessarily a problem, but something to consider too.

maurer:
tl;dr: JNA callbacks are kind of icky technically, but they'll probably make for an easier initial implementation and they aren't intolerable. I still think we should pack them into the struct representation though.

I was looking into how JNA callbacks worked, because it didn't seem like they could possibly be minting new function pointers for every callback, but they were also allowing non-static methods, which meant their callbacks could close over dynamic state, which meant they'd need an unbounded one.

It turns out, they are minting new function pointers for every callback:
1. [Allocate memory](https://github.com/java-native-access/jna/blob/a587921bac074b1bd1b0a0a5536587660a9b954e/native/libffi/src/closures.c#L61-L102)
2. [Generate arch specific raw asm trampolines](https://github.com/java-native-access/jna/blob/a587921bac074b1bd1b0a0a5536587660a9b954e/native/libffi/src/x86/ffi64.c#L754)

This means that creating a callback costs 3 system calls, and costs 2 pages of memory minimum while active.

This is necessary if you want to interface with a *strict* existing C interface - they may only take raw function pointers, and this is the only way to generate them from Java - call a C function that literally makes them by assembling in the process.

In our case, since we can make the ABI look like what we want, I think there might be value in using the struct-based design (e.g. 
```
struct Callback<R> {
  data: *const c_void,
  run: unsafe extern "C" fn(*const c_void, ...) -> R;
  drop: unsafe extern "C" fn(*const c_void);
}
```
)
makes sense. In an initial implementation, we can use JNA callbacks and set `run` equal to a function pointer by Java, `data` to null, and `drop` to a no-op. This leaves the door open to a C++ callback implementation where `run` calls a `std::function`'s invocation method, and `drop` tears it down (applied to `data` containing the `std::function`). This also leaves the door open to a more optimized Java backend in the future, if we make it possible for JNA to produce the `callback` struct without bothering with the `closure` and `x_closure` field initialization (possibly via a flag). We can then use their existing dispatch function to create a static `run`, and wire up `drop` to the teardown for the `*callback` rather than needing to be careful Java side.

jcrist1:
@emarteca @maurer, thanks for the detailed explanations 😄. I had a feeling there was some subtlety that I wasn't aware of. 
In any case 👍 for the struct with null data and no-op drop.

My general understanding is that jna/jni is anyway not the most performant way to do native interop in the JVM. I'm working (slowly) on a Panama java backend, and I can try and think about what callbacks would look like there with this framework. Though I think it will be a while before I get there. [This](https://github.com/openjdk/jextract/blob/c2eed20c9050c42bf9274239c7ed17c835951977/test/jtreg/generator/test8246400/LibTest8246400Test.java#L60) looks like an example of a pure C callback (but I feel like it has to be doing the same thing as jna anyway 🤔, maybe [this](https://github.com/openjdk/jextract/blob/c2eed20c9050c42bf9274239c7ed17c835951977/src/main/java/org/openjdk/jextract/clang/libclang/CXCursorVisitor.java#L49)). In any case definitely open to suggestions as to how to do it there in the most performant way. I'd love to be kept in the loop on this, as the work y'all're doing looks cool.

emarteca:
That makes sense to use JNA callbacks for now! 
We updated the code example such that we still have the `DiplomatCallback` struct, but the `run` function just calls the JNA callback and have `drop` not do anything, as discussed above. I ended up using the `data` pointer to store a pointer to the JNA callback, so that it can be passed as an argument to the `run` function.

Also, the struct is parameterized over the callback return type now, so we support returns (that came for free in the JNA callback example :) )

example/simple_testing/kotlin/src/main/kotlin/dev/diplomattest/simple_testing/Main.kt
```diff
@@ -0,0 +1,54 @@
...
+    companion object {
+        internal val libClass: Class<DiplomatCallback_Lib> = DiplomatCallback_Lib::class.java
+        internal val lib: DiplomatCallback_Lib = Native.load("somelib", libClass)
+
+        // create the JNA Callback object
+        internal val callback: DiplomatCallback_Lib.DiplomatCallbackI32ToI32 = object : DiplomatCallback_Lib.DiplomatCallbackI32ToI32{
+            override fun invoke(input: Int): Int {
+                return input*2;
+            }
+        }
```
jcrist1, line 37: I left a similar comment on the design doc, but would the use case for providing callbacks be that as a library user I can create the callback in my backend language (Kotlin in this case) and then create a `DiplomatCallback` instance that I can pass around to other FFI methods/functions? So I would expect you'd want generated code to create something like this (for each possible signature in the rust library we're running the diplomat tool on)
```kotlin
fun fromFunction(cb: (Int) -> Int): DiplomatCallbackIntToInt {
  val callback: DiplomatCallback_Lib.DiplomatCallbackI32ToI32 = object :  DiplomatCallback_Lib.DiplomatCallbackI32ToI32 {
      override fun invoke(input: Int): Int {
          return cb(input);
      }
    }
    val callback_pointer = lib.diplomat_callback_create_for_jvm__callback(callback);
    return DiplomatCallback(callback_pointer, {callback})
  }
}
```

Then any kotlin method generated from a rust method expecting a callback would then expect a `DiplomatCallback`
```kotlin
interface MyClassLib: Library {
  // general use
  fun MyClass_test_rust_fn_with_callback(cb: Pointer): Int;
}

class MyClass {
  companion object {
    internal val lib: ....
    fun testRustFnWithCallback(cbWrap: DiplomatCallbackIntToInt): Int {
      return lib.MyClass_test_rust_fn_with_callback(cbWrap.pointer)
    }
  }
}
```

Which would be generated from this.
```rust
#[diplomat::bridge]
mod ffi {
  
  pub struct MyClass;

  impl MyClass {
    pub fn test_rust_fn_with_callback(cb: impl Fn(i32) -> i32) -> i32 {
      cb(10)
    }
  }
}
```

Then someone comes along and wants to use the kotlin library generated from rust code. They would want to provide their own callback, like so:
```kotlin
object Main {
  @JvmStatic
  fun main(args: Array<String>) {
    val callBack = DiplomatCallback.fromFun( (input) -> input * 2))
    MyClass.testRustFnWithCallback(callBack)
  }
}
```
Or would we rather expect that the generated method takes a `fun` directly:
```kotlin
    fun testRustFnWithCallback(cbWrap: (Int) -> Int): Int {
      val cbWrapDiplomatCallback = DiplomatCallback.fromFun(cbWrap))
      return lib.MyClass_test_rust_fn_with_callback(cbWrapDiplomatCallback.pointer)
    }
```
and someone can come along and do
```kotlin
object Main {
  @JvmStatic
  fun main(args: Array<String>) {
    MyClass.testRustFnWithCallback({input -> 2 * input})
  }
}
```

emarteca:
Ah, that's a good point! I think passing the callback in to construct the callback wrapper on the Kotlin side makes sense. This also means we'd just need one wrapper per callback type signature, rather than one for each callback. 

Pushed an updated example, closely matching your suggestion!

jcrist1 (approved):
Don't know how much my approval is worth here, but I think all the pieces make sense. I think how they fit together may change a bit as you add the HIR for callbacks, macro expansion and different bits of code gen, but generally I think the best way of figuring those out is to build a minimal end-to-end working pipeline. I don't know if you have a minimum viable callback idea, but I think something like a zero sized type with a single method taking an I32 -> SomeOtherType like
```rust
#[diplomat::bridge]
mod ffi {
  
  pub struct MyClass;

  impl MyClass {
    pub fn test_rust_fn_with_callback(cb: impl Fn(i16) -> i32) -> i64 {
      cb(10) as i64
    }
  }
}
```
could be a good initial goal to build towards.

emarteca:
Hi! I just pushed a draft of changes to the HIR so that we can represent callbacks. The main change is that we have to add a Callback type and this type needs to be input-only. 

In designing the change to HIR, we saw 2 main ways of going about this
1) switching from `TyPosition` to type polarity
2) go with the current structure and add an `InputOnly` type

In the design doc [in this linked section](https://docs.google.com/document/d/1sWeYZY182RQDac3UKtiWyHc4opXap4c2t4jlTECkLgA/edit?tab=t.0) we talk about both options. For sake of simplicity, with the code pushed here, we went with the `InputOnly` version, since that requires less modification to the existing type representation.

emarteca:
I also just pushed a first step of parsing callbacks from the AST. Unfortunately, for arguments of type `impl Fn (args)->ret` the type is just a [`syn::Type::ImplTrait`](https://docs.rs/syn/2.0.72/syn/enum.Type.html), so extracting the function signature is pretty verbose.

With this change, running diplomat on our example
```
#[diplomat::bridge]
mod ffi {

    pub struct Wrapper {}
    impl Wrapper {
        pub fn test_rust_fn(f: impl Fn(i32) -> i32) -> i32 {
            f(10)
        }
    }

}
```
results in the message 
```
not yet implemented: Not supported yet, but we're working on it! You have a function type of [Primitive(i32)]->Primitive(i32)
```
which is the right function signature! We'll plan out how we'll extend the AST type representation to include a function type next.

core/src/hir/defs.rs
```diff
@@ -19,6 +20,8 @@ pub enum TypeDef<'tcx> {
     OutStruct(&'tcx OutStructDef),
     Opaque(&'tcx OpaqueDef),
     Enum(&'tcx EnumDef),
+    Callback(&'tcx Callback),
```
Manishearth, line 23: issue: I don't think Callbacks should be a separate instance of TypeDef, I think Callbacks should be a variant of Type

https://docs.rs/diplomat_core/latest/diplomat_core/hir/enum.Type.html

There's no reason for callbacks to be separately defined IMO, they can be mentioned in the method and that ought to be sufficient.

Manishearth: basically, TypeDef is for custom named types defined by the user, like `struct Foo {}`

Type is for actual types being used in a struct field, parameter, or return type, and can be a broader set of things.

Manishearth (commented):
I suspect the current design will lead to some problems during lowering

Also, it may be beneficial to introduce InputType independently of this PR as a standalone thing, even if it ends up being identical to Type.

emarteca:
Good idea re: having `InputType` in its own PR: https://github.com/rust-diplomat/diplomat/pull/595

emarteca:
I just pushed a prototype of the pipeline from parsing, through the AST, to HIR, so now the example crashes on code generation! 

Now we get the HIR `Method` with the proper `Callback` type for its parameter.
From our example
```
pub struct Wrapper {
        cant_be_empty: bool, // apparently we don't support zero-sized structs
    }
    impl Wrapper {
        pub fn test_rust_fn(f: impl Fn(i32) -> i32) -> i32 {
            f(10)
        }
    }
```
We get the output (I added some debug logs, to print when a `Callback` and a `Method` are parsed):
```
PARSED A CALLBACK ARG Callback(Callback { id: "f", param_self: None, params: [Param { name: "na", ty: Primitive(Int(I32)) }], output: Primitive(Int(I32)) })

PARSED A METHOD Method { docs: Docs("", []), name: "test_rust_fn", lifetime_env: LifetimeEnv { nodes: [], num_lifetimes: 0 }, param_self: None, params: [Param { name: "f", ty: Callback(Callback { id: "f", param_self: None, params: [Param { name: "na", ty: Primitive(Int(I32)) }], output: Primitive(Int(I32)) }) }], output: Infallible(OutType(Primitive(Int(I32)))), attrs: Attrs { disable: false, namespace: None, rename: RenameAttr { pattern: None }, abi_rename: RenameAttr { pattern: None }, special_method: None } }

thread 'main' panicked at tool/src/kotlin/mod.rs:1364:18:
internal error: entered unreachable code: unknown AST/HIR variant
```

So now it is crashing on the code generation.

*Simplifications for this prototype*:  I'll still need to fix, before thinking about code gen:
* As @Manishearth you mentioned, using `InputOnly` for the `Callback` got complicated, mainly due to then being unable to use `lower_param` and the other lowering functions as-is, because they expect `Type<Everywhere>`. And also, because the callbacks are going to appear as the types of parameters of `Method`, which has a parameter list `Vec<Param<Everywhere>>`. So for this prototype, I just changed `Callback` to be able to exist in the `Everywhere` position too.
* Following from the first simplification, I didn't add any checks that the `Callback`s are actually appearing in input-only positions

Also note: I'm reusing `Param` for the arguments of `Callback`, which may not be the best fit since they have a name. In the example, I'm assigning them all the placeholder name `"na"`.

emarteca:
Cleaned up the code from yesterday; the resulting parse is the same, but now `InputOnly` is plumbed through via an enum `CanBeInputType`, that wraps both `InputOnly` and `Everywhere` since these are both valid options for `Param` typosition.

Also added a new struct `CallbackParam` to represent the parameters of callbacks instead of reusing `Param`. This allows them to be `OutputOnly` while changing `Param` to have a type `CanBeInputType`, and it is also cleaner because then there is no need for a placeholder "name".

emarteca:
At this point, we have a working draft of the changes for callback support in C! We've described the changes to the HIR, AST, and code generation that we made starting from [this section](https://docs.google.com/document/d/1sWeYZY182RQDac3UKtiWyHc4opXap4c2t4jlTECkLgA/edit?tab=t.0#heading=h.lt53hyz31cwi) of our design doc.

Would appreciate any feedback on the code and/or design so far! :pray:

example/simple_testing/c/diplomat_callback_structs.h
```diff
@@ -0,0 +1,17 @@
+#ifndef Diplomat_callback_structs_H
+#define Diplomat_callback_structs_H
+
+typedef struct DiplomatCallback_int32_t {
```
Manishearth, line 4: if we have DiplomatCallback use an outparam for the return value we don't need to do this and we can always use `DiplomatCallback` everywhere. We've tried to move the C backend away from needing to generate such typedefs, it causes problems.

example/simple_testing/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
```
Manishearth, line 1: For testing this, what I recommend doing is:

 - Adding `callbacks` to `BackendAttrSupport`.
 - Setting it to true in the C backend, and only the C backend.
 - Using `feature_tests`, and using `cfg(not(supports = callbacks), disable)` on any method that uses callbacks. Make a separate `callbacks.rs` test file.

macro/src/lib.rs
```diff
@@ -194,6 +209,24 @@ fn gen_params_invocation(param: &ast::Param, expanded_params: &mut Vec<Expr>) {
...
+            let mut cb_param_list = vec![];
+            let mut cb_arg_type_list = vec![];
+            for (index, in_ty) in in_types.into_iter().enumerate() {
+                cb_param_list.push(Ident::new(&format!("arg{}", index), Span::call_site()));
+                cb_arg_type_list.push(in_ty.to_syn());
+            }
+            let cb_ret_type = out_type.to_syn();
+
+            let tokens = quote! {
+                move | #(#cb_param_list,)* | unsafe {
```
Manishearth, line 223: one thing I'm a bit unsure of: I was under the impression we performed some amount of type conversion between the FFI type and the Rust type in the macro

specifically:

 - slices are passed over FFI as a `data, len` pair of parameters
 - we call `from_raw_parts_mut()` etc to make that work
 - we also convert `Result` to `DiplomatResult` where needed (this is not yet relevant here)

I think we need to be handling the slices here, similarly, yes?

Though at this stage perhaps it's better for Diplomat to just uniformly use the DiplomatFooView types (cc @robertbastian)? This avoids the issue of splitting things into multiple args, quite cleanly.

Manishearth: We'd still need to call `.into()`, but it would save us the trouble of splatting the arguments

Manishearth: Filed https://github.com/rust-diplomat/diplomat/issues/613. Not a problem for you now, I might refactor a bunch of the macro code if I fix #613, which I think i will

emarteca: Ah that would be helpful, thanks! I'll refactor to fit your changes when you fix that.

example/simple_testing/c/diplomat_callback_structs.h
emarteca: I replied to your comment about this in the design doc, but just to duplicate it here: 

We came up with a different idea to avoid the extra file of structs: rather than generating one `DiplomatCallback` struct per return type (which is currently what we do in C), we would generate one per function argument. So, e.g., if there is a function `foo` with a callback parameter `bar` in struct `BazStruct`, then we would generate `DiplomatCallback_BazStruct_foo_bar`. 

This has a few advantages over the current approach (and which fix some of the issues you listed): 
1. we'll have the specific argument list for the `run_callback` function, instead of having a variadic function 
2. there will never be a name collision if we use multiple crates through Diplomat (which currently would happen, if we ran Diplomat on 2 crates and they both have functions with callback arguments that have the same return type)
3. no need to keep track of which callback types we have already seen, since we'd be generating one `DiplomatCallback` per callback argument 
4. no extra file with `DiplomatCallback` structs: they would be included in the same file as the other generated struct functions

The downside is that there is more duplicated code generated, but this shouldn't come at a performance cost.

Including a pointer to the output in the `run_callback` function would be doable too, but this has the downside of requiring the user to set up the output pointer in advance, and would make the use of the `DiplomatCallback` struct less like calling a normal function. Also, this would be a pain to set up with JNI for Kotlin/Java, and potentially for other non-C/C++ languages too.

emarteca:
I've rebased onto the current main and force-pushed, to pull in the new `InputOnly` stuff! As part of that, I also had to refactor all the C code generation to use the new jinja template setup. 

Note: currently I am generating C-specific struct wrapper creation functions in the proc macro; my next task is to get rid of that and generate an equivalent function in C instead.

emarteca:
Update: we're now generating C-specific struct wrapper creation functions in C, so it's removed from the proc macro!

I think the callback support in C is ready for a preliminary code review.

example/simple_testing/c/Wrapper.h
```diff
@@ -0,0 +1,170 @@
...
+    const void* data;
+    int32_t (*run_callback)(int32_t(*cb_pointer)(int32_t), int32_t );
+    void (*destructor)(const void*);
+} DiplomatCallback_Wrapper_test_multi_arg_callback_f;
+
+int32_t run_create_DiplomatCallback_Wrapper_test_multi_arg_callback_f_callback(int32_t(*cb_pointer)(int32_t), int32_t arg0 ) {
+    return cb_pointer(arg0);
+}
+
+DiplomatCallback_Wrapper_test_multi_arg_callback_f C_create_DiplomatCallback_Wrapper_test_multi_arg_callback_f(int32_t(*cb_pointer)(int32_t)) {
```
Manishearth, line 29: question: so this is just a convenience function for C users, yes?

macro/src/lib.rs
```diff
@@ -94,6 +94,21 @@ fn gen_params_at_boundary(param: &ast::Param, expanded_params: &mut Vec<FnArg>)
                 ),
             }));
         }
+        ast::TypeName::Function(..) => {
```
Manishearth, line 97: issue: I would like to have some snapshot tests at the bottom of this file that show the generated code for callback stuff

example/simple_testing/c/Wrapper.h
```diff
@@ -0,0 +1,170 @@
...
+#include "Wrapper.d.h"
+
+
+
+
+
+
+typedef struct DiplomatCallback_Wrapper_test_multi_arg_callback_f {
+    const void* data;
+    int32_t (*run_callback)(int32_t(*cb_pointer)(int32_t), int32_t );
```
Manishearth, line 21: question: I don't quite understand why there is a nested function here

(corollary: don't understand why we need the convenience functions below)

Manishearth (commented):
working my way through it, done with hir/ast code, looking at codegen now

macro/src/lib.rs
Manishearth: After #615, "types needing conversion" is much simpler to think about.


This can be handled in a followup which we should file when we merge this. There are two routes here:

1. We can forbid types that need conversion in callback args. All types that need conversion have a pure-FFI version that looks uglier (potentially) but is perfectly usable
2. We teach the callback macro codegen code to check for types-needing-conversion and call .into() when passing things to the callback


(2) is preferable, but we can go with (1) initially. For this PR it's fine if we just don't perform conversions for now (it will cause compilation failures if you write callbacks that take types-that-need-conversion (Rust slices, later, `Option<T>` for non-opaque T).

macro/src/lib.rs
```diff
@@ -194,6 +209,24 @@ fn gen_params_invocation(param: &ast::Param, expanded_params: &mut Vec<Expr>) {
...
+            let mut cb_arg_type_list = vec![];
+            for (index, in_ty) in in_types.into_iter().enumerate() {
+                cb_param_list.push(Ident::new(&format!("arg{}", index), Span::call_site()));
+                cb_arg_type_list.push(in_ty.to_syn());
+            }
+            let cb_ret_type = out_type.to_syn();
+
+            let tokens = quote! {
+                move | #(#cb_param_list,)* | unsafe {
+                    std::mem::transmute::<unsafe extern "C" fn (*const c_void, ...) -> #cb_ret_type, unsafe extern "C" fn (*const c_void, #(#cb_arg_type_list,)*) -> #cb_ret_type>
```
Manishearth, line 224: thought: `DiplomatCallback` could have a `get_callback()` function that returns an `impl Fn(..)`

It's nice to have the macro codegen be super simple

emarteca:
force pushed to rebase; still addressing code review

core/src/hir/methods.rs
```diff
@@ -36,6 +38,50 @@ pub struct Method {
...
+pub trait CallbackInstantiationFunctionality {
+    fn set_id(&mut self, new_id: &IdentBuf);
+    fn get_input_types(&self) -> Vec<&Type<OutputOnly>>; // the types of the parameters
+    fn get_output_type(&self) -> &Box<Option<Type>>;
+}
+
+#[derive(Debug)]
+#[non_exhaustive]
+pub struct Callback {
+    pub id: IdentBuf, // this will just piggy-back off the name of the parameter the callback corresponds to for now
```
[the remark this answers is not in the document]
emarteca: ah you're right, i don't think it's necessary, since it's equivalent to the `Param` name, and `Callback` is always a `Param`.  removed.

core/src/ast/types.rs
```diff
@@ -456,6 +457,39 @@ impl TypeName {
                 syn::parse_quote_spanned!(Span::call_site() => Box<[#primitive]>)
             }
             TypeName::Unit => syn::parse_quote_spanned!(Span::call_site() => ()),
+            TypeName::Function(_input_types, output_type) => {
+                let output_type = output_type.to_syn();
+                // should be DiplomatCallback<function_output_type>
+                syn::parse_quote_spanned!(Span::call_site() => DiplomatCallback<#output_type>)
+                // original type
+                // let fn_type = syn::Type::ImplTrait(syn::TypeImplTrait {
```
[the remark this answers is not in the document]
emarteca: good catch

example/simple_testing/c/Wrapper.h
emarteca: simplified/removed! good point, that was my bad for not realizing how these functions are typically provided

example/simple_testing/c/Wrapper.h
emarteca: yes -- removed now

core/src/hir/types.rs
```diff
@@ -31,6 +32,7 @@ pub enum SelfType {
     Opaque(OpaquePath<NonOptional, Borrow>),
     Struct(StructPath),
     Enum(EnumPath),
+    // TODO add Callbacks later
```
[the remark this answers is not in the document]
emarteca: I think that makes sense! And traits would be basically be represented as a set of callbacks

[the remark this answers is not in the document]
emarteca: yes, that's what i was thinking! i'll take the comment out for now though; we can add the actual support in the trait PR

macro/src/lib.rs
emarteca: yes, great point.

core/src/hir/methods.rs
```diff
@@ -36,6 +38,50 @@ pub struct Method {
...
+
+// uninstantiatable; represents no callback allowed
+#[derive(Debug, Clone)]
+pub enum NoCallback {}
+
+impl CallbackInstantiationFunctionality for Callback {
+    fn set_id(&mut self, new_id: &IdentBuf) {
+        self.id = new_id.clone();
+    }
+    fn get_input_types(&self) -> Vec<&Type<OutputOnly>> {
```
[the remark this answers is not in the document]
emarteca: good idea

core/src/hir/methods.rs
```diff
@@ -36,6 +38,50 @@ pub struct Method {
...
+    fn set_id(&mut self, new_id: &IdentBuf);
+    fn get_input_types(&self) -> Vec<&Type<OutputOnly>>; // the types of the parameters
+    fn get_output_type(&self) -> &Box<Option<Type>>;
+}
+
+#[derive(Debug)]
+#[non_exhaustive]
+pub struct Callback {
+    pub id: IdentBuf, // this will just piggy-back off the name of the parameter the callback corresponds to for now
+    // pub lifetime_env: LifetimeEnv,
```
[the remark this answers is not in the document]
emarteca: re: offline discussion, we've added a check that none of the callback parameters have lifetimes. We still need to add an option to disable this check though.

core/src/hir/methods.rs
```diff
@@ -36,6 +38,60 @@ pub struct Method {
...
+pub struct Callback {
+    pub param_self: Option<ParamSelf>, // for now it'll be none, but when we have callbacks as object methods it'll be relevant
+    pub params: Vec<CallbackParam>,
+    pub output: Box<Option<Type>>, // this will be used in Rust (note: can technically be a callback, or void)
+}
+
+#[derive(Debug, Clone)]
+pub enum TyPositionErr {
+    OnlyAllowedForInputTypes,
+    OnlyAllowedForOutputTypes,
```
Manishearth, line 58: thought: I don't think this error is used. Perhaps we can just return `Result<_, ()>` here for now?

core/src/hir/types.rs
```diff
@@ -114,6 +115,7 @@ impl<P: TyPosition> Type<P> {
                     .map(|lt| std::slice::from_ref(lt).iter().copied())
                     .unwrap_or([].iter().copied()),
             ),
+            // TODO the Callback case
```
Manishearth, line 118: thought: callbacks basically break this form of lifetime computation anyway since they contain contravariant lifetimes. so this is actually probably okay as long as we're banning borrows in callbacks

feature_tests/src/callbacks.rs
```diff
@@ -0,0 +1,46 @@
+#[diplomat::bridge]
+mod ffi {
+
+    pub struct Wrapper {
```
Manishearth, line 4: nit: over time it becomes annoying to find the right opaques for the right tests, let's call this CallbackWrapper and CallbackTestingStruct

example/simple_testing/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+name = "simple_testing"
```
Manishearth, line 2: issue: simple_testing is no longer needed and can be yeeted

example/simple_testing/kotlin/build.gradle.kts
```diff
@@ -0,0 +1,52 @@
+plugins {
```
Manishearth, line 1: issue: these kotlin files should be removed, yes?

macro/src/lib.rs
```diff
@@ -319,7 +379,11 @@ fn gen_bridge(mut input: ItemMod) -> ItemMod {
 
     for custom_type in module.declared_types.values() {
         custom_type.methods().iter().for_each(|m| {
-            new_contents.push(gen_custom_type_method(custom_type, m));
+            let gen_m = gen_custom_type_method(custom_type, m);
+            new_contents.push(gen_m);
+            // gen_cb_creates.iter().for_each(|cb_create| {
```
Manishearth, line 384: nit: remove

macro/src/snapshots/diplomat__tests__callback_arguments.snap
```diff
@@ -0,0 +1,118 @@
...
+        pub fn test_multiple_cb_args(f: impl Fn() -> i32, g: impl Fn(i32) -> i32) -> i32 {
+            f() + g(5)
+        }
+    }
+    use core::ffi::c_void;
+    use diplomat_runtime::*;
+    #[no_mangle]
+    extern "C" fn Wrapper_test_multi_arg_callback(f: DiplomatCallback<i32>, x: i32) -> i32 {
+        let f = move |arg0: i32| unsafe {
+            std::mem::transmute::<
```
Manishearth, line 46: praise: nice and straightforward

macro/src/lib.rs
```diff
@@ -32,18 +32,78 @@ fn param_ty(param: &ast::Param) -> syn::Type {
...
         }
-        _ => param.ty.to_syn(),
+        _ => param_ty.to_syn(),
     }
 }
 
-fn param_conversion(param: &ast::Param) -> Option<proc_macro2::TokenStream> {
-    let name = &param.name;
-    match &param.ty {
+fn orig_syn(ty: &ast::TypeName) -> syn::Type {
```
Manishearth, line 39: issue: as discussed in chat, this isn't necessary, `ast::TypeName.to_syn()` is expected to roundtrip, and if it doesn't, that's a bug (we identified the bug in the PrimitiveSlice's to_syn impl)

the macro uses `to_syn()` to get the user-specified type, and `param_ty` to convert it to the ffi-safe type

Manishearth (commented):
just need to review the macro conversion code

macro/src/lib.rs
```diff
@@ -32,18 +32,78 @@ fn param_ty(param: &ast::Param) -> syn::Type {
...
         | ast::TypeName::StrSlice(.., StdlibOrDiplomat::Stdlib)
         | ast::TypeName::PrimitiveSlice(.., StdlibOrDiplomat::Stdlib)
-        | ast::TypeName::Result(..) => Some(quote!(let #name = #name.into();)),
+        | ast::TypeName::Result(..) => Some(if let Some(cast_to) = cast_to {
+            quote!(let #name: #cast_to = #name.into();)
+        } else {
+            quote!(let #name = #name.into();)
+        }),
+        ast::TypeName::Function(in_types, out_type) => {
+            let cb_wrap_ident = &name;
```
Manishearth, line 75: thought: i don't like how much code this needs but it's understandable.

Manishearth (approved):

macro/src/snapshots/diplomat__tests__callback_arguments.snap
emarteca: thanks! :)

example/simple_testing/Cargo.toml
emarteca: yeet!

core/src/hir/methods.rs
emarteca: good call

core/src/hir/types.rs
emarteca: ok fair enough!

example/simple_testing/kotlin/build.gradle.kts
emarteca: yes, yeeted with the rest of `simple_example` !

macro/src/lib.rs
emarteca: yeah, maybe later we can put it somewhere else to avoid bulking out the macro code

Manishearth: yep. i already want to fix param_ty and param_conversion to be `to_ffi_safe_tt()` and `ffi_safe_conversion_tt()` or something

macro/src/snapshots/diplomat__tests__slices.snap
```diff
@@ -145,11 +145,11 @@ mod ffi {
         Foo::boxes_runtime_types(a, b, c, d, e, f)
     }
     #[no_mangle]
-    extern "C" fn Foo_a(this: Foo) -> diplomat_runtime::DiplomatSlice<u8> {
+    extern "C" fn Foo_a(this: Foo) -> &[u8] {
         this.a().into()
     }
     #[no_mangle]
-    extern "C" fn Foo_b(this: Foo) -> diplomat_runtime::DiplomatSlice<u16> {
```
Manishearth, line 150: argh, that's a bug. but that's my bug, not yours, i'll fix it

feature_tests/src/callbacks.rs
```diff
@@ -1,41 +1,28 @@
 #[diplomat::bridge]
 mod ffi {
 
-    pub struct Wrapper {
+    pub struct CallbackWrapper {
```
Manishearth, line 4: nit: I would tag this struct (and the other one) with the cfg

macro/src/lib.rs
emarteca: ah sweet, that's a good idea

feature_tests/c/include/AttrEnum.h
```diff
@@ -20,4 +20,5 @@
 
 
 
+
```
Manishearth, line 23: nit: would be nice if we avoided this extra blank line generation in the c backend

tool/templates/c/impl.h.jinja
```diff
@@ -1,6 +1,13 @@
 {% if is_for_cpp -%}
 extern "C" {
 {%- endif %}
+{% for cb_and_struct_def in cb_structs_and_defs %}
+typedef struct {{ cb_and_struct_def.name }} {
+    const void* data;
+    {{ cb_and_struct_def.return_type }} (*run_callback)(const void*{% if cb_and_struct_def.params_types != "" %}, {{ cb_and_struct_def.params_types }} {% endif %});
+    void (*destructor)(const void*);
+} {{ cb_and_struct_def.name }};
+{% endfor %}
```
Manishearth, line 10:
```suggestion
{%- endfor %}
```

probably this will fix your whitespace woes

core/src/hir/lowering.rs
```diff
@@ -716,6 +716,27 @@ impl<'ast> LoweringContext<'ast> {
                     .map(|(lt, m)| Borrow::new(ltl.lower_lifetime(lt), *m)),
                 PrimitiveType::from_ast(*prim),
             ))),
+            ast::TypeName::Function(input_types, out_type) => {
```
Manishearth, line 719: issue: we should validate that callbacks are allowed in the backend first

just check `!self.attr_validator.attrs_supported().callback` and return an error otherwise, like so


https://github.com/rust-diplomat/diplomat/pull/558/commits/b5a78a7775eef94e7be113e2b77b19e7709c8536

tool/templates/c/impl.h.jinja
emarteca: thank you, i was trying to track this down!

core/src/hir/lowering.rs
emarteca: good catch!

Manishearth (approved):

core/src/hir/lowering.rs
```diff
@@ -1064,10 +1096,10 @@ impl<'ast> LoweringContext<'ast> {
         ltl: &mut impl LifetimeLowerer,
         in_path: &ast::Path,
     ) -> Result<Param, ()> {
-        let name = self.lower_ident(&param.name, "param name");
-        let ty = self.lower_type::<InputOnly>(&param.ty, ltl, in_path);
+        let name = self.lower_ident(&param.name, "param name")?;
```
Manishearth, line 1099: oh, this change to the `?` shouldn't have happened, we lazily `?` so that compilation can proceed as far as possible before errors cause it to fail

PR #553, opened 2024-07-17
Fix JS2 typo

robertbastian:


robertbastian:
Merging because I need this on main to work in icu4x.

PR #572, opened 2024-07-19
Use directories for C++ namespaces

robertbastian:
https://github.com/unicode-org/icu4x/issues/5267


Manishearth (approved):
I was originally holding off on this because I didn't want to do the logic to do relative paths in the include. But you don't actually need to.


The downside of not doing relative paths is that you now _have_ to include with the `icu4x/foo.hpp` prefix, you cannot just `-I icu4x` and have that work

robertbastian:
#427

PR #593, opened 2024-07-23
Align JS `char` representation with Dart

robertbastian:
Between JS and Dart we have two different approaches to `char`:
* Dart defines a `typedef Rune = int`, "rune" being the term used by the standard library for a Unicode code point (well, [almost](https://github.com/dart-lang/sdk/issues/56304))
* JS defines a `type char = string` and then panics if the string contains more than one code point

I think the Dart representation is superior, as APIs that work with chars are usually low-level APIs where the integer-ness of the char is somewhat important.

This PR changes JS to `type char = number` (commit 1), and then to `type codepoint = number` (commit 2), as JS uses the term "code point" (e.g. `String.codePointAt`, `String.fromCodePoints`).


Manishearth (approved):
I feel like there are two purposes for returning a char, either because you want to be low level or because you want to be performant in C++/etc. This means people wishing to do the latter will need to do some diplomat gating in js, but that isn't too bad.

PR #597, opened 2024-07-25
Dart: remove views

robertbastian:
This is an indirection that is not actually needed.


Manishearth (approved):
praise: nice, this makes it a bit more compact

PR #618, opened 2024-08-05
Demo Gen Backend Customization

ambiguousname:
Includes a host of improvements that make configuring default demo_gen behavior much easier. 

Firstly, all rendering has been moved to one file, but it uses search params to select a terminus function to render.

There are also a few customization options with `-l` that allow for things like disabling the default renderer output and explicit opt-in for demo_gen output.


Manishearth (approved):
Overall lgtm!!

sffc (commented):
:heavy_check_mark:

Issue #532, opened 2024-07-08
JS2 Backend: Separate typescript header file templates


ambiguousname:
Separate .d.ts templates with shared structures like MethodInfo (to avoid generating the same information twice for each type)

PR #620, opened 2024-08-06
JS Backend: Struct Sharing to avoid generating 

ambiguousname:
Per #532. I didn't exactly want to make separate typescript template files, and the most re-use for typescript vs javascript was: struct fields, method information, and special method information. So I separated each into its own struct that we modify in a loop for JS vs. TS files.

The code I think could still use with some cleaning (especially the `gen_struct` function), but any thoughts would be appreciated.


feature_tests/js/api/MyStruct.d.ts
```diff
@@ -1,5 +1,6 @@
 // generated by diplomat-tool
 import type { MyEnum } from "./MyEnum"
+import type { MyZst } from "./MyZst"
```
robertbastian, line 3: obesrvation: I guess importing the error types don't hurt, especially if we want to reference them in documentation

robertbastian (approved):

Manishearth (approved):

Manishearth (approved):

PR #629, opened 2024-08-14
Code generation for callback support in Kotlin

emarteca:
Code generation following the pattern described in [this section of the design doc for callback support](https://docs.google.com/document/d/1sWeYZY182RQDac3UKtiWyHc4opXap4c2t4jlTECkLgA/edit?pli=1&tab=t.0#heading=h.d1qir2uyetb2).

### DIFF
--- a/feature_tests/kotlin/somelib/src/main/kotlin/dev/diplomattest/somelib/CallbackWrapper.kt
+++ b/feature_tests/kotlin/somelib/src/main/kotlin/dev/diplomattest/somelib/CallbackWrapper.kt
@@ -113,7 +113,7 @@ internal class DiplomatCallback_CallbackWrapper_test_no_args_diplomatCallback_h 
     }
 }
 internal interface Runner_DiplomatCallback_CallbackWrapper_test_cb_with_struct_diplomatCallback_f: Callback {
-    fun invoke(lang_specific_context: Pointer?, arg0: CallbackTestingStruct ): Int
+    fun invoke(lang_specific_context: Pointer?, arg0: CallbackTestingStructNative ): Int
 }
 
 internal class DiplomatCallback_CallbackWrapper_test_cb_with_struct_diplomatCallback_f_Native: Structure(), Structure.ByValue {
@@ -122,7 +122,7 @@ internal class DiplomatCallback_CallbackWrapper_test_cb_with_struct_diplomatCall
     @JvmField
     internal var run_callback: Runner_DiplomatCallback_CallbackWrapper_test_cb_with_struct_diplomatCallback_f
         = object :  Runner_DiplomatCallback_CallbackWrapper_test_cb_with_struct_diplomatCallback_f {
-                override fun invoke(lang_specific_context: Pointer?, arg0: CallbackTestingStruct ): Int {
+                override fun invoke(lang_specific_context: Pointer?, arg0: CallbackTestingStructNative ): Int {
                     throw Exception("Default callback runner -- should be replaced.")
                 }
             }
@@ -146,8 +146,8 @@ internal class DiplomatCallback_CallbackWrapper_test_cb_with_struct_diplomatCall
         
         fun fromCallback(cb: (CallbackTestingStruct)->Int): DiplomatCallback_CallbackWrapper_test_cb_with_struct_diplomatCallback_f {
             val callback: Runner_DiplomatCallback_CallbackWrapper_test_cb_with_struct_diplomatCallback_f = object :  Runner_DiplomatCallback_CallbackWrapper_test_cb_with_struct_diplomatCallback_f {
-                override fun invoke(lang_specific_context: Pointer?, arg0: CallbackTestingStruct ): Int {
-                    return cb(arg0);
+                override fun invoke(lang_specific_context: Pointer?, arg0: CallbackTestingStructNative ): Int {
+                    return cb(CallbackTestingStruct(arg0));
                 }
             }
             val cb_wrap = DiplomatCallback_CallbackWrapper_test_cb_with_struct_diplomatCallback_f_Native()
--- a/feature_tests/kotlin/somelib/src/main/kotlin/dev/diplomattest/somelib/CallbackWrapper.kt
+++ b/feature_tests/kotlin/somelib/src/main/kotlin/dev/diplomattest/somelib/CallbackWrapper.kt
@@ -69,7 +69,7 @@ internal class DiplomatCallback_CallbackWrapper_test_multi_arg_callback_diplomat
     }
 }
 internal interface Runner_DiplomatCallback_CallbackWrapper_test_no_args_diplomatCallback_h: Callback {
-    fun invoke(lang_specific_context: Pointer?): Void
+    fun invoke(lang_specific_context: Pointer?): Unit
 }
 
 internal class DiplomatCallback_CallbackWrapper_test_no_args_diplomatCallback_h_Native: Structure(), Structure.ByValue {
@@ -78,7 +78,7 @@ internal class DiplomatCallback_CallbackWrapper_test_no_args_diplomatCallback_h_
     @JvmField
     internal var run_callback: Runner_DiplomatCallback_CallbackWrapper_test_no_args_diplomatCallback_h
         = object :  Runner_DiplomatCallback_CallbackWrapper_test_no_args_diplomatCallback_h {
-                override fun invoke(lang_specific_context: Pointer?): Void {
+                override fun invoke(lang_specific_context: Pointer?): Unit {
                     throw Exception("Default callback runner -- should be replaced.")
                 }
             }
@@ -100,9 +100,9 @@ internal class DiplomatCallback_CallbackWrapper_test_no_args_diplomatCallback_h 
     companion object {
         val NATIVESIZE: Long = Native.getNativeSize(DiplomatCallback_CallbackWrapper_test_no_args_diplomatCallback_h_Native::class.java).toLong()
         
-        fun fromCallback(cb: ()->Void): DiplomatCallback_CallbackWrapper_test_no_args_diplomatCallback_h {
+        fun fromCallback(cb: ()->Unit): DiplomatCallback_CallbackWrapper_test_no_args_diplomatCallback_h {
             val callback: Runner_DiplomatCallback_CallbackWrapper_test_no_args_diplomatCallback_h = object :  Runner_DiplomatCallback_CallbackWrapper_test_no_args_diplomatCallback_h {
-                override fun invoke(lang_specific_context: Pointer?): Void {
+                override fun invoke(lang_specific_context: Pointer?): Unit {
                     return cb();
                 }
             }
@@ -259,7 +259,7 @@ class CallbackWrapper internal constructor (
             return returnVal
         }
         
-        fun testNoArgs(h: ()->Void): Int {
+        fun testNoArgs(h: ()->Unit): Int {
             
             val returnVal = lib.CallbackWrapper_test_no_args(DiplomatCallback_CallbackWrapper_test_no_args_diplomatCallback_h.fromCallback(h).nativeStruct);
             return returnVal
--- a/feature_tests/kotlin/somelib/src/main/kotlin/dev/diplomattest/somelib/CallbackWrapper.kt
+++ b/feature_tests/kotlin/somelib/src/main/kotlin/dev/diplomattest/somelib/CallbackWrapper.kt
@@ -25,17 +25,17 @@ internal class CallbackWrapperNative: Structure(), Structure.ByValue {
 
 
 internal interface Runner_DiplomatCallback_CallbackWrapper_test_multi_arg_callback_diplomatCallback_f: Callback {
-    fun invoke(ignored: Pointer?, arg0: Int ): Int
+    fun invoke(lang_specific_context: Pointer?, arg0: Int ): Int
 }
 
-class DiplomatCallback_CallbackWrapper_test_multi_arg_callback_diplomatCallback_f_Native: Structure(), Structure.ByValue {
+internal class DiplomatCallback_CallbackWrapper_test_multi_arg_callback_diplomatCallback_f_Native: Structure(), Structure.ByValue {
     @JvmField
-    internal var dataa: Pointer = Pointer(0L);
+    internal var data_: Pointer = Pointer(0L);
     @JvmField
     internal var run_callback: Runner_DiplomatCallback_CallbackWrapper_test_multi_arg_callback_diplomatCallback_f
         = object :  Runner_DiplomatCallback_CallbackWrapper_test_multi_arg_callback_diplomatCallback_f {
-                override fun invoke(ignored: Pointer?, arg0: Int ): Int {
-                    throw Exception("ERROR NOT IMPLEMENTED")
+                override fun invoke(lang_specific_context: Pointer?, arg0: Int ): Int {
+                    throw Exception("Default callback runner -- should be replaced.")
                 }
             }
     @JvmField
@@ -43,13 +43,13 @@ class DiplomatCallback_CallbackWrapper_test_multi_arg_callback_diplomatCallback_
 
     // Define the fields of the struct
     override fun getFieldOrder(): List<String> {
-        return listOf("dataa", "run_callback", "destructor")
+        return listOf("data_", "run_callback", "destructor")
     }
 }
 
-class DiplomatCallback_CallbackWrapper_test_multi_arg_callback_diplomatCallback_f internal constructor (
+internal class DiplomatCallback_CallbackWrapper_test_multi_arg_callback_diplomatCallback_f internal constructor (
     internal val nativeStruct: DiplomatCallback_CallbackWrapper_test_multi_arg_callback_diplomatCallback_f_Native) {
-    val dataa: Pointer = nativeStruct.dataa
+    val data_: Pointer = nativeStruct.data_
     val run_callback: Callback = nativeStruct.run_callback
     val destructor: Pointer = nativeStruct.destructor
 
@@ -58,7 +58,7 @@ class DiplomatCallback_CallbackWrapper_test_multi_arg_callback_diplomatCallback_
         
         fun fromCallback(cb: (Int)->Int): DiplomatCallback_CallbackWrapper_test_multi_arg_callback_diplomatCallback_f {
             val callback: Runner_DiplomatCallback_CallbackWrapper_test_multi_arg_callback_diplomatCallback_f = object :  Runner_DiplomatCallback_CallbackWrapper_test_multi_arg_callback_diplomatCallback_f {
-                override fun invoke(ignored: Pointer?, arg0: Int ): Int {
+                override fun invoke(lang_specific_context: Pointer?, arg0: Int ): Int {
                     return cb(arg0);
                 }
             }
@@ -69,17 +69,17 @@ class DiplomatCallback_CallbackWrapper_test_multi_arg_callback_diplomatCallback_
     }
 }
 internal interface Runner_DiplomatCallback_CallbackWrapper_test_no_args_diplomatCallback_h: Callback {
-    fun invoke(ignored: Pointer?): Void
+    fun invoke(lang_specific_context: Pointer?): Void
 }
 
-class DiplomatCallback_CallbackWrapper_test_no_args_diplomatCallback_h_Native: Structure(), Structure.ByValue {
+internal class DiplomatCallback_CallbackWrapper_test_no_args_diplomatCallback_h_Native: Structure(), Structure.ByValue {
     @JvmField
-    internal var dataa: Pointer = Pointer(0L);
+    internal var data_: Pointer = Pointer(0L);
     @JvmField
     internal var run_callback: Runner_DiplomatCallback_CallbackWrapper_test_no_args_diplomatCallback_h
         = object :  Runner_DiplomatCallback_CallbackWrapper_test_no_args_diplomatCallback_h {
-                override fun invoke(ignored: Pointer?): Void {
-                    throw Exception("ERROR NOT IMPLEMENTED")
+                override fun invoke(lang_specific_context: Pointer?): Void {
+                    throw Exception("Default callback runner -- should be replaced.")
                 }
             }
     @JvmField
@@ -87,13 +87,13 @@ class DiplomatCallback_CallbackWrapper_test_no_args_diplomatCallback_h_Native: S
 
     // Define the fields of the struct
     override fun getFieldOrder(): List<String> {
-        return listOf("dataa", "run_callback", "destructor")
+        return listOf("data_", "run_callback", "destructor")
     }
 }
 
-class DiplomatCallback_CallbackWrapper_test_no_args_diplomatCallback_h internal constructor (
+internal class DiplomatCallback_CallbackWrapper_test_no_args_diplomatCallback_h internal constructor (
     internal val nativeStruct: DiplomatCallback_CallbackWrapper_test_no_args_diplomatCallback_h_Native) {
-    val dataa: Pointer = nativeStruct.dataa
+    val data_: Pointer = nativeStruct.data_
     val run_callback: Callback = nativeStruct.run_callback
     val destructor: Pointer = nativeStruct.destructor
 
@@ -102,7 +102,7 @@ class DiplomatCallback_CallbackWrapper_test_no_args_diplomatCallback_h internal 
         
         fun fromCallback(cb: ()->Void): DiplomatCallback_CallbackWrapper_test_no_args_diplomatCallback_h {
             val callback: Runner_DiplomatCallback_CallbackWrapper_test_no_args_diplomatCallback_h = object :  Runner_DiplomatCallback_CallbackWrapper_test_no_args_diplomatCallback_h {
-                override fun invoke(ignored: Pointer?): Void {
+                override fun invoke(lang_specific_context: Pointer?): Void {
                     return cb();
                 }
             }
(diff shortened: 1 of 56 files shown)


emarteca:
+ @jcrist1

Manishearth:
I plan to review this soonish, but @jcrist1 should let me know what their expectations are for their own review: do you want us to wait on landing this until you've also reviewed, or okay with landing with just my review, or some other thing?

(keeping in mind i understand very little about Kotlin)

jcrist1:
Skimming over it quickly it doesn't look like it will take long to review. I'd like to try and do a review tonight still. But if I don't get one in by tomorrow feel free to move forward with it.

Manishearth:
sounds good, seems like we have roughly the same timeline for reviewing this. No need to rush if you can't get to it, I mostly wanted to make sure we communicated review needs and timelines

feature_tests/kotlin/somelib/src/main/kotlin/dev/diplomattest/somelib/CallbackWrapper.kt
```diff
@@ -0,0 +1,281 @@
...
+}
+
+class DiplomatCallback_CallbackWrapper_test_multi_arg_callback_diplomatCallback_f_Native: Structure(), Structure.ByValue {
+    @JvmField
+    internal var dataa: Pointer = Pointer(0L);
+    @JvmField
+    internal var run_callback: Runner_DiplomatCallback_CallbackWrapper_test_multi_arg_callback_diplomatCallback_f
+        = object :  Runner_DiplomatCallback_CallbackWrapper_test_multi_arg_callback_diplomatCallback_f {
+                override fun invoke(ignored: Pointer?, arg0: Int ): Int {
+                    throw Exception("ERROR NOT IMPLEMENTED")
```
Manishearth, line 38: question: why this? is this a default value?

(perhaps worth improving the exception text)

tool/templates/kotlin/Struct.kt.jinja
```diff
@@ -27,6 +28,55 @@ internal class {{type_name}}Native: Structure(), Structure.ByValue {
 
 {% endif -%}
 
+{% if !callback_params.is_empty() -%}
```
Manishearth, line 31: issue: callbacks can exist on structs, enums, or opaques, we should support all three.

you may wish to use an included template to make it easy

feature_tests/kotlin/somelib/src/main/kotlin/dev/diplomattest/somelib/CallbackWrapper.kt
```diff
@@ -0,0 +1,281 @@
...
+class CallbackWrapper internal constructor (
+    internal val nativeStruct: CallbackWrapperNative) {
+    val cantBeEmpty: Boolean = nativeStruct.cantBeEmpty > 0
+
+    companion object {
+        internal val libClass: Class<CallbackWrapperLib> = CallbackWrapperLib::class.java
+        internal val lib: CallbackWrapperLib = Native.load("somelib", libClass)
+        val NATIVESIZE: Long = Native.getNativeSize(CallbackWrapperNative::class.java).toLong()
+        
+        fun testMultiArgCallback(f: DiplomatCallback_CallbackWrapper_test_multi_arg_callback_diplomatCallback_f, x: Int): Int {
```
Manishearth, line 256: issue: I think this should take `fn(Int) -> Int` and internally call fromCallback?

Manishearth: We always want the user-facing APIs to have nice idiomatic types

also this means that the diplomatCallback can be private

jcrist1 (commented):
First it looks like the code gen is broken but that has nothing to do with your PR. I'll look into it in a bit to see how much effort it would be to fix. 

Otherwise it looks super cool! I manually cleaned up the generated kotlin code to get it compiling, and created a little test:
```kotlin
package dev.diplomattest.somelib

import org.junit.jupiter.api.Assertions.assertEquals
import org.junit.jupiter.api.Test

class CallbackTest {
    @Test
    fun testCallback() {
        val cb: (Int ) -> Int = { i -> i + 2 }
        val calledBack = CallbackWrapper.testMultiArgCallback(
            DiplomatCallback_CallbackWrapper_test_multi_arg_callback_diplomatCallback_f.fromCallback(cb),
            10
        )
        assertEquals(22, calledBack)
    }
}
```
which passed. It would be nice to add it to the main code so in principle it would be tested with `cargo make test-kotlin-feature`. 

I think, however, it highlights how it is a bit unergonomic in its current form. It would be nice if the `CallbackWrapper.testMultiArgCallback` method could directly take the cb without having to wrap it. WDYT? I don't think this would be that difficult to change as you could just move the kotlin code gen around a little bit. 

I'll try to have a more careful look at the generated jna code tomorrow, but would also be happy if we just test out a bunch of the generated callback consuming methods.

tool/templates/kotlin/Struct.kt.jinja
```diff
@@ -27,6 +28,55 @@ internal class {{type_name}}Native: Structure(), Structure.ByValue {
...
+    {%- for callback_param in callback_params %}
+internal interface Runner_{{callback_param.name}}: Callback {
+    fun invoke(ignored: Pointer?{% if callback_param.input_params_and_types != "" %}, {{callback_param.input_params_and_types}} {% endif %}): {{callback_param.output_type}}
+}
+
+class {{callback_param.name}}_Native: Structure(), Structure.ByValue {
+    @JvmField
+    internal var dataa: Pointer = Pointer(0L);
```
jcrist1, line 39: Is this a typo?
```suggestion
    internal var data: Pointer = Pointer(0L);
```

tool/src/kotlin/snapshots/diplomat_tool__kotlin__test__struct.snap
```diff
@@ -49,6 +52,51 @@ internal class MyNativeStructNative: Structure(), Structure.ByValue {
...
+    @JvmField
+    internal var destructor: Pointer = Pointer(0L);
+
+    // Define the fields of the struct
+    override fun getFieldOrder(): List<String> {
+        return listOf("dataa", "run_callback", "destructor")
+    }
+}
+
+class DiplomatCallback_MyNativeStruct_test_multi_arg_callback_diplomatCallback_f internal constructor (
```
jcrist1, line 79: If these are supposed to be public then the combined snake/Pascal case feels unidiomatic. If, however we adapted it so this wrapper is only instantiated inside the `CallbackWrapper.testMultiArgCallback` method body, then I think we wouldn't need to care.

jcrist1 (commented):
Actually had a couple of code comments that I thought I had saved, but hadn't. So quickly added them.

tool/src/kotlin/mod.rs
```diff
@@ -861,6 +869,8 @@ retutnVal.option() ?: return null
         special_methods: &mut SpecialMethods,
         method: &'cx hir::Method,
         self_type: Option<&'cx SelfType>,
+        struct_name: Option<&str>,
+        callback_params: &mut Vec<CallbackParamInfo>,
```
Manishearth, line 873: thought: these callback params could also be collected on the TyGenContext, probably

feature_tests/kotlin/somelib/src/main/kotlin/dev/diplomattest/somelib/CallbackWrapper.kt
```diff
@@ -0,0 +1,281 @@
...
+    val dataa: Pointer = nativeStruct.dataa
+    val run_callback: Callback = nativeStruct.run_callback
+    val destructor: Pointer = nativeStruct.destructor
+
+    companion object {
+        val NATIVESIZE: Long = Native.getNativeSize(DiplomatCallback_CallbackWrapper_test_multi_arg_callback_diplomatCallback_f_Native::class.java).toLong()
+        
+        fun fromCallback(cb: (Int)->Int): DiplomatCallback_CallbackWrapper_test_multi_arg_callback_diplomatCallback_f {
+            val callback: Runner_DiplomatCallback_CallbackWrapper_test_multi_arg_callback_diplomatCallback_f = object :  Runner_DiplomatCallback_CallbackWrapper_test_multi_arg_callback_diplomatCallback_f {
+                override fun invoke(ignored: Pointer?, arg0: Int ): Int {
```
Manishearth, line 61: question: what's the deal with the "ignored"?

feature_tests/kotlin/somelib/src/main/kotlin/dev/diplomattest/somelib/CallbackWrapper.kt
```diff
@@ -0,0 +1,281 @@
...
+        val NATIVESIZE: Long = Native.getNativeSize(DiplomatCallback_CallbackWrapper_test_multiple_cb_args_diplomatCallback_g_Native::class.java).toLong()
+        
+        fun fromCallback(cb: (Int)->Int): DiplomatCallback_CallbackWrapper_test_multiple_cb_args_diplomatCallback_g {
+            val callback: Runner_DiplomatCallback_CallbackWrapper_test_multiple_cb_args_diplomatCallback_g = object :  Runner_DiplomatCallback_CallbackWrapper_test_multiple_cb_args_diplomatCallback_g {
+                override fun invoke(ignored: Pointer?, arg0: Int ): Int {
+                    return cb(arg0);
+                }
+            }
+            val cb_wrap = DiplomatCallback_CallbackWrapper_test_multiple_cb_args_diplomatCallback_g_Native()
+            cb_wrap.run_callback = callback;
```
Manishearth, line 242: question: where is the data/dtor getting set here?

tool/templates/kotlin/Struct.kt.jinja
emarteca: I found that `data` is a keyword in Kotlin, so I just added an "a" :sweat_smile: 
Perhaps I should rename it to some other word -- I was trying to preserve the same naming structure as the macro-side callback struct, that does have `data` as a field.

tool/src/kotlin/snapshots/diplomat_tool__kotlin__test__struct.snap
emarteca: Good point -- I'll make them internal, this goes along well with your comments about making the usage more ergonomic and taking the callback itself as an argument.

tool/templates/kotlin/Struct.kt.jinja
Manishearth: ah! i'd maybe call it `data_`?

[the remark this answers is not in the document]
emarteca: @jcrist1 noticed that too -- I found that `data` is a keyword in Kotlin, so I just added an "a". Could rename it to another word, but then it wouldn't match the callback struct on the macro side

feature_tests/kotlin/somelib/src/main/kotlin/dev/diplomattest/somelib/CallbackWrapper.kt
emarteca: Good idea! Yes, it is a default value -- I'll make this clear in the exception string.

feature_tests/kotlin/somelib/src/main/kotlin/dev/diplomattest/somelib/CallbackWrapper.kt
emarteca: In the previous line it's calling the internal constructor on the `Native` callback wrapper, which has the default value for `run_callback`, and a 0 pointer for `dataa`. The destructor is also set to a 0 pointer -- it won't do anything on the Rust side, since I think we decided Kotlin will do all the memory management for the callback structs. The direct assignment to `run_callback` here is overwriting the default value for this field.

feature_tests/kotlin/somelib/src/main/kotlin/dev/diplomattest/somelib/CallbackWrapper.kt
emarteca: Ah, so the `run_callback` signature on the Rust side is `unsafe extern "C" fn(*const c_void, ...) -> ReturnType` -- the first argument is a pointer that could contain language-specific information/context, but for Kotlin none is needed so it's ignored. Maybe it should have a different though?

tool/templates/kotlin/Struct.kt.jinja
emarteca: Oh, good point. I didn't think of that. I'll add tracking for those other custom types too.

tool/templates/kotlin/Struct.kt.jinja
emarteca: oh, good idea! will do

emarteca:
Thanks for the comments! I've pushed some fixes, the main one being refactoring so that a user can directly pass the callback in and the `.fromCallback` is called internally. I still need to fix the tests to reflect this change, and to add support for callbacks in methods on other custom types (enums, opaques).

feature_tests/kotlin/somelib/src/main/kotlin/dev/diplomattest/somelib/CallbackWrapper.kt
Manishearth: yeah let's call it lang_specific_context or something

feature_tests/kotlin/somelib/src/main/kotlin/dev/diplomattest/somelib/CallbackWrapper.kt
Manishearth: so wait, how does Rust tell Kotlin that it's no longer holding on to the callback?

I might be missing something in the design here.

emarteca:
@jcrist1 to fun the tests, should it just be `cargo make gen-kotlin-feature && cargo make test-kotlin-feature` ? This is still failing for me.

tool/src/kotlin/mod.rs
```diff
@@ -1545,6 +1549,36 @@ retutnVal.option() ?: return null
             _ => unreachable!("unknown AST/HIR variant"),
         }
     }
+
+    fn gen_non_wrapped_type_name(
```
Manishearth, line 1553: nit: docs on what "non wrapped" is here

jcrist1:
> @jcrist1 to run the tests, should it just be `cargo make gen-kotlin-feature && cargo make test-kotlin-feature` ? This is still failing for me. The first error is because the duckscript is trying to copy a `dylib` file which is not generated, and when I change `dylib` to `so` (since this is generated), the `cargo make test-kotlin-feature` fails with many Kotlin compilation errors.

`dylib` to `so` makes sense. I think the cargo make test-kotlin-feature doesn't run on CI so that wouldn't have been caught, and I only wrote it thinking about my mac.

As for the compile error did you merge the latest main? I had to fix some of the kotlin code gen that fell out of sync in https://github.com/rust-diplomat/diplomat/pull/631 If not, then give that a try. Regardless copy the output of the error, and I'll see if I can figure out why it's failing

jcrist1:
@emarteca pinging, cause I neglected to ping in the actual message ☝️

Manishearth:
@jcrist1 yeah, we really should get that CI setup

jcrist1:
@Manishearth I can look into adding a test-kotlin tomorrow to the ci step. Would need to install java and gradle, but it looks pretty straightforward. https://docs.github.com/en/actions/use-cases-and-examples/building-and-testing/building-and-testing-java-with-gradle.

feature_tests/kotlin/somelib/src/main/kotlin/dev/diplomattest/somelib/CallbackWrapper.kt
jcrist1: So I wasn't paying that much attention to the other PR but the callback is expanded into 
```rust
        #[no_mangle]
        extern "C" fn CallbackWrapper_test_multiple_cb_args(
            f: DiplomatCallback<i32>,
            g: DiplomatCallback<i32>,
        ) -> i32 {
            let f = move || unsafe {
                std::mem::transmute::<
                    unsafe extern "C" fn(*const c_void, ...) -> i32,
                    unsafe extern "C" fn(*const c_void) -> i32,
                >(f.run_callback)(f.data)
            };
            let g = move |arg0: i32| unsafe {
                std::mem::transmute::<
                    unsafe extern "C" fn(*const c_void, ...) -> i32,
                    unsafe extern "C" fn(*const c_void, i32) -> i32,
                >(g.run_callback)(g.data, arg0)
            };
            CallbackWrapper::test_multiple_cb_args(f, g)
        }
```
So the impl is only ever monomorphised into the closure `g`. As I understand it `g.data` and `g.run_callback` are copied into the closure and the closure is dropped after the call. But because they aren't owned pointers nothing is done to the backing data. And Kotlin will always wait until the method `CallbackWrapper.testMultipleCbArgs` is finished before it can GC the callback. 

I expect there would be a problem if a struct can have a callback as a field, or if we could wrap a callback in an opaque but this at least fails to compile because of no lifetime bounds on the `impl Fn`:
```
    #[diplomat::opaque]
    struct BadThing(Box<dyn Fn(i32) -> i32>);
    impl CallbackWrapper {
        ...
        pub fn test_opaque_callback(f: impl Fn(i32) -> i32) -> Box<BadThing> {
            Box::new(BadThing(Box::new(f)))
        }
    }
```

jcrist1: But if we want callbacks as params or in opaques, then we need lifetime bounds afaics, and to add the callback struct to the edges of the the object.

emarteca: Good points -- indeed (as discussed offline too) we need to have Rust store a GC root to the callback wrapper object so that the JVM doesn't get rid of it until it's dropped on the Rust side too. We'll do this in a subsequent PR, but basically the mechanism will be to have the `data` field store a boxed JNI `GlobalRef` that references the callback wrapper object. Then in the `Drop`, Rust will drop the `GlobalRef`.

emarteca:
I just pushed callback param support for opaques and enums -- the only thing left (I think) is to fix the tests! 
I'm going to make a separate PR with the proper drop setup.

jcrist1:
> @jcrist1 thanks for taking a look! Here's the errors: this is after running `cargo make gen-kotlin-feature && cargo make test-kotlin-feature`: https://pastebin.com/EUEAegEu
> 
> I did merge the newest updates from `main` first! Also FWIW I am using linux, so maybe there is an OS difference causing these issues

@emarteca can you paste the output of `gradle --version`? I'm expecting an incorrect java version. I tested on java 19.

jcrist1:
I've got a linux based ci pipeline up here: https://github.com/rust-diplomat/diplomat/pull/638

emarteca:
> > @jcrist1 thanks for taking a look! Here's the errors: this is after running `cargo make gen-kotlin-feature && cargo make test-kotlin-feature`: https://pastebin.com/EUEAegEu
> > I did merge the newest updates from `main` first! Also FWIW I am using linux, so maybe there is an OS difference causing these issues
> 
> @emarteca can you paste the output of `gradle --version`? I'm expecting an incorrect java version. I tested on java 19.

Thanks! Yes, you're right -- I'm using Java 17, and gradle 8.3. The output of `gradle --version` is below:
```
------------------------------------------------------------
Gradle 8.3
------------------------------------------------------------

Build time:   2023-08-17 07:06:47 UTC
Revision:     8afbf24b469158b714b36e84c6f4d4976c86fcd5

Kotlin:       1.9.0
Groovy:       3.0.17
Ant:          Apache Ant(TM) version 1.10.13 compiled on January 4 2023
JVM:          17.0.10 (Debian 17.0.10+7-Debian-1)
OS:           Linux 6.7.12-1rodete1-amd64 amd64
```
I'll get back to you on whether we need to get this working on Java 17 or if we can upgrade.

jcrist1:
> > > @jcrist1 thanks for taking a look! Here's the errors: this is after running `cargo make gen-kotlin-feature && cargo make test-kotlin-feature`: https://pastebin.com/EUEAegEu
> > > I did merge the newest updates from `main` first! Also FWIW I am using linux, so maybe there is an OS difference causing these issues
> > 
> > 
> > @emarteca can you paste the output of `gradle --version`? I'm expecting an incorrect java version. I tested on java 19.
> 
> Thanks! Yes, you're right -- I'm using Java 17, and gradle 8.3. The output of `gradle --version` is below:
> 
> ```
> ------------------------------------------------------------
> Gradle 8.3
> ------------------------------------------------------------
> 
> Build time:   2023-08-17 07:06:47 UTC
> Revision:     8afbf24b469158b714b36e84c6f4d4976c86fcd5
> 
> Kotlin:       1.9.0
> Groovy:       3.0.17
> Ant:          Apache Ant(TM) version 1.10.13 compiled on January 4 2023
> JVM:          17.0.10 (Debian 17.0.10+7-Debian-1)
> OS:           Linux 6.7.12-1rodete1-amd64 amd64
> ```
> 
> I'll get back to you on whether we need to get this working on Java 17 or if we can upgrade.

I managed to test with Java 17 in IntelliJ, and there were a couple of panama imports that I apparently have in one one test, that broke the compile. But it managed to compile once I deleted the imports. I can have a look at downgrading the supported java version.

jcrist1:
and once again cause I never remember to ping: @emarteca ☝️

emarteca:
@jcrist1 thanks!! It turns out I do need to keep this compatible with Java 17. Does the test work in newer Java without the imports? 
I think ideally, we'd be able to support various versions of Java; but if we're going to pick one it probably makes sense to stick to the newer version. If you give me the hack to fix the build for old Java then I can test it manually, and just make sure the callback tests work on your end with the newer Java?

jcrist1:
@emarteca there should be no issue with JVM 19 support if this works with JVM 17. It was just some accidental code that was I think I accidentally added when I started working on #499.

jcrist1:
I pushed some changes to https://github.com/rust-diplomat/diplomat/pull/638 that should make it compatible with JVM 17

Manishearth (approved):
overall lgtm, please let me know when this is in a mergeable state

I believe dtors still need to be fixed but that is going to be done later

jcrist1:
@emarteca the kotlin ci changes that should support jvm 17 were added. Can you try merging main and see if it works?

emarteca:
Thanks @jcrist1 , it does compile now!! :pray:  I have to fix a bug that cropped up (the generated callback code is showing up in all of the Kotlin files, not just the file for the struct that includes the callbacks), and then I'll ping you!

feature_tests/kotlin/somelib/src/test/kotlin/dev/diplomattest/somelib/CallbackWrapperTest.kt
```diff
@@ -0,0 +1,35 @@
...
+    fun testNoArgCallback() {
+        val cb: () -> Unit = { println("Hello this is a callback"); }
+        val calledBack = CallbackWrapper.testNoArgs(cb)
+        assertEquals(-5, calledBack)
+    }
+
+    @Test
+    fun testCbWithStructCallback() {
+        val cb: (CallbackTestingStruct) -> Int = { s -> s.x + s.y}
+        val calledBack = CallbackWrapper.testCbWithStruct(cb)
```
emarteca, line 24: @jcrist1  this test is broken with the error `java.lang.IllegalArgumentException: Callback argument class dev.diplomattest.somelib.CallbackTestingStruct requires custom type conversion` -- do you know how to fix this? Is this a bug, or is this a Kotlin feature?

emarteca:
@jcrist1 my code gen is working again, and I made a `feature_test` test for it! The only thing is that the test for the callback that takes a custom struct is broken -- could you maybe take a look, I am not well versed in Kotlin! (this comment: [linked](https://github.com/rust-diplomat/diplomat/pull/629/files#r1722481649))

jcrist1:
> I am not well versed in Kotlin!

😂 sure I can try, but it's actually the same story for me (my background is scala). 
image
Will have a look tonight [CEST](https://en.wikipedia.org/wiki/Central_European_Summer_Time).

emarteca:
> > I am not well versed in Kotlin!
> 
> 😂 sure I can try, but it's actually the same story for me (my background is scala). image Will have a look tonight [CEST](https://en.wikipedia.org/wiki/Central_European_Summer_Time).

ahah i had thought you were an expert! But no worries, I'll look at it again today too :-)

Manishearth:
For passing the failing tests, rerun tests in that crate and then run `cargo insta accept` (you'll need to install `insta`)

You may need to do this multiple times

Manishearth:
(I forget if you'd figured out how to do this already, but in case this is the first time you're seeing this, worth explaining)

Manishearth:
I think this is more or less ready to land.

emarteca:
> For passing the failing tests, rerun tests in that crate and then run `cargo insta accept` (you'll need to install `insta`)
> 
> You may need to do this multiple times

oh thanks, I didn't know about this

emarteca:
> I think this is more or less ready to land.

sweet! that sounds good -- i think there might be an issue with the struct argument callbacks (that's the type conversion message i was debugging with @jcrist1 ), but I can also comment out that test for now and figure that out in a subsequent PR. The other types of arguments work.

emarteca:
Ok I figured out the issue with struct arguments -- just going to push a fix, and then we should be good to go with this PR

Manishearth (approved):

Manishearth:
Still needs a regen

jcrist1 (approved):
@emarteca I assume your comment

> Ok I figured out the issue with struct arguments -- just going to push a fix, and then we should be good to go with this PR

means you don't need any help debugging and all the tests are passing so I'm approving

means that no more help is needed

Manishearth:
Merged. The dtors / GlobalRef stuff are still a todo

emarteca:
Thank you!! :pray:

PR #638, opened 2024-08-16
kotlin-ci

jcrist1:
~Unfortunately I can't really debug the ci process other than trying to run them via PR, so I'm opening this up as a draft.~

I've created a linux backed ci pipeline for kotlin that tests both functionalities. I've tried to keep the ci pipeline working locally for macos but I haven't set up ci pipeline for macos. One thing I don't really know anything about is how to test this on windows. Otherwise this feels pretty minimal. Another downside is that loading gradle takes a while. I wonder if we can cache spinning up the gradle daemon, but somehow that seems unlikely.


jcrist1:
@robertbastian  @Manishearth I've added a ci pipeline for kotlin. Linux only for now. Should I create one for Mac too? Windows will be a bit of a challenge cause I don't really know how to work on it. But if necessary I can figure something out.

Manishearth (approved):

jcrist1:
@Manishearth I'd like to keep this open a bit to see if I can get it to work on JVM 17

.github/workflows/ci.yml
```diff
@@ -178,3 +178,19 @@ jobs:
...
+  test-kotlin:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - name: Install cargo-make
+      uses: taiki-e/install-action@cargo-make
+    - name: Set up JDK 17 for x64
+      uses: actions/setup-java@v4
+      with:
+        java-version: '17'
```
jcrist1, line 190: As far as I can see @emarteca this now uses a JVM 17 version. I've also locally configured a temurin 17 JVM in my intelliJ and successfully ran the tests. So it should be good to go

feature_tests/kotlin/somelib/src/test/kotlin/dev/diplomattest/somelib/NativeTest.kt
```diff
@@ -1,8 +1,6 @@
 package dev.diplomattest.somelib
 
 import org.junit.jupiter.api.Test
-import java.lang.foreign.MemorySession
-import java.lang.foreign.SegmentAllocator
```
jcrist1, line 5: @emarteca the offending lines, if you want to try and remove these, maybe it'll work.

jcrist1:
@Manishearth I'm re-requesting review cause I made a few small changes to make this compatible with JVM 19

Manishearth (approved):

PR #646, opened 2024-08-19
Fixing wrong primitive slice constructor

ambiguousname:
My mistake, just called it the wrong thing in `converter.rs`


Manishearth (commented):
lg

PR #673, opened 2024-09-03
Demo Gen support for DiplomatOption

ambiguousname:
Because we currently don't pass context when we recursively evaluate parameters, DiplomatOption effectively is ignored for the purposes of demonstration.

Especially difficult to resolve since we flatten all struct and opaque construction parameters into one render terminus (making it hard to label what belongs where and thus set up optional support), so this is a workaround for now.

Should help with the ICU4X PR for `DiplomatOption<>`


Manishearth (approved):
Sounds like we'll need proper support for specifying Nones for both optional Opaques and DiplomatOptions.

PR #677, opened 2024-09-04
Fix Dart default ctor generation code around Options

robertbastian:
#676


robertbastian:
I'm also going to preemptively merge this since it is broken.

Manishearth (approved):

PR #692, opened 2024-09-17
Fixing code generation for `Result<bool>` in Kotlin

emarteca:
The general case for dealing with `Result` in Kotlin doesn't work for booleans.
Previously, generating the Kotlin code for a Rust function that returned a `Result<bool>` would be as follows:

Rust
```
pub fn test() -> Result<bool, ()> {
    todo!()            
}
```

Generated Kotlin (with the bug specified in a comment):
```
    fun test(): Res<Boolean, Unit> {
          val returnVal = lib.Why_test();
          if (returnVal.isOk == 1.toByte()) {
              return returnVal.union.ok.ok() > 0 // THIS IS A BUG THIS IS INVALID
          } else {
              return Err(Unit)
          }
      }
```

For booleans, this error line should instead be `return (returnVal.union.ok > 0).ok()` . This CL fixes this problem.


emarteca:
@jcrist1

emarteca:
oh hang on, the same bug shows up for some other primitive types too. let me fix them all in this PR

Manishearth (approved):

PR #693, opened 2024-09-17
Fixing bug in `Result<T>` when `T` is a primitive that requires modification in the Kotlin code generation

emarteca:
The modification applied to primitives when returned as a result were specified in the wrong order in the Kotlin code generation. 

For example, the following Rust:
```
pub fn test() -> Result<u32, ()> {
    todo!()
}
```

Generated the following Kotlin, with the bug specified with a comment:
```
      fun test(): Res<UInt, Unit> {
            val returnVal = lib.Why_test();
            if (returnVal.isOk == 1.toByte()) {
                return returnVal.union.ok.ok().toUInt() // THIS IS A BUG
            } else {
                return Err(Unit)
            }
        }
```

The buggy line should actually be `return (returnVal.union.ok.toUInt()).ok()` . This CL fixes the bug, for `u32` and for all other primitives that require modification on return as a `Res` in Kotlin.


emarteca:
@jcrist1

PR #722, opened 2024-11-04
Fix primitive slice js

jcrist1:
This PR attempts to address #699. According to the generated `ts` files primitive slices are returned as an `Array<number>` on the js side, but the generated js code looks like it's trying to return types of `Float64Array`. I personally think it makes more sense to return typed arrays which have a perfectly ergonomic API and allow the caller to decide what to do with, but don't really want to challenge the existing implementation. Just checking that I understand it correctly. I had to add an `Array.from` wrap to get that to work. However, it appears there is a bug in the actual interpretation of the raw bytes that causes the first byte to not be interpreted correctly, I tried to get to the bottom of it, but couldn't quickly figure it out, so I thought I'd open the PR, and solicit help. Does anyone know what might be going on, and why the wasm backed array seems to be incorrect?


jcrist1:
I'm going to open this up as ready for review despite the failing test, in the hope that a reviewer can point me out to where the byte interpretation goes wrong. I got as far as seeing that, the value `  -   2.26666497653787e-308,` seems to come from the uninitialized bytes, but as far as I could see the bytes get correctly written to when the array is passed to the native call.

jcrist1:
Okay I think I realised the issue. The method `newFromOwned` creates a slice and which is the backing `Vec` of the `Float64Vec`, but we free this slice as part of the method call, in
```js
...
    } finally {
      functionCleanupArena.free();
    }
```
I didn't actually check if the js backend supports owned slices in parameters. I've got a better test now.

feature_tests/js/test/slices.mjs
```diff
@@ -9,4 +9,4 @@ test("MyString functionality", (t) => {
 test("String List", (t) => {
 	let str = MyString.newFromFirst(["This", "is", "a", "test."]);
 	t.is(str.str, "This");
-});
\ No newline at end of file
+});
```
ambiguousname, line 12: I don't think there's a good reason for why we do this, but tests are duplicated across all files. So we'd add:

```suggestion
});

test("Float64Vec", (t) => {
    let input = [1, 2, 3, 4, 5];
    let data = Float64Vec.newIsize(input);
    t.deepEqual(data.borrow(), input);
});
```

I think though for the long term we should really remove all the `.mjs` files that are not directly generated from typescript.

ambiguousname (approved):
Generally looks good to me!

Thanks for adding new tests as well, very much appreciated.

PR #727, opened 2024-11-08
fix js str decode

jcrist1:
I uncovered another bug in the js backend, while I was working on a library using diplomat. 
The bug is that methods returning string slices return `DiplomatStrSlice` class instead of a js `string`. It's the same issue that affected #722 , and the fix is the same.

### DIFF
--- a/feature_tests/js/test/slices.mjs
+++ b/feature_tests/js/test/slices.mjs
@@ -1,18 +1,23 @@
-import test from 'ava';
-import { MyString, Float64Vec} from "diplomat-wasm-js-feature-tests";
+import test from "ava";
+import { MyString, Float64Vec } from "diplomat-wasm-js-feature-tests";
 
 test("MyString functionality", (t) => {
-    let str = MyString.new_("This is a test value.");
-    t.is(str.str, "This is a test value.");
+  let str = MyString.new_("This is a test value.");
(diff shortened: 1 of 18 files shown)


feature_tests/js/test/slices.mjs
```diff
@@ -1,18 +1,23 @@
-import test from 'ava';
```
jcrist1, line 1: my editor reformatted some stuff. I hope this is okay.

tool/src/js/converter.rs
```diff
@@ -257,7 +257,7 @@ impl<'jsctx, 'tcx> TyGenContext<'jsctx, 'tcx> {
                         // We basically iterate through and read each string into the array.
                         // TODO: Need a test for this.
                         format!(
-                            r#"new diplomatRuntime.DiplomatSliceStrings(wasm, {variable_name}, "string{}", {edges})"#,
+                            r#"new diplomatRuntime.DiplomatSliceStrings(wasm, {variable_name}, "string{}", {edges}).getValue()"#,
```
jcrist1, line 260: As far as i could see there isn't actually a `DiplomatSliceStrings` in the runtime, but I thought consistency would be a good guide

Manishearth (approved):

PR #738, opened 2024-12-03
Adding support for overriding functions in Kotlin

emarteca:
Adding support for `override` for generated kotlin functions that have the compatible signature of the function they should be overriding.
Right now just adding support for overriding `toString`. 
This means that, if a Kotlin function with the signature `fun toString(): String` is generated, `override` is added to it so that it overrides the default toString. 

This is specifically useful for allowing default printing of objects. If a Kotlin `Result` has an error type that overrides `toString`, then that is called to display the error. With this PR, we can customize what's printed here if the Result has a Rust-backed error type.

### DIFF
--- a/feature_tests/kotlin/somelib/src/main/kotlin/dev/diplomattest/somelib/Float64Vec.kt
+++ b/feature_tests/kotlin/somelib/src/main/kotlin/dev/diplomattest/somelib/Float64Vec.kt
@@ -144,7 +144,7 @@ class Float64Vec internal constructor (
         
     }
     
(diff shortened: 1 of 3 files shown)


Manishearth (commented):
Not opposed to this way of solving this, but we already have the `stringifier` attribute, can we use that and have any method with `stringifier` automatically have its signature set to `override fun toString()`?

emarteca:
Oh, good point! I didn't find the `stringifier` attribute, I'll use that instead.

Manishearth (approved):

PR #758, opened 2025-01-05
Kotlin `Throwable` primitives and non-opaque enums (if marked as `attr(auto, error)`)

emarteca:
In https://github.com/rust-diplomat/diplomat/pull/741, we added support for throwable custom types for opaques and non-opaque structs that are returned as errors.

In this PR, we add support for returning `Throwable` enums and primitives. This requires creating a `Throwable` wrapper class for every primitive type, and generating one for any enum marked as `attr(auto, error)`.  We also now throw an error during code generation if a user attempts to return a Result of an enum not marked as `attr(auto, error)` (same behaviour as for opaques and non-opaque structs).

Last part of fixing https://github.com/rust-diplomat/diplomat/issues/697


Manishearth (approved):
I guess enums can't be throwable?

emarteca:
Yeah, sadly enums aren't throwable in Kotlin

PR #763, opened 2025-01-15
Fix owned opaque in out struct

robertbastian:


robertbastian:
I'm going to merge this because I need it in ICU4X.

PR #795, opened 2025-02-19
ZST methods

robertbastian:


Manishearth:
Oh, sorry, somehow missed this. Happy to review this if you still want this in.

I know @tyler-zeromatter was interested at _some_ point in having more ZST stuff but I don't really recall what the status quo is.

robertbastian:
this was in draft. it's too old now and needs a lot of work for the new backends

Manishearth:
FWIW Diplomat is explicitly designed to allow new features to be only implemented partially across backends. We can even add an attr config for this, though for random combinations of features like this we usually just prefer the backend to return an error.

Manishearth:
But if returning an error would require its own set of nitpicky backend changes then adding a backend supports value (methods_on_zsts) would let you gradually enable things across backends.

robertbastian:
> FWIW Diplomat is explicitly designed to allow new features to be only implemented partially across backends.

I am aware. This code was written before that was the case, so again, updating it would be about the same amount of work as redoing it.

Manishearth:
Fair enough! Let's see if Tyler is interested in this.

Manishearth:
(Otherwise I might knock this out at some point)

tyler-zeromatter:
If I remember rightly, my interest in ZST stuff was mostly due to Nanobind handling ZSTs weirdly. It would be nice to have ZST methods, but it's not a *must* for Zeromatter (opaques work well enough for now)

tyler-zeromatter:
Anyways, this isn't high on my to-do list, so anyone else feel free to tackle this

PR #831, opened 2025-03-29
Add Lifetimes to DiplomatSlice types

ambiguousname:
We return `Some` only for `DiplomatOwnedSlice` lifetimes, when in fact:
- `DiplomatSlice` and `DiplomatSliceMut` should have lifetimes, and
- `DiplomatOwnedSlice` should not have a lifetime, as it is owned.

This fixes that


ambiguousname:
Because a `None` lifetime is valid, JS has also been update to account for this.

Manishearth (approved):

PR #867, opened 2025-05-13
Support returning primitives and enums in `demo_gen`

robertbastian:


book/src/demo_gen/renderer.md
```diff
@@ -86,7 +86,7 @@ Here's the current script that [the quickstart](quickstart.md) has to list all p
 ```js
 import { RenderInfo } from "./demo_gen/index.mjs";
 
-Object.values(RenderInfo.termini).forEach((t) => {
+Object.values(RenderInfo.termini).toSorted((a, b) => a.funcName < b.funcName ? -1 : 1).forEach((t) => {
```
ambiguousname, line 89: This looks good, that'll have to be updated in https://github.com/rust-diplomat/demo-gen-quickstart/blob/main/adder_bindings/demo/index.mjs

ambiguousname (approved):
This looks great to me, everything looks good!

tool/src/demo_gen/terminus.rs
```diff
@@ -173,6 +173,16 @@ impl RenderTerminusContext<'_, '_> {
...
+            || method
+                .output
+                .success_type()
+                .as_type()
+                .is_some_and(|t| matches!(t, hir::OutType::Enum(_) | hir::OutType::Primitive(_)))
+                && method
+                    .param_self
+                    .iter()
+                    .all(|s| !s.ty.is_mutably_borrowed())
+                && method.params.iter().all(|p| !p.ty.is_mutably_borrowed())
```
sffc, line 185: Issue: It looks to me like these conditions should apply to methods that return strings as well as methods that return primitives, but currently they are only applying to methods that return primitives due to operator precedence

https://doc.rust-lang.org/reference/expressions.html#r-expr.precedence

Even if this is intentional, please use parentheses so I don't need to check the precedence table

sffc (commented):
Looks quite good overall; thanks for adding this. One comment/question

tool/src/demo_gen/terminus.rs
robertbastian: it's probably not going to make a difference (doesn't for ICU4X right now), as mutators don't usually return strings (but they might return status codes). will change this though

PR #908, opened 2025-06-30
&[Struct] support

tyler-zeromatter:
Fix for https://github.com/rust-diplomat/diplomat/issues/769, adds support for `&[Struct]` types. With the limitation that the Structs must only contain non-slice primitives (currently need to actually put checks for this limitation in HIR code).

Currently this is just the rough outline of the setup inside the AST/HIR. A lot of how this gets implemented will be down to the backends. Let me know if this matches roughly the skeleton proposed with the original issue.

I plan on also adding a sample implementation for C++/Nanobind.

### DIFF
--- a/core/src/ast/types.rs
+++ b/core/src/ast/types.rs
@@ -456,8 +456,9 @@ pub enum TypeName {
     /// If StdlibOrDiplomat::Stdlib, it's specified as `&[&DiplomatFoo]`, if StdlibOrDiplomat::Diplomat it's specified
     /// as `DiplomatSlice<&DiplomatFoo>`
     StrSlice(StringEncoding, StdlibOrDiplomat),
-    /// `&[Struct]`. Meant for structs that are only composed of primitives (or structs of primitives) that are *not* slices.
-    /// This makes custom math types like `Vec<Vector3>` easier to pass directly, and this is implemented on a per-backend basis.
+    /// `&[Struct]`. Meant for passing slices of structs where the struct's layout is known to be shared between Rust
+    /// and the backend language. Primarily meant for large lists of compound types like `Vector3f64` or `Color4i16`. 
+    /// This is implemented on a per-backend basis.
     PrimitiveStructSlice(Option<(Lifetime, Mutability)>, Box<TypeName>),
     /// The `()` type.
     Unit,
--- a/core/src/ast/types.rs
+++ b/core/src/ast/types.rs
@@ -457,7 +457,7 @@ pub enum TypeName {
     /// as `DiplomatSlice<&DiplomatFoo>`
     StrSlice(StringEncoding, StdlibOrDiplomat),
     /// `&[Struct]`. Meant for passing slices of structs where the struct's layout is known to be shared between Rust
-    /// and the backend language. Primarily meant for large lists of compound types like `Vector3f64` or `Color4i16`. 
+    /// and the backend language. Primarily meant for large lists of compound types like `Vector3f64` or `Color4i16`.
     /// This is implemented on a per-backend basis.
     PrimitiveStructSlice(Option<(Lifetime, Mutability)>, Box<TypeName>),
     /// The `()` type.
--- a/core/src/ast/types.rs
+++ b/core/src/ast/types.rs
@@ -459,7 +459,7 @@ pub enum TypeName {
     /// `&[Struct]`. Meant for passing slices of structs where the struct's layout is known to be shared between Rust
     /// and the backend language. Primarily meant for large lists of compound types like `Vector3f64` or `Color4i16`.
     /// This is implemented on a per-backend basis.
-    PrimitiveStructSlice(Option<(Lifetime, Mutability)>, Box<TypeName>),
+    CustomTypeSlice(Option<(Lifetime, Mutability)>, Box<TypeName>),
     /// The `()` type.
     Unit,
     /// The `Self` type.
@@ -580,7 +580,7 @@ impl TypeName {
             // can only be passed across the FFI boundary; callbacks and traits are input-only
             TypeName::Function(..) | TypeName::ImplTrait(..) |
             // These are specified using FFI-safe diplomat_runtime types
-            TypeName::StrReference(.., StdlibOrDiplomat::Diplomat) | TypeName::StrSlice(.., StdlibOrDiplomat::Diplomat) |TypeName::PrimitiveSlice(.., StdlibOrDiplomat::Diplomat) | TypeName::PrimitiveStructSlice(..) => true,
+            TypeName::StrReference(.., StdlibOrDiplomat::Diplomat) | TypeName::StrSlice(.., StdlibOrDiplomat::Diplomat) |TypeName::PrimitiveSlice(.., StdlibOrDiplomat::Diplomat) | TypeName::CustomTypeSlice(..) => true,
             // These are special anyway and shouldn't show up in structs
             TypeName::Unit | TypeName::Write | TypeName::Result(..) |
             // This is basically only useful in return types
@@ -695,7 +695,7 @@ impl TypeName {
                     primitive.get_diplomat_slice_type(ltmt)
                 }
             }
-            TypeName::PrimitiveStructSlice(ltmt, type_name) => {
+            TypeName::CustomTypeSlice(ltmt, type_name) => {
                 let inner = type_name.to_syn();
                 if let Some((ref lt, ref mtbl)) = ltmt {
                     let reference = ReferenceDisplay(lt, mtbl);
@@ -789,7 +789,7 @@ impl TypeName {
                         }
                         return TypeName::StrSlice(encoding, StdlibOrDiplomat::Stdlib);
                     }
-                    return TypeName::PrimitiveStructSlice(
+                    return TypeName::CustomTypeSlice(
                         Some((lifetime, mutability)),
                         Box::new(TypeName::from_syn(slice.elem.as_ref(), self_path_type)),
                     );
@@ -1311,10 +1311,10 @@ impl fmt::Display for TypeName {
                 write!(f, "DiplomatSlice{maybemut}<{lt}{typ}>")
             }
             TypeName::PrimitiveSlice(None, typ, _) => write!(f, "Box<[{typ}]>"),
-            TypeName::PrimitiveStructSlice(Some((lifetime, mutability)), type_name) => {
+            TypeName::CustomTypeSlice(Some((lifetime, mutability)), type_name) => {
                 write!(f, "{}[{type_name}]", ReferenceDisplay(lifetime, mutability))
             }
-            TypeName::PrimitiveStructSlice(None, type_name) => write!(f, "Box<[{type_name}]>"),
+            TypeName::CustomTypeSlice(None, type_name) => write!(f, "Box<[{type_name}]>"),
             TypeName::Unit => "()".fmt(f),
             TypeName::Function(input_types, out_type, _mutability) => {
                 write!(f, "fn (")?;
--- a/core/src/hir/attrs.rs
+++ b/core/src/hir/attrs.rs
@@ -850,9 +850,8 @@ impl Attrs {
         }
 
         if *allowed_in_slices && !matches!(context, AttributeContext::Type(TypeDef::Struct(..))) {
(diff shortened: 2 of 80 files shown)


core/src/hir/types.rs
```diff
@@ -22,7 +22,7 @@ pub enum Type<P: TyPosition = Everywhere> {
     Struct(P::StructPath),
     ImplTrait(P::TraitPath),
     Enum(EnumPath),
-    Slice(Slice),
+    Slice(Slice<P::StructPath>),
```
Manishearth: thought: should we only support slices that contain input structs? I can't see a usecase for `&[OutStruct]`

There may be a use case for `-> Box<[OutStruct]>` but I don't think we support returning owned slices anyway

core/src/hir/type_context.rs
```diff
@@ -459,6 +464,49 @@ impl TypeContext {
             }
         }
     }
+
+    fn validate_primitive_slice_struct<P: TyPosition>(
+        &self,
+        errors: &mut ErrorStore,
+        st: &P::StructPath,
+    ) {
+        let ty = self.resolve_type(st.id());
```
Manishearth: issue: we should also validate that it has no lifetimes

That's a removable restriction but it basically doesn't make sense for now.

core/src/hir/type_context.rs
```diff
@@ -459,6 +464,49 @@ impl TypeContext {
...
+        match ty {
+            TypeDef::Struct(st) => {
+                for f in &st.fields {
+                    match &f.ty {
+                        hir::Type::Primitive(..) => {}
+                        hir::Type::Struct(st) => {
+                            self.validate_primitive_slice_struct::<hir::Everywhere>(errors, st);
```
Manishearth: thought: I don't like having to do this type of recursive validation (that jumps to other types): it has a potential to get quadratic really quickly.


I think we should have `#[diplomat::attr(auto, allowed_in_slices)]`? Then validation checks:

 - This attribute is applied on structs that only have allowed-in-slices fields
 - This attribute exists on all structs found in slices

Not crucial for a first pass, but I'd want _something_ here for a followup. Another design would be to cache the validation results. I generally prefer explicitness but I can understand how explicitness would be annoying to users.

Another benefit of the explicitness is that `is_sliceable`-type code in the backend just needs to check an attribute. Easy.

Manishearth (commented):
Overall looks good. I think I'd like sliceability to be an explicit opt in.

core/src/hir/types.rs
tyler-zeromatter: I think that makes sense to me, @Walter-Reactor can you think of any use cases for slices of out structs?

core/src/hir/type_context.rs
tyler-zeromatter: I agree, that sounds good!

core/src/hir/types.rs
Manishearth: One potential reason I could see is if having code that is "input structs _only_" might actually complicate things since all the rest of the code is generic around struct kind. I doublt it, but that's a valid reason to be a bit more permissive here

core/src/hir/type_context.rs
tyler-zeromatter: I'm curious, since I've never worked with lifetimes in the HIR before. How would I verify that the slice (with lifetime information stored in `Option<Borrow>`) has no lifetimes? Would it just be to check `Option<Borrow>.is_none()`?

Manishearth: No you want to check that the _struct_ doesn't have any lifetime parameters. So check on `ty`.

I think you can even do `ty.lifetimes().is_empty()` or something

core/src/hir/types.rs
tyler-zeromatter: Yeah, I think we'll avoid output structs for now

core/src/hir/attrs.rs
```diff
@@ -833,6 +848,13 @@ impl Attrs {
                 "`generate_mocking_interface` can only be used on opaque types".to_string(),
             ));
         }
+
+        if *allowed_in_slices && !matches!(context, AttributeContext::Type(TypeDef::Struct(..))) {
+            // TODO: Allow OutStructs?
+            errors.push(LoweringError::Other(
+                "`allowed_in_slices` can only be used on input-only struct types.".into(),
```
Manishearth: nit: non-output-only structs.

regular structs can be used anywhere.

tyler-zeromatter:
Barring any major issues (and the documentation which I will now be adding), I believe this is pretty much ready to merge.

core/src/ast/types.rs
```diff
@@ -456,6 +456,9 @@ pub enum TypeName {
     /// If StdlibOrDiplomat::Stdlib, it's specified as `&[&DiplomatFoo]`, if StdlibOrDiplomat::Diplomat it's specified
     /// as `DiplomatSlice<&DiplomatFoo>`
     StrSlice(StringEncoding, StdlibOrDiplomat),
+    /// `&[Struct]`. Meant for structs that are only composed of primitives (or structs of primitives) that are *not* slices.
+    /// This makes custom math types like `Vec<Vector3>` easier to pass directly, and this is implemented on a per-backend basis.
```
walter-zeromatter:
```suggestion
    /// `&[Struct]`. Meant for passing slices of structs where the struct's layout is known to be shared between Rust
    /// and the backend language. Primarily meant for large lists of compound types like `Vector3f64` or `Color4i16`. 
    /// This is implemented on a per-backend basis.
```

feature_tests/c/include/BorrowedFieldsWithBounds.d.h
```diff
@@ -21,4 +21,6 @@ typedef struct BorrowedFieldsWithBounds_option {union { BorrowedFieldsWithBounds
 
 
 
+
```
walter-zeromatter: Probably best if you can eliminate the change adding all this whitespace

tool/src/dart/mod.rs
```diff
@@ -275,7 +275,7 @@ impl<'cx> TyGenContext<'_, 'cx> {
 
                 /// Get the name/initializer of the allocator needed for a particular type
                 fn alloc_name<P: TyPosition>(ty: &hir::StructDef<P>, field_ty: &Type<P>) -> Option<String> {
-                    if let &hir::Type::Slice(slice) = field_ty {
+                    if let hir::Type::Slice(ref slice) = field_ty {
```
walter-zeromatter: I thought the syntax for this was `if let hir::Type::Slice(&slice) = field_ty {` and the use of `ref` in this context was archaic?
https://doc.rust-lang.org/1.33.0/book/ch18-03-pattern-syntax.html#legacy-patterns-ref-and-ref-mut

tyler-zeromatter:
Just did a bunch of work to install gradle and test in WSL, but tests are all passing locally. So I assume there's something wonky going on in the CI? Not sure what it is yet though.

Walter-Reactor (commented):
Looks good overall. I've got a few near-nits, and some suggested rewording on one of the comments but that's about it.
Nice work

tool/src/dart/mod.rs
tyler-zeromatter: I think some of this stuff is archaic, I didn't realize that `ref` was supposed to be avoided. I'll remove it here.

tyler-zeromatter:
Anyways, note the latest version of the Kotlin CI is passing, so it seems to be random chance whether or not it passes or fails right now??

tyler-zeromatter:
Turns out, we do want to add support for `-> &[Struct]`. nanobind has some funny limitations for passing around slices, and there we want to ensure that users create a type that will return C++ ABI friendly `diplomat::span`s. I'll work on adding this functionality in a separate PR, once this gets merged.

Manishearth:
@tyler-zeromatter returning `&[Struct]` is one thing, but returning slices that contain _output structs_ is another. Remember, the main thing output structs can contain is `Box<Opaque>`, and if that's in a `&[Struct]` , you won't be able to access it anyway (unless https://github.com/rust-diplomat/diplomat/issues/792 happens). That's a situation when you should just return an iterator; though presumably we could eventually support returning `Box<[StructWithOwnedStuff]>`

We aboslutely should support returning `&[Struct]`.

core/src/ast/types.rs
```diff
@@ -456,6 +456,10 @@ pub enum TypeName {
     /// If StdlibOrDiplomat::Stdlib, it's specified as `&[&DiplomatFoo]`, if StdlibOrDiplomat::Diplomat it's specified
     /// as `DiplomatSlice<&DiplomatFoo>`
     StrSlice(StringEncoding, StdlibOrDiplomat),
+    /// `&[Struct]`. Meant for passing slices of structs where the struct's layout is known to be shared between Rust
+    /// and the backend language. Primarily meant for large lists of compound types like `Vector3f64` or `Color4i16`.
+    /// This is implemented on a per-backend basis.
+    PrimitiveStructSlice(Option<(Lifetime, Mutability)>, Box<TypeName>),
```
Manishearth: nit: Probably should be called CustomTypeSlice

core/src/hir/type_context.rs
```diff
@@ -459,6 +463,54 @@ impl TypeContext {
...
+                }
+
+                if st.lifetimes.num_lifetimes() > 0 {
+                    errors.push(LoweringError::Other(format!(
+                        "Struct {:?} cannot have any lifetimes if it is within a slice.",
+                        st.name
+                    )));
+                }
+
+                for f in &st.fields {
```
Manishearth: issue: we should not be doing this validation here; this validation will be repeated for every `&[Struct]` in code.

The purpose of the flag is in part so we can do validation once.

Furthermore, this branch only validates types in methods.
We should instead:

 - Add a `validate_type_def` that calls a `validate_struct` 
 - `validate_struct` calls `validate_ty` on all of its fields. We should also probably have a `validate_trait`.
 - `validate_ty` ensures that slice types contain only `allowed_in_slices` structs
 - `validate_ty_in_method` calls `validate_ty`
 - `validate_struct` ensures it has valid fields for `allowed_in_slices` if that flag is set.

We want to decouple use-time validation from def-time validation.

Manishearth (commented):
Reviewed hir code

Manishearth (commented):
Rest looks good

Philosophically I would like each backend to make the choice of what types to support, but we should have per-backend validation hooks for that.

core/src/hir/type_context.rs
tyler-zeromatter: Let me know if these newer changes look good

core/src/hir/type_context.rs
```diff
@@ -464,51 +465,69 @@ impl TypeContext {
...
-                if !st.attrs.allowed_in_slices {
-                    errors.push(LoweringError::Other(format!(
-                        "Cannot construct a slice of {:?}. Try marking with `#[diplomat::attr(auto, allowed_in_slices)]`",
-                        st.name
-                    )));
-                    return;
-                }
+    fn validate_type_def(&self, errors: &mut ErrorStore, def: TypeDef) {
+        if let TypeDef::Struct(st) = def {
+            self.validate_struct(errors, st);
```
Manishearth, line 497: nit: probably should also have a `validate_trait` so that we can `validate_ty` everything in traits as well

core/src/hir/type_context.rs
```diff
@@ -464,51 +465,69 @@ impl TypeContext {
...
+                            "Cannot construct a slice of {:?} with non-primitive, non-struct field {:?}",
+                            st.name, f.name
+                        )));
+                    }
+                }
+            }
+        }
+    }
+
+    fn validate_ty<P: super::TyPosition>(&self, errors: &mut ErrorStore, ty: &hir::Type<P>) {
```
Manishearth: nit: comment about validate_ty vs validate_ty_in_method, on both functions, and they should live next to each other

core/src/hir/type_context.rs
tyler-zeromatter: Added in a quick version of this to validate all the parameters in the methods of traits, let me know if you'd like something more robust

Manishearth (approved):

Manishearth:
Thanks for all the work so far! I'm excited for this!

:tada:

PR #910, opened 2025-07-10
Fix `RecordUse` in Dart

mosuem:
The first commit has the changes, the second the regeneration of generated files.


robertbastian (approved):
Huh why am I having a deja vu? I swear we added the annotation to finalizers before 🤔

PR #912, opened 2025-07-10
Nanobind reference leak fix

tyler-zeromatter:
Not sure how `nb::keep_alive` works, but the current way it's being used in the nanobind bindings is leading to reference leaks for returning references.

This fixes that (and adds a failing test to check nanobind for memory leaks for some functions), but @Walter-Reactor your thoughts here are appreciated.

### DIFF
--- a/tool/src/nanobind/ty.rs
+++ b/tool/src/nanobind/ty.rs
@@ -341,12 +341,14 @@ impl<'ccx, 'tcx: 'ccx> TyGenContext<'ccx, 'tcx> {
 
         let mut visitor = method.borrowing_param_visitor(self.c2.tcx, false);
 
+        let borrowed_self = if let Some(s) = &method.param_self {
+            Some(visitor.visit_param(&s.ty.clone().into(), "self"))
+        } else {
+            None
+        };
+
         // Collect all the relevant borrowed params, with self in position 1 if present
-        let mut param_borrows = method
-            .param_self
-            .iter()
-            .map(|s| visitor.visit_param(&s.ty.clone().into(), "self"))
-            .collect::<Vec<_>>();
+        let mut param_borrows = Vec::new();
         // Must be a separate call *after* collect to avoid double-borrowing visitor
         param_borrows.extend(
             method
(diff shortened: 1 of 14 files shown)


tyler-zeromatter:
Wait, did the reference checking thing wrong, fixing now

tyler-zeromatter:
After chatting with @Walter-Reactor, seems like nanobind is smart enough to figure out that we're returning a reference and handles things accordingly. In fact, `nb::rv_policy::reference` is ignored unless we're returning a previously unknown C++ instance, which tends only to be the case for returning iterators. So we only need `keep_alive` and `rv_policy::reference` in those cases. Will spend some time today updating nanobind gen accordingly.

tyler-zeromatter:
There are also some cases involving owned Opaque or Struct types that are returned that I didn't catch, but the tests thankfully did.

tool/src/nanobind/ty.rs
```diff
@@ -366,11 +366,19 @@ impl<'ccx, 'tcx: 'ccx> TyGenContext<'ccx, 'tcx> {
 
         let mut lifetime_args = vec![];
 
+        // Any returned values that are created on return (i.e., iterators, owned structs, owned opaques)
+        // that reference lifetimes are the only returned values that need to be kept alive, per: https://nanobind.readthedocs.io/en/latest/api_core.html#_CPPv4N8nanobind9rv_policyE
+        // They are an instantiated C++ type that is unknown to nanobind.
+        // For non-iterator references, Nanobind handles reference conversion automatically.
```
walter-zeromatter:
```suggestion
        // Only returned values that are either created on return or return addresses previously unknown
        // to nanobind require additional annotation with keep_alive per: https://nanobind.readthedocs.io/en/latest/api_core.html#_CPPv4N8nanobind9rv_policyE
        // For any return of a reference to an existing object, nanobind is smart enough to locate it's python wrapper object & correctly increment it's refcount
```

Walter-Reactor (approved):
Overall this looks good. Glad you found this! figuring out the advanced parts of how nanobind handles refcounts is definitely a journey.

PR #913, opened 2025-07-16
Diplomat Macros

tyler-zeromatter:
Implements #752.

This is currently very barebones, but there is a working example in feature_tests! So the TODOs:

- [ ] Fix macro matcher reading (and add errors for when inputs don't match). So things like `macro!($var:tt asdkfja $other_var:expr)` aren't supported yet.
- [ ] Add validation for location of macros
- [ ] Add support for macro usage in Item impl
- [ ] Write a working test for macros

### DIFF
--- a/core/src/ast/macros.rs
+++ b/core/src/ast/macros.rs
@@ -1,11 +1,11 @@
 use std::collections::BTreeMap;
 
-use proc_macro2::TokenStream;
-use quote::ToTokens;
-use syn::{braced, bracketed, buffer::{Cursor, TokenBuffer}, parenthesized, parse::{self, Parse}, parse_macro_input, spanned::Spanned, token::{self, Token}, Error, Expr, ExprParen, Ident, Item, ItemMacro, PatParen, Path, Token};
+use proc_macro2::{Delimiter, TokenStream, TokenTree};
+use quote::{ToTokens, TokenStreamExt};
+use syn::{braced, bracketed, buffer::{Cursor, TokenBuffer}, parenthesized, parse::{self, Parse, ParseBuffer}, parse_macro_input, spanned::Spanned, token::{self, Token}, Error, Expr, ExprParen, Ident, Item, ItemMacro, PatParen, Path, Token};
 
 pub struct Macros {
-    defs : BTreeMap<String, MacroRules>,
+    defs : BTreeMap<Ident, MacroRules>,
 }
 
 impl Macros {
@@ -17,20 +17,18 @@ impl Macros {
 
     pub fn read_item_macro(&mut self, input : &ItemMacro) -> Option<Vec<Item>> {
         let mac = Macro::from_syn(&input);
-        // FIXME: Extremely hacky.
-        let path_str = input.mac.path.to_token_stream().to_string();
-        if let Ok(mac) = mac {
+        if let Ok((ident, mac)) = mac {
             match mac {
                 Macro::MacroRules(rules) => {
-                    self.defs.insert(path_str, rules);
+                    self.defs.insert(ident, rules);
                     // Macro rules add no new items: 
                     None
                 },
                 Macro::MacroMatch(matched) => {
-                    if let Some(def) = self.defs.get(&path_str) {
+                    if let Some(def) = self.defs.get(&ident) {
                         Some(def.evaluate(matched))
                     } else {
-                        panic!("Could not find definition for {:?}", path_str);
+                        panic!("Could not find definition for {:?}", ident);
                     } 
                 }
             }
@@ -48,14 +46,16 @@ pub enum Macro {
 }
 
 impl Macro {
-    pub fn from_syn(input : &ItemMacro) -> Result<Macro, syn::Error> {
+    pub fn from_syn(input : &ItemMacro) -> Result<(Ident, Macro), syn::Error> {
         // Are we macro_rules!
-        if input.ident.is_some() {
+        if let Some(ident) = &input.ident {
             let r = input.mac.parse_body()?;
-            Ok(Macro::MacroRules(r))
+            Ok((ident.clone(), Macro::MacroRules(r)))
         } else {
             let m = input.mac.parse_body()?;
-            Ok(Macro::MacroMatch(m))
+            // FIXME: Extremely hacky.
+            let path_ident = input.mac.path.segments.last().unwrap().ident.clone();
+            Ok((path_ident, Macro::MacroMatch(m)))
         }
     }
 
@@ -140,6 +140,7 @@ impl Parse for MacroRules {
             return Err(Error::new(input.span(), "Expected {}, (), or []"))
         }
 
+        // FIXME: This is not a comma separated list in actuality.
         let punc = arm.parse_terminated(MacroIdent::parse, Token![,])?;
 
         let match_tokens = punc.iter().map(|i| { i.clone() }).collect();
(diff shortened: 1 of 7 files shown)


tyler-zeromatter:
@Walter-Reactor

tyler-zeromatter:
Some additional issues to test: blocks in blocks?

core/src/ast/macros.rs
```diff
@@ -0,0 +1,302 @@
...
+pub struct MacroMatch {
+    pub args: Vec<Expr>,
+}
+
+impl Parse for MacroMatch {
+    fn parse(input: parse::ParseStream) -> syn::Result<Self> {
+        let mut args = Vec::new();
+
+        // TODO: Need some custom logic based on the parent macro definition.
+        let args_p = input.parse_terminated(Expr::parse, Token![,])?;
```
tyler-zeromatter: This currently works fine with most cases I've tested, but I'm curious if we want to use `Expr` to parse matchers from a macro call?

This will hopefully be replaced in future updates to macro parsing regardless, but I'm wondering if we want something more robust now.

core/src/ast/macros.rs
```diff
@@ -0,0 +1,302 @@
...
+}
+
+impl Macros {
+    pub fn new() -> Macros {
+        Macros {
+            defs: BTreeMap::new(),
+        }
+    }
+
+    pub fn read_item_macro(&mut self, input: &ItemMacro) -> Option<Vec<Item>> {
```
Manishearth: nit: `add_item_macro`?

(same below)

core/src/ast/macros.rs
```diff
@@ -0,0 +1,302 @@
...
+        let mac = Macro::from_syn(input);
+        if let Ok((ident, mac)) = mac {
+            match mac {
+                Macro::MacroRules(rules) => {
+                    self.defs.insert(ident, rules);
+                    // Macro rules add no new items:
+                    None
+                }
+                Macro::MacroMatch(matched) => {
```
Manishearth: issue: I'd prefer a cleaner `add_macro` / `evaluate_macro` separation, with asserts about the right type of `Macro` coming in

core/src/ast/macros.rs
```diff
@@ -0,0 +1,302 @@
...
+        } else {
+            // We handle errors automatically in `diplomat/macro`
+            Vec::new()
+        }
+    }
+}
+
+#[derive(Debug)]
+#[non_exhaustive]
+pub enum Macro {
```
Manishearth: issue: what's the benefit of treating these as an enum? they should be constructed and handled in different places. I think we should have `MacroDef` / `MacroUse` or something similar to everything else.

core/src/ast/modules.rs
```diff
@@ -154,138 +165,223 @@ impl Module {
...
-                .attrs
-                .iter()
-                .any(|a| a.path().to_token_stream().to_string() == "diplomat :: bridge");
-
-        let mod_attrs: Attrs = (&*input.attrs).into();
-
-        let impl_parent_attrs: Attrs =
-            mod_attrs.attrs_for_inheritance(AttrInheritContext::MethodOrImplFromModule);
-        let type_parent_attrs: Attrs = mod_attrs.attrs_for_inheritance(AttrInheritContext::Type);
+    fn evaluate_item(a: &Item, mst: &mut ModuleSynTraverse) {
```
Manishearth: nit: add_item, we're not evaluating

core/src/ast/modules.rs
```diff
@@ -154,138 +165,223 @@ impl Module {
...
                 }
-                Item::Trait(trt) => {
-                    if analyze_types {
-                        let ident = (&trt.ident).into();
-                        let trt = Trait::new(trt, &type_parent_attrs);
-                        custom_traits_by_name
-                            .insert(ident, trt);
+            }
+            Item::Macro(mac) => {
+                let should_eval = if let Some(i) = &mac.ident {
```
Manishearth: issue: this logic  seems overnested; the `if let Some(i) = mac.ident` check should be hoisted up so that the code handling defs and invocations is separate. Then the invocations code doesn't need a separate should_eval variable, it can just return early.

core/src/ast/modules.rs
```diff
@@ -154,138 +165,223 @@ impl Module {
...
-                            }
-                            Err(errors) => {
-                                panic!("Multiple conflicting Diplomat struct attributes, there can be at most one: {errors:?}");
-                            }
-                        };
+            }
 
-                        custom_types_by_name.insert(Ident::from(&strct.ident), custom_type);
-                    }
+            Item::Enum(enm) => {
```
Manishearth, line 163: note: I'm assuming these other branches are largely unchanged.

core/src/ast/macros.rs
```diff
@@ -0,0 +1,279 @@
...
+
+        let buf = TokenBuffer::new2(self.body.clone());
+        let mut c = buf.begin();
+        // Search until we find a token to replace:
+        while let Some((tt, next)) = c.token_tree() {
+            match &tt {
+                TokenTree::Punct(punct) if punct.as_char() == '$' => {
+                    if let Some((tt, next)) = next.token_tree() {
+                        if let TokenTree::Ident(i) = tt {
+                            let arg = self.match_tokens.iter().position(|mi| mi.ident == i);
```
Manishearth: issue: We should be storing these match tokens as a map from ident to index, i nthe parsed macro.

core/src/ast/modules.rs
```diff
@@ -106,6 +106,17 @@ pub struct Module {
     pub attrs: Attrs,
 }
 
+struct ModuleSynTraverse {
```
Manishearth: nit: docs please.

Probably call this "ModuleBuilder"?

And `add_item` can be a `.add(&mut self)` method on this.

Manishearth (commented):
Okay, I think I finally got to do a full review. This looks good, some minor changes.

core/src/ast/modules.rs
tyler-zeromatter: Yes, I haven't touched anything but adding `Item::Macro`

core/src/ast/modules.rs
tyler-zeromatter: Let me know if the new version looks cleaner

core/src/ast/macros.rs
```diff
@@ -102,14 +106,14 @@ impl Parse for MacroIdent {
...
-        Ok(Self { ident, ty })
+        Ok(Self { ident, _ty: ty })
     }
 }
 
 #[derive(Debug)]
 #[non_exhaustive]
 pub struct MacroDef {
-    pub match_tokens: Vec<MacroIdent>,
+    pub match_tokens: HashMap<Ident, usize>,
```
Manishearth, line 125: nit: docs

core/src/ast/macros.rs
```diff
@@ -93,7 +96,8 @@ impl Parse for MacroUse {
 #[derive(Debug, Clone)]
 pub struct MacroIdent {
     pub ident: Ident,
-    pub ty: Ident,
+    // To be used in future arg parsing:
+    pub _ty: Ident,
```
Manishearth, line 108: question: say more?

core/src/ast/modules.rs
```diff
@@ -310,15 +310,19 @@ impl Module {
...
-                    let macro_rules_attr = mac.attrs.iter().find_map(move |a| {
-                        if a.path()
-                            == &syn::parse_str::<syn::Path>("diplomat::macro_rules").unwrap()
-                        {
-                            Some(true)
-                        } else {
-                            None
-                        }
-                    }).unwrap_or(false);
+                    let macro_rules_attr = mac
```
Manishearth: nit: I htink this can be `mac.attrs...find(|a| a.path == ... )` without any of the option or unwrap_or stuff

Manishearth (approved):

core/src/ast/macros.rs
Manishearth: Expr seems fine for now. Most of the time I expect people to be passing in idents or literals.

PR #915, opened 2025-07-21
Nanobind Support for Primitive Struct Slices

tyler-zeromatter:


Manishearth (approved):
Performing surface level review of nanobind backend, feel free to merge or wait for @Walter-Reactor

tool/src/nanobind/ty.rs
```diff
@@ -165,6 +165,7 @@ impl<'ccx, 'tcx: 'ccx> TyGenContext<'ccx, 'tcx> {
         def: &'tcx hir::StructDef<P>,
         id: TypeId,
         out: &mut W,
+        header: &mut W,
```
walter-zeromatter: nit: Header is used throughout the codebase to refer to code that goes in a header file - Could we name this something else like 'binding_prefix`

tool/templates/nanobind/struct_impl.cpp.jinja
```diff
@@ -1,5 +1,9 @@
 {%- let is_self_opaque = false -%}
 
+{%- if is_sliceable %}
+nb::bind_vector<std::vector<{{type_name}}>>(mod, "{{type_name}}Slice");
```
walter-zeromatter, line 7: I think it's worth noting, maybe here, maybe somewhere else, that the reason we have to create these vector bindings is that the native python list type doesn't actually store elements contiguously. We will probably want to add support for numpy arrays eventually, since they serve the same purpose and are commonly used in our domain of work

Walter-Reactor (approved):
Overall looks good, though I have 2 minor nits about naming & adding some comments for future maintainers about why we're binding vector instead of std::span

PR #918, opened 2025-07-23
Add support for &mut Struct for Structs of primitive-only types.

tyler-zeromatter:
Renamed `supports=struct_primitive_slices` to `abi_compatibles` and `#[diplomat::attr(auto, allowed_in_slices)]` to `(auto, abi_compatible)`. Also started on support for passing in (and returning) `&mut Struct` types.

If the rough concept gets approval, then there will be some TODOs:

- [x] ~~Nanobind support (should be in the same PR since Nanobind currently has `abi_compatibles=true`.~~ Nevermind, Nanobind handles this automatically.
- [x] More robust tests (DiplomatOption, passing in, returning) for Mutable and Immutable refs of structs (this is now handled in `type_context.rs`)
- [x] Attribute authentication (if you're trying to use a struct reference, ensure it has the correct type). (Remove corresponding FIXME)
  - Then run `insta` to get the failing tests passing
- [x] Need to do a more thorough survey of the HIR where Opaque mutability is checked (to add structs).
- [x] Docs for this expanded definition


tyler-zeromatter:
I'll CC @Walter-Reactor for this

Walter-Reactor:
The most immediate use case here is being able to implement in-place math operations on Vector3f types or related - You can't do += without &mut access, for example, and there are other operations for which we'd like to be able to do in-place modification

core/src/hir/lowering.rs
```diff
@@ -1377,41 +1394,48 @@ impl<'ast> LoweringContext<'ast> {
         match self_param.path_type.resolve(in_path, self.env) {
             ast::CustomType::Struct(strct) => {
                 if let Some(tcx_id) = self.lookup_id.resolve_struct(strct) {
-                    if self_param.reference.is_some() {
-                        self.errors.push(LoweringError::Other(format!("Method `{method_full_path}` takes a reference to a struct as a self parameter, which isn't allowed")));
-                        Err(())
+                    let (borrow, mut param_ltl) = if let Some((lt, mt)) = &self_param.reference {
+                        if self.attr_validator.attrs_supported().abi_compatibles {
+                            let (borrow_lt, param_ltl) = self_param_ltl.lower_self_ref(lt);
```
Manishearth, line 1399: We should fully disallow lifetimes in these types I think

Manishearth: oh, wait, that's the ref lifetime itself, we need that.


Rephrasing: we should disallow these lifetimes from being borrowed-from, probably. I'm not sure where that would happen.

Manishearth: Then again, all current ABI-compatible backends do not care about lifetime safety and consider it a matter of documentation. Which we currently don't even generate (we should).

tyler-zeromatter:
I think a discussion is warranted. I've opened a new issue for that: https://github.com/rust-diplomat/diplomat/issues/919

core/src/hir/lowering.rs
tyler-zeromatter: I'm not sure what you mean when you say lifetimes being borrowed from. Are you talking about disabling further borrows of `&Struct`? Something like:

```rs
fn takes_mut<'a>(a : &'a mut Struct) -> &'a mut Struct {
  a
}
```

Where the above should be disallowed to restrict further access to that mutable reference?

Manishearth: Yes.

tyler-zeromatter: I agree that should be disallowed. I think good place for checking could be the TypeContext validation functions.

tyler-zeromatter: So based on some quick testing, I think we should disallow two things in the AST:

1. Borrowing structs inside of other structs (or opaques)
```rs
struct StructA {
 x : f32,
}

struct StructB<'a> {
  a : &'a StructA,
}
```
2. Returning references to structs
```rs
fn takes_mut(a : &Struct) -> &Struct
```

Which should both hopefully prevent these references from existing past any one function call.

Manishearth: That's necessary but not sufficient, you also want to disallow `fn takes_mut(a: &'a mut Struct) -> Box<Opaque<'a>>`

tyler-zeromatter: Sounds good, I'll get started on implementing

tyler-zeromatter: Just added checks for this in `type_context.rs`. My changes for checking struct borrows inside of methods feels messy, so feedback there is certainly welcome.

tyler-zeromatter: I also didn't include any checks for Opaques borrowing Structs, since I'm not really sure where I could implement that? `TypeDef` doesn't seem to store that information

core/src/hir/type_context.rs
```diff
@@ -539,6 +552,15 @@ impl TypeContext {
 
         for f in &st.fields {
             self.validate_ty(errors, &f.ty);
+
+            if matches!(f.ty, hir::Type::Struct(..))
+                && (f.ty.is_immutably_borrowed() || f.ty.is_mutably_borrowed())
+            {
+                errors.push(LoweringError::Other(format!(
+                    "Struct {:?} field {:?} cannot be borrowed. Structs cannot be borrowed inside of other structs.",
```
Manishearth: nit: suggest storing it directly instead

core/src/hir/lowering.rs
Manishearth: the only way an opaque would borrow a struct is if a method let it, and you're already checking for that

tyler-zeromatter: Scratch the second half of that question, the AST validator already handles that for us

Manishearth (commented):
was hoping the check could be done with preexisting lifetime infra but it's a small amount of code.

the HIR lifetime infra probably _should_ be written to precalculate these things so you can quickly get these answers

Walter-Reactor (approved):
Looks excellent! This should unblock some of the math ops we want to do on our geometry primitives

core/src/hir/attrs.rs
```diff
@@ -55,8 +55,8 @@ pub struct Attrs {
     pub demo_attrs: DemoInfo,
     /// From #[diplomat::attr()]. If true, generates a mocking interface for this type.
     pub generate_mocking_interface: bool,
-    /// From #[diplomat::attr()]. If true, will be used for generation of [`super::Slice::Struct`] types.
-    pub allowed_in_slices: bool,
+    /// From #[diplomat::attr()]. If true, Diplomat will check that this struct has the same memory layout in backends which support it. Allows this struct to be used in slices ([`super::Slice::Struct`]) and to be returned mutably.
```
Manishearth:
```suggestion
    /// From #[diplomat::attr()]. If true, Diplomat will check that this struct has the same memory layout in backends which support it. Allows this struct to be used in slices ([`super::Slice::Struct`]) and to be borrowed in function parameters.
```

core/src/hir/type_context.rs
```diff
@@ -383,7 +383,37 @@ impl TypeContext {
...
+
+                if let Some(super::ParamSelf {
+                    ty: super::SelfType::Struct(s),
+                    attrs: _attrs,
+                }) = &method.param_self
+                {
+                    if let Some(b) = s.owner {
+                        let ty = s.resolve(self);
+                        if !ty.attrs.abi_compatible {
+                            // TODO: Remove once C, C++ support is in
```
Manishearth: nit: file a followup, link to it

core/src/hir/type_context.rs
```diff
@@ -409,19 +449,39 @@ impl TypeContext {
...
+                    _ => unreachable!(),
+                }
+            }
+            hir::Type::Struct(st) => {
+                if st.owner().is_some() {
+                    let ty = self.resolve_type(st.id());
+                    match ty {
+                        TypeDef::Struct(st) => {
+                            if !st.attrs.abi_compatible {
+                                // TODO: Remove once C and C++ have this support.
```
Manishearth: nit: same

Manishearth (approved):
Looks great, thanks for your patience with the design discussions!

Very minor nits; feel free to hit merge yourself once they are fixed.

PR #923, opened 2025-07-29
Improved Macro arg parsing

tyler-zeromatter:
This greatly expands `macros.rs` to allow for more complex parsing of arguments. A lot of this is built on the assumption that rustc will catch the superset of all errors related to macro parsing, so we don't have to match that superset. We're also heavily reliant on `syn` to parse a lot of this Rust syntax for us.

This still does not support multiple arms, or `$($...)*?+` parsing. But it's a decent step towards that sort of support.

TODOs:

- [x] Rigorous testing of each `MacroFragSpec` type.
- [X] ~~`pat` MacroFragSpec parsing~~ (Parsing looks to be too complicated for right now, can be done in a subsequent PR)
- [X] MacroMatcher sub-parsing (basically more delimited parsing)
- [ ] File issues for follow ups


tyler-zeromatter:
Adding validation for MacroUse evaluation in the `macro` crate would be good for a subsequent PR

core/src/ast/macros.rs
```diff
@@ -105,7 +148,7 @@ pub struct MacroIdent {
     /// This is a very hands-off approach to parsing arguments in the usage of macros, and will work in most cases.
     /// However, this makes some valid macro frag specs unusable in the current parser (I believe :block doesn't work, for instance).
     /// The hope is that this will eventually be used in a more advanced parser.
```
ambiguousname: Docs

Manishearth:
thought: we might want to consider turning this into a (not-versioned-with-diplomat) library, eventually.

core/src/ast/macros.rs
```diff
@@ -114,43 +153,342 @@ impl Parse for MacroIdent {
...
         input.parse::<Token![:]>()?;
         let ty: Ident = input.parse()?;
-        Ok(Self { ident, _ty: ty })
+        Ok(Self { ident, ty })
+    }
+}
+
+/// Hack to read information from the TokenTree while saving the rest to be later copied into a buffer.
+/// Will attempt to read `T`, and on a success all of the unread tokens after the stream will be written to `remaining`.
+struct MaybeParse<T: Parse> {
```
Manishearth: nit: ParsePartial or something.

core/src/ast/macros.rs
```diff
@@ -114,43 +153,342 @@ impl Parse for MacroIdent {
...
     }
 }
 
 #[derive(Debug)]
 #[non_exhaustive]
-/// Struct for defining a macro (i.e., `macro_rules! example`)
-pub struct MacroDef {
-    pub match_tokens: HashMap<Ident, usize>,
-    pub body: TokenStream,
+pub struct MacroUse {
```
Manishearth, line 197: nit: docs, please.

core/src/ast/macros.rs
```diff
@@ -114,43 +153,342 @@ impl Parse for MacroIdent {
...
+                    ),
+                ));
+            }
+        }
+        Ok(())
+    }
+}
+
+#[derive(Debug)]
+pub enum MacroMatch {
```
Manishearth, line 386: nit: docs

core/src/ast/macros.rs
```diff
@@ -114,43 +153,342 @@ impl Parse for MacroIdent {
...
+                format!("Did not recognize token. {input:?}"),
+            ))
+        }
+    }
+}
+
+#[derive(Debug)]
+/// Initially constructed as the root of a macro's tokens. I.e., for `example!(...)`` the macro matcher is `(...)`
+pub struct MacroMatcher {
+    pub delim: proc_macro2::Delimiter,
```
Manishearth, line 454: nit: docs, I thought this had to do with the comma

core/src/ast/macros.rs
```diff
@@ -114,43 +153,342 @@ impl Parse for MacroIdent {
...
+        args: &mut HashMap<Ident, MacroFrag>,
+    ) -> syn::Result<()> {
+        let mut buf = TokenBuffer::new2(stream);
+        let mut c = buf.begin();
+
+        let mut match_iter = matches.iter();
+        while let Some((tt, next)) = c.token_tree() {
+            let curr_match = match_iter.next();
+            match curr_match {
+                Some(MacroMatch::Ident(i)) => match i.ty.to_string().as_str() {
```
Manishearth, line 224: suggestion (nb): this can probably be turned into a method on MacroMatch, like `consume_tokens()`

Manishearth (approved):
Overall this looks pretty good! I would like some more comments and there's some minor changes to make (plus a suggestion for a larger one)

core/src/ast/macros.rs
tyler-zeromatter: I've held off on doing that since there's a lot of troublesome lifetimes to deal with. `Cursor<'_>` needs a `TokenBuffer` to borrow from, and `ParsePartial` means we are continually changing the contents of our buffer.

core/src/ast/macros.rs
```diff
@@ -192,7 +192,9 @@ where
 
 #[derive(Debug)]
 #[non_exhaustive]
+/// Constructed in [`MacroUse::parse`] (called by [`MacroDef::evaluate`]) when a previously defined macro is used. We use the definition of [`MacroDef`] to then construct the args in the used macro.
```
Manishearth, line 196: I'd still like a description of what this _is_ at the top. We have multiple MacroFoo types, it should be clear what they represent.

I think a first line doc comment of "A macro use-site" would be fine, and then you can have more docs below

core/src/ast/macros.rs
```diff
@@ -377,6 +379,8 @@ impl MacroUse {
 }
 
 #[derive(Debug)]
+/// Used for determining how to parse [`MacroUse`]. This covers everything inside of the brackets for a macro's definition (i.e., example!(...), everything inside of `...` is parsed as a MacroMatch.
```
Manishearth: ditto

core/src/ast/macros.rs
```diff
@@ -441,7 +445,8 @@ impl Parse for MacroMatch {
 }
 
 #[derive(Debug)]
-/// Initially constructed as the root of a macro's tokens. I.e., for `example!(...)`` the macro matcher is `(...)`
+/// A MacroMatcher is a delimited list of [`MacroMatch`]es. When you call the macro `example!(...)`, the macro matcher is the parentheses delimited tokens: `(...)`.
```
Manishearth, line 451: ditto

core/src/ast/macros.rs
tyler-zeromatter: Let me know if these look better

Manishearth (approved):

PR #924, opened 2025-07-30
Make Callbacks and Traits use ReturnType

tyler-zeromatter:
I've adjusted the backends that use callbacks (C, C++, Kotlin) to use ReturnType, but I panic for anything that is not `ReturnType::Infallible(SuccessType::Unit | SuccessType::OutType(..))`. Not sure if we want to make this a full on `supports=` flags for non-infallible types (edit: we should).

TODOs:

- [x] `supports=` flag for Fallible/Nullable Return types in traits/callbacks.
- [x] Test DiplomatOption<&Opaque> 
- [x] Test DiplomatResult<&Opaque, ()>
- [x] Test conversion of types inside of DiplomatResult (i.e., `-> Result<Struct, ()>`).
- [x] Test conversion of strings and slices
- [x] Fix C++ designated initializers
- [x] Add full support to C/C++ (should this be a separate PR?)
  - [x] Tests
- [x] Fix Nanobind CI (seems like return types aren't being converted properly)
- [x] Nullable represents DiplomatOption, so support needs to be added back in to C/C++.
- [x] Fix Nanobind building on MSVC not working.
- [x] Test error `&Opaque` types (in C++ *and* Nanobind)
- [x] Fix C++ CAPI callback namespacing in template
- [x] Docs

### DIFF
--- a/tool/src/c/ty.rs
+++ b/tool/src/c/ty.rs
@@ -3,8 +3,8 @@ use super::header::Header;
 use crate::ErrorStore;
 use askama::Template;
 use diplomat_core::hir::{
-    self, CallbackInstantiationFunctionality, MaybeOwn, OpaqueOwner,
-    StructPathLike, SymbolId, TraitIdGetter, TyPosition, Type, TypeDef, TypeId,
+    self, CallbackInstantiationFunctionality, MaybeOwn, OpaqueOwner, StructPathLike, SymbolId,
+    TraitIdGetter, TyPosition, Type, TypeDef, TypeId,
 };
 use diplomat_core::hir::{ReturnType, SuccessType, TypeContext};
 use std::borrow::Cow;


tyler-zeromatter:
It looks like Kotlin doesn't support `DiplomatOption` for callbacks to begin with, so I think I'm good to not add support there.

tyler-zeromatter:
C support is in, now C++ support needs some additional work to convert `diplomat::result` and `std::optional` to the C ABI.

tyler-zeromatter:
Option 3 could be we wrap the callback in another callback, to do the conversions ourselves. This is probably the best middle of the road approach.

tyler-zeromatter:
Yeah, I'm going to keep the above code for conversion, but I think wrapping around the callbacks is the best way to go.

tyler-zeromatter:
Ahh, wait I missed the fn_traits struct, this makes things easier.

Manishearth:
> Not sure if we want to make this a full on supports= flags for non-infallible types (edit: we should).

I don't think we should; this is the type of thing that I'm fine with being a backend panic. Generally I don't consider this type of feature to be "special", though I guess catching and wrapping exceptions in a backend like JS might not always be possible.

> 2\. On `diplomat::result`, create a templated `c_value` struct that will match the C ABI exactly. Make the user call `diplomat_result->as_ffi()`, and tell the `extern "C"` part of the API to accept this `c_value` type. I've tested this so far with just `diplomat::result<std::monostate, std::monostate>`, and it's working.

This would require `diplomat::result` to know how to convert back to C

> Ahh, wait I missed the fn_traits struct, this makes things easier.

That _should_ work, I think, but another way of doing it would be to add a `diplomat::result->ToFFI()` function that takes in two closures that let you perform conversions.

tyler-zeromatter:
Yeah, I think that makes more sense to avoid having to intercept the return type with clever templating.

tyler-zeromatter:
Having some issues with returning FFI friendly Result types from C to Rust, I'm curious how we're handling that with top-level method return types from Rust -> C

core/src/ast/types.rs
```diff
@@ -720,6 +721,16 @@ impl TypeName {
         }
     }
 
+    pub fn callback_ret_to_syn(&self) -> syn::Type {
```
Manishearth: issue: I think this should be a part of `ffi_safe_version()`

(unless I'm missing something)

Manishearth: That will also correctly handle `Result<Option<Foo>>` and friends

core/src/hir/attrs.rs
```diff
@@ -1062,6 +1062,8 @@ impl BackendAttrSupport {
             "traits_are_send" => Some(self.traits_are_send),
             "traits_are_sync" => Some(self.traits_are_sync),
             "abi_compatibles" => Some(self.abi_compatibles),
+            "struct_refs" => Some(self.struct_refs),
+            "fallible_trait_callbacks" => Some(self.fallible_constructors),
```
Manishearth: thought: I don't think we need a toggle for this.

The point of these toggles is so that you can write multi-backend Diplomat code that still falls back to something sensible when something isn't supported.

By and large, this should be things that just _don't_ translate cleanly to all languages: e.g. not all languages have constructors, and not all languages with constructors allow the constructors to return errors. Not all languages have a builtin iterator concept. Some languages handle strings differently and the most efficient API for them is different.

However, I have a hard time seeing languages that support callbacks and fallibility not supporting diplomat callbacks; this is a matter of missing functionality as opposed to impossible functionality. For that I'd prefer to just have the backends error: Diplomat is okay with backends throwing additional support errors; you can always `cfg(kotlin, disable)` or whatever.

core/src/hir/lowering.rs
```diff
@@ -1715,6 +1727,329 @@ impl<'ast> LoweringContext<'ast> {
         .map(|r_ty| (r_ty, return_ltl.finish()))
     }
 
+    // TODO: Merging this with lower_out_type somehow would be great.
```
Manishearth: question: what's the reason it needs to be separate? Is it mostly because of `Type<OutputOnly>` and `Type<InputOnly>` (yeah, we have to add a lot more trait things to TyPosition to make that type of thing work)

It is disappointingly a lot of code. Would it at least be possible to write this so that it internally calls `lower_type` for non-Result types? I think `lower_type` should just work here.

(not going to review this code in detail yet)

tool/src/cpp/ty.rs
```diff
@@ -805,12 +804,56 @@ impl<'ccx, 'tcx: 'ccx> TyGenContext<'ccx, 'tcx, '_> {
...
+
+                        let err_type_name = match err {
+                            Some(o) => match o {
+                                Type::Primitive(ref p) => self.formatter.fmt_primitive_as_c(*p).into(),
+                                _ => self.formatter.fmt_type_name(o.id().unwrap()).to_string(),
+                            }
+                            None => "std::monostate".into(),
+                        };
+
+                        let return_type = format!("diplomat::capi::DiplomatCallback_{}_{cpp_name}_result", method_abi_name.unwrap());
```
Manishearth: thought: this should probably be a method on the formatter.

tool/src/cpp/ty.rs
```diff
@@ -805,12 +804,56 @@ impl<'ccx, 'tcx: 'ccx> TyGenContext<'ccx, 'tcx, '_> {
...
+
+                        format!("diplomat::fn_traits({cpp_name}).c_run_callback_result<{ok_type_name}, {err_type_name}, {return_type}>")
```
Manishearth: also this, maybe? Where you give it the ok/err types

tool/src/kotlin/mod.rs
```diff
@@ -1163,25 +1163,33 @@ returnVal.option() ?: return null
...
+                                    Type::Enum(..) => ".toNative()",
+                                    Type::Struct(..) => ".nativeStruct",
+                                    _ => "",
+                                }
+                                .into(),
+                            ),
+                            SuccessType::Unit => ("Unit".into(), "".into()),
+                            _ => panic!("Unsupported success type {success:?}"),
+                        },
+                        _ => panic!("Unsupported return type {output:?}"),
```
Manishearth: suggestion: perhaps explicitly say Result returns are not supported

tool/src/kotlin/mod.rs
```diff
@@ -1163,25 +1163,33 @@ returnVal.option() ?: return null
...
-                            match ty {
-                                Type::Enum(..) => ".toNative()",
-                                Type::Struct(..) => ".nativeStruct",
-                                _ => "",
-                            }
-                            .into(),
-                        ),
-                        None => ("Unit".into(), "".into()),
+                    let (native_output_type, return_modification) = match &**output {
+                        ReturnType::Infallible(success) => match success {
```
Manishearth, line 1167: issue: We unfortunately lose support for returning options over callbacks here, since what used to be modeled as a `Type::DiplomatOption` will now be modeled as a `ReturnType::Nullable`.

We should try and retain that support here. A hack where you turn nullable returns into infallible DiplomatOption returns is fine.

tool/templates/cpp/runtime.hpp.jinja
```diff
@@ -167,6 +168,64 @@ public:
       return result<T2, E>(Ok<T2>(std::move(t)));
     }
   }
+  
+  template<typename T2, typename U = T, typename V = E, typename std::enable_if_t<!std::is_same_v<U, std::monostate> && !std::is_same_v<V, std::monostate>, std::nullptr_t> = nullptr>
```
Manishearth: thought: I think this should be a helper function; I don't want internal capi stuff on the `result` public API.

tool/templates/cpp/runtime.hpp.jinja
```diff
@@ -299,6 +358,49 @@ template <typename Ret, typename... Args> struct fn_traits<std::function<Ret(Arg
     static Ret c_run_callback(const void *cb, replace_fn_t<Args>... args) {
         return (*reinterpret_cast<const function_t *>(cb))(replace<Args>(args)...);
     }
+    
+    template<typename T, typename E, typename T2>
+    static T2 c_run_callback_result(const void *cb, replace_fn_t<Args>... args) {
+      result<T, E> out = c_run_callback(cb, args...);
+      return out.template to_ffi<T2>();
```
Manishearth: question: where do the types inside Ok/Err get converted? I don't see any conversion call here or in the cpp backend code, did I miss something?

Basically, `-> Result<Struct, Err>` should work: where does Struct::ToFFI() get called?

Manishearth (commented):
Apologies for the delay, I'm rather busy with Temporal stuff.

Overall this looks good; it's definitely not too complicated, which is nice. Couple issues/comments

core/src/hir/attrs.rs
tyler-zeromatter: Hmm, I thought I removed this but it seems to have been re-introduced.

core/src/hir/lowering.rs
tyler-zeromatter: It is because of `Type<OutputOnly>` vs. `Type<InputOnly>`. I'll see what's easy for me to do right now and will update.

Manishearth: Yeah, we already have code that lowers `Type<InputOnly>`; so hopefully we can use it.

tyler-zeromatter: We also need to account for `OutStruct`, but I think that's doable.

tyler-zeromatter: Looks like `lower_type` also has some generics problems that make it difficult to use inside of `lower_out_type`. `OpaquePath` and `StructPath` have constructors that need to be specified with `P: TyPosition<StructPath = StructPath, OpaqueOwnership = Borrow>`

I think there's no real way around adding traits to avoid all this repetition, regardless of which function we're calling.

tyler-zeromatter: So I guess the question is, should I be working on all of these traits now, or in a different PR?

I think it would be better to tackle all of this in a follow-up PR, since there's already a lot in this one.

Manishearth: > Looks like `lower_type` also has some generics problems that make it difficult to use inside of `lower_out_type`. `OpaquePath` and `StructPath` have constructors that need to be specified with `P: TyPosition<StructPath = StructPath, OpaqueOwnership = Borrow>`

Right, that makes sense: output types cover structs as well, input types cover callbacks as well. `lower_out_type` isn't what we want here; we want to merge `lower_type` and `lower_callback_out_type`. Callback outputs are input types.

I'd still like to see if this is possible; I can perhaps try tomorrow. But yes, we can do this as a followup.

Overall I don't think there's any reason we should have three such lower functions: two is not-great but understandable, three is a bit much.

tyler-zeromatter: Ah, I think merging those two makes sense. There's still a lot of gotchas when it comes to what I'd accept as a callback return type vs. an input type, but I did a little bit of merging the two to avoid repetition.

Should see that when I next push.

tyler-zeromatter: Pushed

tool/src/kotlin/mod.rs
tyler-zeromatter: I looked into this. The function `gen_native_type_name` (which was called in the original version of this PR irrespective of type) calls `unreachable!` for `DiplomatOption`. I don't think DiplomatOptions were supported by callbacks in Kotlin to begin with.

tool/templates/cpp/runtime.hpp.jinja
tyler-zeromatter: Ah I missed that, thanks. I'll add a test.

tool/src/kotlin/mod.rs
tyler-zeromatter: But I'm still happy to add the hack if that still sounds good

tool/templates/cpp/runtime.hpp.jinja
tyler-zeromatter: Note to self: I think I can just get away with calling `replace(out)`. For both result and std::optional

tyler-zeromatter: Actually, @Manishearth where does conversion happen for structs during regular calls? Where do we do `-> Struct` conversion to FFI during a regular c_run_callback?

Manishearth: `Struct`s never get converted; they are always FFI safe. It is other types that may not be.

(as a consequence, even though you can use `Option` and `&[u8]` over FFI, you will be forced to use DiplomatOption, DiplomatSlice, etc when in a struct0

tool/src/kotlin/mod.rs
Manishearth: Ah, hm. That's fine then!

tyler-zeromatter:
Note that while Nanobind CI shows a success here, it's actually failing on windows. Will comment when that's fixed.

core/src/hir/lowering.rs
Manishearth: It looks like this function is still separate: am I missing something? Where's the merging?

macro/src/lib.rs
```diff
@@ -39,6 +39,16 @@ fn param_ty(param_ty: &ast::TypeName) -> syn::Type {
     }
 }
 
+fn callback_return_conversion(ty: &ast::TypeName) -> TokenStream {
```
Manishearth: issue: You shouldn't need this, see calls of `is_ffi_safe()` in the macro and do the same thing.

You may need to add Result to `is_ffi_safe()`.

By and large the macro shouldn't need to actually deal with this stuff; it should just call methods on core.

tool/src/cpp/formatter.rs
```diff
@@ -205,6 +205,92 @@ impl<'tcx> Cpp2Formatter<'tcx> {
...
+
+    pub fn fmt_callback_success_type(&self, success: &hir::SuccessType<hir::InputOnly>) -> String {
+        match success {
+            hir::SuccessType::Unit => "std::monostate".into(),
+            hir::SuccessType::OutType(o) => self.fmt_callback_out_type(o),
+            _ => unreachable!("unknown AST/HIR variant"),
+        }
+    }
+
+    pub fn fmt_callback_out_type(&self, ty: &hir::Type<hir::InputOnly>) -> String {
```
Manishearth: issue: this is incorrect; it should be a `gen_` method since it may need to generate imports.

Is there a reason `gen_type_name()` does not work for this? We should not need to add any such large matches to support Result.

tyler-zeromatter: Mostly because `gen_cpp_to_c_for_type` does not take a mutable `&self` reference. I'll draft a version that uses `gen_type_name`, though.

core/src/hir/lowering.rs
tyler-zeromatter: It's a partial merge, it uses `lower_type` in the last arm (for Primitive, Named, Ordering, and Self types). Part of the problem is that `lower_callback_out_type` needs to error on certain types that `lower_type` supports. I can add a `context` enum to `lower_type` if you think that would help clean things up.

Manishearth:
So I'm trying to figure out the differences between `lower_callback_out_type` and `lower_type`.

My goal is that `lower_type` should manage _all_ of the differences, using `TypeLoweringContext` as needed to differentiate.

So far here's what I've noticed:

 - `lower_callback_out_type` does not support struct references. Which makes sense, though we _might_ be able to get away with still lowering this type and having the lifetime checker _later_ validate it.
 - `lower_callback_out_type` does not support Box
 - (wip)

Manishearth:
Okay, I think it works. If there are cases I missed, it should be easy enough to handle them with the check.

By and large we should only ever need two such `lower_type` functions: one for types that go from Rust to FFI, and one for the other way around. There are some differences between the two which makes having two such functions useful, but anything further can be handled by a context enum and checked in situ.

In theory we could even merge `lower_type` and `lower_out_type`, but then we'd need to expand the traits to be able to generically construct various TyPosition-dependent types, and that's less clean.

PR #928, opened 2025-08-08
Nanobind Overloading Support

tyler-zeromatter:
Since we never know when an overload will happen (and all overloaded functions need to know their argument types), this requires us to call `gen_type_name` consistently for all functions, which may destroy some of the functionality Nanobind is trying to avoid.

I can either leave this as-is, or maybe we could re-factor how `includes` are used within `gen_type_name`, maybe added to `ParamInfo`? Then the includes are only processed if we know for certain they will be used in the function.


tyler-zeromatter:
There's also quite a bit of code re-use from C+++. I think being able to use C++'s functions instead would make things a lot easier and avoid a lot more breaking changes.

Walter-Reactor:
I like the idea simplification of how gen_type_name gets called, and your proposal of adding includes to ParamInfo so we can more accurately use only the required ones, rather than adding one for every type that ever gets inspected.

Walter-Reactor:
@Manishearth would have to chime in on the code-reuse issue, I ran into it before & the thought was that we wanted to keep things separate since the similarity of structure was sort of superficial at the time, but increasingly since the python generator is, under the hood, actually a C++ generator, I'm thinking maybe that's a discussion worth revisiting

feature_tests/src/attrs.rs
```diff
@@ -211,6 +211,14 @@ pub mod ffi {
             Box::new(Self { x, y })
         }
 
+        #[diplomat::attr(*, rename="make")]
```
Manishearth: nit: this needs to be `supports = method_overloading`

I'm surprised it works in JS, unless we added overload support to JS.

tool/src/nanobind/ty.rs
```diff
@@ -457,6 +455,36 @@ impl<'ccx, 'tcx: 'ccx> TyGenContext<'ccx, 'tcx> {
...
+        if def.attrs().disable {
+            self.errors
+                .push_error(format!("Found usage of disabled type {type_name}"))
+        }
+
+        self.includes
+            .insert(self.formatter.cxx.fmt_impl_header_path(id));
+        if let Some(borrow) = st.owner() {
+            let mutability = borrow.mutability;
+            match (borrow.is_owned(), false) {
```
Manishearth: issue: Structs should never be pointer-owned; `Borrow::is_owned()` returns false unconditionally.

Manishearth: Also, come to think of it, `StructPathLike::owner` should return a `MaybeOwn`, not an `Option<Borrow>`

Manishearth: I'm working on this.

Manishearth: https://github.com/rust-diplomat/diplomat/pull/929

feature_tests/src/attrs.rs
tyler-zeromatter: I haven't tested it with JS, so I'll just make sure to mark this appropriately.

tyler-zeromatter: Oh, it's not generating with JS since the struct is disabled with `not(supports = arithmetic)` to begin with. Still, I've marked the function appropriately.

Manishearth: Thanks!

PR #936, opened 2025-08-11
Refactor Nanobind use C++ generation

tyler-zeromatter:
Following from discussion in https://github.com/rust-diplomat/diplomat/pull/928


Manishearth (approved):

tyler-zeromatter:
Waiting on approval by @Walter-Reactor before merging

tool/src/nanobind/mod.rs
```diff
@@ -110,26 +110,53 @@ pub(crate) fn run<'cx>(
...
+        let mut header = crate::cpp::Header {
+            path: cpp_decl_path.clone(),
+            includes: BTreeSet::new(),
+            decl_include: None,
+            body: "".into(),
+            indent_str: "",
+            forwards: BTreeMap::new(),
+        };
+
+        let mut unused_header = crate::cpp::Header {
```
walter-zeromatter: Would it be possible to use Default:;default() here, and in the other instances below since this is going to be unused? Would save significantly on boilerplate

tyler-zeromatter: Both look unused actually, so I've used `Default::default` for both

PR #937, opened 2025-08-12
Free Functions support

tyler-zeromatter:
The changes to the HIR/AST were fairly minor, I found repurposing `Method` to be very useful (since it contains all the same information as a function, just with the guarantee that `self_param` will never be used).

Re-writing the backends is proving somewhat tricky however, since each comes with the assumption that any function being generated must be associated with a type. TODOs:

- [ ] Fix C++ backend (Figure out borrowing of `MethodInfo` if possible)
- [ ] Fix Nanobind backend
- [ ] Tests


core/src/hir/type_context.rs
```diff
@@ -68,6 +74,7 @@ pub enum TypeId {
 pub enum SymbolId {
     TypeId(TypeId),
     TraitId(TraitId),
+    FunctionId(FunctionId),
```
Manishearth, line 77: nit: please document on `hir::Method` that it can be used for a function or a method

tool/src/cpp/imp.rs
```diff
@@ -0,0 +1,354 @@
...
+    /// DiplomatWrite, if present, is saved to a variable named `output`.
+    pub(super) c_to_cpp_return_expression: Option<Cow<'a, str>>,
+
+    /// If the method returns a writeable, the info for that
+    pub(super) writeable_info: Option<MethodWriteableInfo<'a>>,
+    pub(super) docs: String,
+}
+
+#[derive(Template, Default)]
+#[template(path = "cpp/impl_block_impl.h.jinja", escape = "none")]
```
Manishearth: I don't really understand why these are impls. They're free functions, yes? They're explicitly not impls.

tyler-zeromatter:
I added a `supports=` attribute, but I didn't add any erroring logic since the backends can choose to ignore all free functions if they so wish.

tyler-zeromatter:
Hi @Manishearth, wanted to check to see if there's anything you'd like me to address here?

core/src/hir/type_context.rs
```diff
@@ -54,6 +56,10 @@ pub struct EnumId(usize);
 #[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct TraitId(usize);
 
+/// Key used to index into a [`TypeContext`] representing a function.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct FunctionId(usize);
```
Manishearth, line 61: issue: we shouldn't actually need this, I think. Functions can't be referenced elsewhere.

macro/src/lib.rs
```diff
@@ -241,12 +241,35 @@ fn gen_custom_type_method(strct: &ast::CustomType, m: &ast::Method) -> Item {
...
+    func_ident: Ident,
+    extern_ident: Ident,
+    self_param: Option<&'a crate::ast::SelfParam>,
+    params: &'a [crate::ast::Param],
+    return_type: Option<&'a crate::ast::TypeName>,
+    lifetime_env: &'a crate::ast::LifetimeEnv,
+    attrs: &'a crate::ast::Attrs,
+}
 
+fn gen_custom_function(func_info: FuncGen) -> Item {
```
Manishearth, line 268: nit: make this a method on FuncGen

tool/src/c/mod.rs
```diff
@@ -119,5 +122,39 @@ pub(crate) fn run<'tcx>(
     }
     // loop over traits too
 
+    // Loop over free functions, put them all in one file (currently this is diplomat_runtime.h):
```
Manishearth, line 124: issue: diplomat_runtime.h should not vary between generations

This should be a separate file, potentially requiring a config-provided `lib.name`.

tool/src/cpp/ty.rs
```diff
@@ -290,7 +247,7 @@ impl<'ccx, 'tcx: 'ccx> TyGenContext<'ccx, 'tcx, '_> {
         let methods = def
             .methods
             .iter()
-            .flat_map(|method| self.gen_method_info(id, method))
+            .flat_map(|method| FuncGenContext::gen_method_info(id.into(), method, self))
```
Manishearth: issue: I don't think this is correct; the Headers for the TyGenContext get ignored.

I'm not really convinced of the TyGenContext / FuncGenContext split.

core/src/hir/type_context.rs
tyler-zeromatter: This is mostly used for `SymbolId`, since a lot of functions can be re-written to use `SymbolId` for formatting names and the like.

I'm happy to refactor though if you want it gone.

tool/src/cpp/ty.rs
tyler-zeromatter: I'm a little unsure of what you mean, how are they ignored?

core/src/hir/type_context.rs
Manishearth: > This is mostly used for SymbolId, since a lot of functions can be re-written to use SymbolId for formatting names and the like.


Hmm. I think having separate format functions is fine. The current `fmt_symbol_name` just proxies an existing method for Type and it panics on Trait; I don't think that's a particularly useful addition right now.

tool/src/cpp/ty.rs
Manishearth: Ah, I see what's happening here.

This is rather confusing and I'm not really sure if FuncGenContext serves any useful purpose here.

TyGenContext and FuncGenContext both have decl/impl headers. It seems like we unconditionally push to TyGenContext's headers list. What is FuncGenContext doing, then? We spend a bunch of work constructing them and sticking them in a map, but that just seems off.

I think I'd prefer for the churn moving things off of TyGenContext to be undone. TyGenContext can really just be GenContext and we can have an inner enum for type-specific stuff if we need.

(This also just makes this PR messier to review: there are a lot of changes, some of which just move code around. That's often fine if the PR is reviewable commit-by-commit, but this isn't)

core/src/hir/type_context.rs
```diff
@@ -209,7 +196,7 @@ impl TypeContext {
         match id {
             SymbolId::TypeId(id) => self.fmt_type_name_diagnostics(id),
             SymbolId::TraitId(id) => self.resolve_trait(id).name.as_str().into(),
-            SymbolId::FunctionId(id) => self.resolve_function(id).name.as_str().into(),
+            SymbolId::Function => panic!("Functions not supported."),
```
Manishearth: ah, I see how this is used. For diagnostics I guess it's fine, I think you can add the function id back.

I was asking to see if it was used anywhere and I couldn't see it, but this counts as use. It makes sense that this owuld be found in diagnostics, it doesn't make sense that it would be found elsewhere since by and large we do not need to generically resolve function symbols

tyler-zeromatter: Alright, will revert then.

tyler-zeromatter: Would you still want the use of `SymbolId` for other helpers like formatting and function generation to be separated out?

Manishearth: Yes please, I'd rather not be using SymbolId everywhere unless it actually helps

Manishearth: Narrowly, for method_info context it makes sense.

tool/src/c/ty.rs
```diff
@@ -593,4 +529,94 @@ impl<'tcx> TyGenContext<'_, 'tcx> {
 
         ty_name
     }
+
+    pub fn gen_method(
+        &self,
+        method: &'tcx hir::Method,
+        header: &mut Header,
+        template: &mut FuncBlockTemplate,
```
Manishearth, line 537: nit: `out_template`

tool/src/c/ty.rs
```diff
@@ -593,4 +529,94 @@ impl<'tcx> TyGenContext<'_, 'tcx> {
 
         ty_name
     }
+
+    pub fn gen_method(
```
Manishearth, line 533: note to self: the main difference here is that it now takes FuncBlockTemplate

tool/src/c/mod.rs
Manishearth: I think for now doing the narrow `free_functions_file_name` config would also be fine since we have some confusion between `lib.name` and namespaces in C++ that we should resolve.

OR we _require_ free functions be namespaced. That would work just as well.

This would mean generating multiple files, one per namespace.

ambiguousname:
Closing in favor of #949

PR #944, opened 2025-08-21
Remove extension artifacts

robertbastian:
These are created by https://github.com/redhat-developer/vscode-java/issues/634, god knows why


Manishearth (approved):

robertbastian:
cleared the caches, works now

PR #951, opened 2025-09-12
Fix GCC template error & behavioral error with Result types

Walter-Reactor:
In GCC 14.2.0, the OK & Err types type show an underspecified template error when passing a trivially copiable R-value, such as the common std::monostate.
This pointed out to me that due to universal referencing rules, we were actually defining the T& constructor *twice*, one explicitly as the T& constructor, and another implicitly with T&& due to the universal referencing template rules.

As a fix, instead of defining the T& constructor for trivially copyable types, we instead delete it for *non* trivially copiable types.

This is more correct, shows proper behavior, and fixes the GCC error


Walter-Reactor:
Closing temporarily, a deeper dive shows that this actually wasn't a universal reference after all, but correcting it shows some other troublesome behavior when the type of T is a &

PR #966, opened 2025-09-23
Support iterators over non-owning pointer types in C++

brendan-zeromatter:
We noticed that iterators returning reference types such as:

```
#[diplomat::attr(auto, iterator)]
pub fn next(&'a mut self) -> Option<&'a OpaqueType>
```

were failing to compile in C++, because this `next_to_iter_helper` expects owned `unique_ptr`'s or `optional`'s, but was being given value_types of `T*`. This created some type mismatches in `operator*` because `T*` was not properly decayed into `T` by `inner`. In fact, the base case of `inner { using type = T; }` will pretty much never work, because there is no "inner" type to be dereferenced.

Specific compiler error: (user types replaced with `T` for clarity)

```
no viable conversion from returned value of type 'const T' to function return type 'const value_type' (aka 'const T *const')
  501 |   const value_type& operator*() const { return *_curr; }
```


feature_tests/src/attrs.rs
```diff
@@ -206,6 +211,32 @@ pub mod ffi {
...
+        }
+    }
+
+    #[diplomat::opaque]
+    #[diplomat::attr(not(supports = iterators), disable)]
+    struct OpaqueRefIterator<'a>(std::slice::Iter<'a, AttrOpaque1>);
+    impl<'a> OpaqueRefIterator<'a> {
+        #[diplomat::attr(auto, iterator)]
+        pub fn next(&'a mut self) -> Option<&'a AttrOpaque1> {
+            self.0.next()
```
brendan-zeromatter, line 236: The other test cases have a `.map(Box::new)` or a `.copied()` here, that's the main difference. This one uses the natural &T returned from the Rust iterator.

PR #977, opened 2025-10-05
Only use wasm_glue on wasm32-unknown-unknown, not WASI

tschneidereit:
The functional change here is a one-line fix, with which all functionality (at least as covered by tests) works under both WASIp1 and p2.

The second commit contains changes to add CI coverage. I'm happy to remove that, or only test one of the targets, if you don't want to spend the CI resources, of course.


Manishearth (commented):
Could you say more about what's broken here? wasm_glue does a bunch of things: it sets up logging and panic handling. WASI targets would also need this type of thing, yes?

I don't mind disabling this for WASI right now; but really what this should be is behind a JS feature, rather than behind a wasm target thing; because these are useful when you wish to use Diplomat-generated code from JS, regardless of the actual platform.

(Don't make that change here, we can make it as a part of stabilizing diplomat-runtime)


I don't think the tests will do much: `cargo test` largely runs tests for the codegen framework; the runtime isn't tested. `cargo make test-js` does runtime testing of the WASM backend; and that would need to be given a `test-js-wasi` mode or something.

tschneidereit:
> Could you say more about what's broken here? wasm_glue does a bunch of things: it sets up logging and panic handling. WASI targets would also need this type of thing, yes?

They wouldn't, no: from this perspective they're basically just normal, vaguely POSIX-y targets that use normal approaches to logging, etc.

> I don't mind disabling this for WASI right now; but really what this should be is behind a JS feature, rather than behind a wasm target thing; because these are useful when you wish to use Diplomat-generated code from JS, regardless of the actual platform.

Agreed, yes: targeting `wasm32-whatever` obviously doesn't guarantee the availability of a `console.log` import, and indeed a feature that not only enables the glue module but also ensures that `wasm-bindgen` is configured properly makes much more sense.

> I don't think the tests will do much: `cargo test` largely runs tests for the codegen framework; the runtime isn't tested. `cargo make test-js` does runtime testing of the WASM backend; and that would need to be given a `test-js-wasi` mode or something.

Yeah, what I really wanted to do here is to show that the `wasip*` targets fully work as already supported native targets do. Meaning it's possible to ship wasm builds of all the different parts and have them behave like their native counterparts.

For the runtime itself, it seems like an okay state of affairs to have the same test coverage as other targets? And see the use case motivating all of this: with this change, `diplomat-runtime` works as expected when used in SpiderMonkey via `icu_cabi`, and passes the test suite for the `Intl` API.

(Also, hi Manish, it's been a very long time! 👋🏻)

Manishearth:
> They wouldn't, no: from this perspective they're basically just normal, vaguely POSIX-y targets that use normal approaches to logging, etc.


Right, I'm talking about WASI-from-JS. I think the real distinction here is "where are you using diplomat from", not "what platform are you using it from", and target(wasm32-unknown-unknown) is a poor proxy for that. It's a fine proxy for the next release of Diplomat.


> Yeah, what I really wanted to do here is to show that the wasip* targets fully work as already supported native targets do. Meaning it's possible to ship wasm builds of all the different parts and have them behave like their native counterparts.

We get used on a lot of strange targets, so I don't think we need to CI these, specifically. We just don't really have `cargo test` tests for such targets.

> (Also, hi Manish, it's been a very long time! 👋🏻)

Hi! Yes, it has, hope stuff is going well for you!

tschneidereit:
> So I'd suggest removing the test changes and keeping the target change. I've already included fixing this as on the cards for diplomat-runtime 1.0 (#958)

Done!

> Hi! Yes, it has, hope stuff is going well for you!

It is, and I hope the same for you!

Manishearth:
Published 0.13.1

tschneidereit:
@Manishearth unfortunately the a `0.13` patch release doesn't help me, because I'd have to update `icu_cabi` in Firefox, then let that percolate through to mozjs some unknown time in the future. That's why I was asking about the `0.8.3` release, specifically. (And I do understand that doing that release is annoying, and apologize for even asking.)

Manishearth:
Published.

Does Gecko have a way to do local patches of vendored code? I _highly_ recommend having such a setup, we use it a lot in Chromium, Android, and internal Google codebases. I know Gecko didn't have that back in the day, but I'm not clear on the current status.

tschneidereit:
> Published.

Thank you! ❤️

> Does Gecko have a way to do local patches of vendored Rust code? I _highly_ recommend having such a setup, we use it a lot in Chromium, Android, and internal Google codebases. I know Gecko didn't have that back in the day, but I'm not clear on the current status.

Gecko by now vendors all Rust code, yes. My use case isn't for Gecko itself though, but for a [project](https://github.com/bytecodealliance/StarlingMonkey/) using/soon to use mozjs. And I really don't want to switch to vendoring things in StarlingMonkey.

PR #1014, opened 2026-01-14
Add Support for Custom Bindings

### LINKED ISSUE #1013: Ability to include "custom" additional files
It would be nice to be able to include:

 - Additional full-fledged files (extra header files)
 - Additional code within an impl block
 - Additional code within a def block

This should be designed in such a way as to support patching in code in many different places.

tyler-zeromatter:
Closes #1013 

This is still very much a draft and open for feedback.

My TODOs:
- [x] Potential attribute re-work?
  - What limits do we impose on the attribute? 
- [x] Support def/impl block includes in C++
  - [x] Potentially add support for non-opaque def blocks
- [x] Flesh out C++ test
  - [X] File replacement
  - [x] Block replacement
- [x] Add full Nanobind support
- [x] Extend C++ test to Nanobind
- [x] Pass all tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 # Diplomat
-Diplomat is an experimental Rust tool for generating FFI definitions allowing many other languages to call Rust code. With Diplomat, you can simply define Rust APIs to be exposed over FFI and get high-level C, C++, and JavaScript bindings automatically!
+Diplomat is a Rust tool for generating FFI definitions allowing many other languages to call Rust code. With Diplomat, you can simply define Rust APIs to be exposed over FFI and get high-level C, C++, and JavaScript bindings automatically!
 
 Diplomat supports generating bindings from Rust to:
 - C
--- a/tool/src/kotlin/formatter.rs
+++ b/tool/src/kotlin/formatter.rs
@@ -179,8 +179,8 @@ impl<'tcx> KotlinFormatter<'tcx> {
             panic!("Trait methods need a name");
         }
         let name = method.name.clone().unwrap().as_str().to_lower_camel_case();
-        let name = if method.attrs.is_some() {
-            method.attrs.as_ref().unwrap().rename.apply(name.into())
+        let name = if let Some(attrs) = &method.attrs {
+            attrs.rename.apply(name.into())
         } else {
             name.into()
         };


core/src/hir/attrs.rs
```diff
@@ -59,6 +60,31 @@ pub struct Attrs {
...
+    Block(String),
+    /// Include the whole file.
+    /// The implicit assumption is that the backend will scrap whatever file it was originally going to include instead.
+    File(String),
+}
+
+/// Information specifying how a custom binding should be included.
+#[derive(Clone, Default, Debug)]
+#[non_exhaustive]
+pub struct IncludeInfo {
```
Manishearth: Please use the ontology from https://github.com/rust-diplomat/diplomat/issues/1013#issuecomment-3755100497

I think this should have an `IncludeSource { File(String), Source(String) }` and an `IncludeLocation { DefBlock, ImplBlock ,...}` and be stored in a map of location to source.

core/src/hir/attrs.rs
```diff
@@ -59,6 +63,37 @@ pub struct Attrs {
...
+#[non_exhaustive]
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub enum IncludeLocation {
+    /// An extension to the definition of the class (i.e., in C++, the .d.hpp file)
+    DefBlock,
+    /// An extension to the implementation of the class (i.e., in C++, the .hpp file)
+    ImplBlock,
+    /// An overwrite of the definition of the class.
+    /// The implicit assumption is that the backend will scrap whatever file it was originally going to include instead.
+    DefFile,
```
Manishearth: I don't think these should overwrite, if we want that it should be a different option, and probably not a part of the MVP.

tyler-zeromatter: Sounds good, the examples I'm using anyways don't require an overwrite, so I will scrap this.

core/src/hir/attrs.rs
```diff
@@ -59,6 +63,31 @@ pub struct Attrs {
     pub generate_mocking_interface: bool,
     /// From #[diplomat::attr()]. If true, Diplomat will check that this struct has the same memory layout in backends which support it. Allows this struct to be used in slices ([`super::Slice::Struct`]) and to be borrowed in function parameters.
     pub abi_compatible: bool,
+
+    /// Information on if a type declaration/impl block has custom bindings, and if so, what kind.
```
Manishearth, line 67: nit: I'd call this "custom extra code" and the field name `custom_extra_code`, here and everywhere.

These aren't necessarily bindings.

core/src/hir/attrs.rs
```diff
@@ -389,9 +418,80 @@ impl Attrs {
                             }
                             this.abi_compatible = true;
                         }
+                        "include" => {
+                            let mut source: Option<IncludeSource> = None;
```
Manishearth: issue: please separate out parsing code. See how it's been done for SpecialMethod.

core/src/hir/attrs.rs
```diff
@@ -389,9 +418,80 @@ impl Attrs {
...
+                                errors.push(LoweringError::Other(format!(
+                                    "Error parsing `{}`: {e}",
+                                    attr.meta.to_token_stream()
+                                )));
+                                continue;
+                            }
+
+                            match (&location, &source) {
+                                (Some(l), Some(s)) => {
+                                    this.binding_includes.insert(l.clone(), s.clone());
```
Manishearth: nit: should error on duplicate includes, OR should store as a `Vec<CustomCodeInclude>` and allow duplicates

tool/src/config.rs
```diff
@@ -17,6 +21,8 @@ pub struct SharedConfig {
     /// Whether or not callbacks support references in parameters. This is unsafe: you need to be careful to not
     /// retain these references on the foreign side.
     pub unsafe_references_in_callbacks: Option<bool>,
+    /// The folder to pull custom bindings from. Defaults to the lib.rs folder.
+    pub custom_binding_location: PathBuf,
```
Manishearth: name here too

tool/src/cpp/mod.rs
```diff
@@ -111,10 +114,19 @@ pub(crate) fn run<'tcx>(
         };
         context.impl_header.decl_include = Some(decl_header_path.clone());
 
+        let block_source = if let Some(s) = ty_attrs
```
Manishearth: issue: this code is the responsibility of the individual `gen` calls.

We should write a separate function called `gen_custom_extra_code`, and call it in each of the four functions.

tool/src/cpp/mod.rs
```diff
@@ -132,6 +144,16 @@ pub(crate) fn run<'tcx>(
         context.impl_header.includes.remove(&*impl_header_path);
         context.impl_header.includes.remove(&*decl_header_path);
 
+        // Decl headers require some more special logic, but we can write to the impl header body directly:
+        if let Some(s) = ty_attrs
```
Manishearth: issue: same for impl stuff

tool/src/nanobind/mod.rs
```diff
@@ -174,6 +180,14 @@ pub(crate) fn run<'cx>(
         }
         drop(guard);
 
+        let binding_info = &ty.attrs().binding_includes;
+
+        if let Some(s) = binding_info.get(&hir::IncludeLocation::InitializationBlock) {
```
Manishearth: issue: same here.

tool/src/cpp/mod.rs
Manishearth: My reasoning is that this should be a part of the template, and should be possible to move around per-type.

Also, with the current code, this only works for opaques, which is an unnecessary restriction.

tool/templates/cpp/opaque_decl.h.jinja
```diff
@@ -34,6 +34,9 @@ private:
     {{type_name_unnamespaced}} operator=({{const_ref}}) = delete;
     {{type_name_unnamespaced}} operator=({{move_ref}}) noexcept = delete;
     static void operator delete[](void*, size_t) = delete;
+{% if !def_block.is_empty() %}
+{{ def_block }}
```
Manishearth: issue: I'd call the variable `extra_def_code` to make it clear that it's extra

core/src/hir/attrs.rs
```diff
@@ -419,70 +487,15 @@ impl Attrs {
                             this.abi_compatible = true;
                         }
                         "include" => {
```
Manishearth: The attribute itself should also be `custom_extra_code` i think.

`include` is rather C++ specific IMO.

Manishearth: Thanks, this is much easier to review now

tyler-zeromatter:
Currently blocked by https://github.com/rust-diplomat/diplomat/pull/1018

Manishearth (approved):

Issue #1016, opened 2026-01-21
Respect `#[cfg(feature = ...)]` attributes on `gen` call


tyler-zeromatter:
It would be nice to be able to selectively disable types and have `gen` respect the feature gates:

```rs
#[cfg(feature = testing_bindings)]
#[diplomat::bridge]
mod ffi {
  pub struct FfiTest() {}
  impl FfiTest {
    fn this_is_a_test_function() {}
  }
}
// Or
#[cfg(feature = testing_bindings)]
mod somelib;
```

This allows me to selectively create modules just for testing parts of FFI bindings that I don't want to expose to customers.

I would also settle for something like
```rs
#[cfg(feature = testing_bindings)]
#[diplomat::attr(auto, feature="testing_bindings")]
#[diplomat::bridge]
mod ffi {
  // ...
}
```

Either way, a list of features would have to be passed to `gen` in some way, I'm thinking maybe through `SharedConfig`?

tyler-zeromatter:
Here's a potential design:

# Enabling/Disabling bindings by feature

## The Problem We're Trying to Solve

When we have something like
```rs
#[cfg(feature = some_feature)]
mod some_feature
```

Or
```rs
#[cfg(feature=some_feature)]
mod ffi {}
```

Diplomat will automatically search and add bindings, even if the feature is disabled by default.
This is a trap that new users can easily fall into. Diplomat will generate the bindings, but Rust does not generate the code.

At compile-time, this leads to `unresolved symbol` errors that might be a little confusing to first-time users. We want to preserve default behavior, but we should still warn users if they are stumbling into a potential trap:

### Initial Safeguards
When validating with the hir, if Diplomat ever detects a feature gate, it should output a warning:
```
WARNING: Detected #[cfg(feature=some_feature)]. Diplomat automatically generates bindings for all features by default. To disable warnings, update config disable_feature_gate_warnings to `true`. See LINK for more.
```

## SharedConfig

Stores:
- `features_enabled`, an `Option<HashSet>`. If set to `None`, Diplomat will generate all features by default. If set to `Some`, any features detected in the `HashSet` will be passed to the `AttributeValidator` for validation that any `cfg` attributes have an associated Diplomat attribute with them, defined below. We error otherwise.
- `disable_feature_gate_warnings` disables warnings related to `cfg(feature=)`. This is both for the above warning, and for the safeguards for `diplomat::attr` detailed below.

### Safeguards
[...]

PR #1019, opened 2026-01-23
Add Diplomat Feature gates

tyler-zeromatter:
Closes https://github.com/rust-diplomat/diplomat/issues/1016

Very rough draft of the plan described above.

This currently does not implement warning/validation since that requires modifying the structure of the HIR attribute validator some (it needs access to `cfg(feature=)` from the AST), and I wanted to wait on further comments before going head with that.

### DIFF
--- a/tool/src/config.rs
+++ b/tool/src/config.rs
@@ -80,7 +80,7 @@ impl SharedConfig {
                             let slice = &st[1..st.len() - 1];
                             let hash = slice
                                 .split(",")
-                                .map(|s| s.replace("\"", "").replace(" ", "").to_string())
+                                .map(|s| s.replace("\"", "").trim().to_string())
                                 .collect();
                             hash
                         } else {


tool/src/config.rs
```diff
@@ -23,6 +23,10 @@ pub struct SharedConfig {
     pub unsafe_references_in_callbacks: Option<bool>,
     /// The folder to pull custom bindings from. Defaults to the lib.rs folder.
     pub custom_extra_code_location: PathBuf,
+    /// List of features to enable/disable generation for. If `None`, all features are enabled. 
+    pub features_enabled : Option<HashSet<String>>,
+    /// Whether or not to warn if #[cfg(feature=...)] is detected. If `None` or `Some(false)`, warnings are not generated.
+    pub disable_feature_gate_warnings : Option<bool>,
```
Manishearth: This should not be a part of the MVP, I'd like to discuss these warnings separately.

Also, depending on the design (cfg based or attr based) these warnings can be adequately dismissed with `features_enabled = all/true`

tool/src/config.rs
```diff
@@ -56,6 +60,37 @@ impl SharedConfig {
...
+                        Some(str_arr)
+                    }
+                    Value::Boolean(b) => {
+                        if b {
+                            None
+                        } else {
+                            Some(HashSet::new())
+                        }
+                    }
+                    Value::String(st) => {
```
Manishearth: IMO we should not be parsing a string as a single-element array here. The strings should be reserved for keywords like `all` etc.

feature_tests/src/attrs.rs
```diff
@@ -473,4 +473,7 @@ pub mod ffi {
         )
     )]
     pub struct BlockOverride();
+
+    #[diplomat::attr(not(feature=some_feature), disable)]
```
Manishearth, line 478: So just to be clear, it looks like you're oging with the design where Diplomat has a parallel universe of features that have nothing to do with Cargo features, yes? Users can choose to make them the same as Cargo features if they want, but they don't have to, and Diplomat does not read `cfg`s at all.

That's fine by me, I slightly prefer Diplomat not having to think about Cargo stuff.

tool/src/config.rs
Manishearth: Actually, see my comment o nthe design issue, maybe we don't want keywords supported.

feature_tests/src/attrs.rs
tyler-zeromatter: Yes, I think that's easiest. Not only was this like 5 seconds to implement, but I also think it allows for more versatility for Diplomat users for whatever they want to selectively enable/disable

Manishearth: Yeah, good to hear! Let's stick with that.

tool/src/config.rs
Manishearth: Yeah, let's just not support keywords. This can stay as is

core/src/hir/attrs.rs
```diff
@@ -1327,7 +1323,7 @@ pub struct BasicAttributeValidator {
     #[allow(clippy::type_complexity)] // dyn fn is not that complex
     pub is_name_value: Option<Box<dyn Fn(&str, &str) -> bool>>,
     /// The features supported
-    pub features_supported : Option<HashSet<String>>,
+    pub features_supported: Option<HashSet<String>>,
```
Manishearth: nit: `features_enabled` I think

And mention that None = all features are enabled.

core/src/hir/attrs.rs
```diff
@@ -1322,6 +1322,8 @@ pub struct BasicAttributeValidator {
     /// override is_name_value()
     #[allow(clippy::type_complexity)] // dyn fn is not that complex
     pub is_name_value: Option<Box<dyn Fn(&str, &str) -> bool>>,
+    /// The features enabled. If `None`, all features are enabled.
```
Manishearth: Actually! I think given that we're not tied to Cargo features we should just not have an all mode for now. `features_enabled` can be a HashSet, only taking list values, and we don't need `all_features_enabled`.

The "all features" thing is mostly useful because in a lot of use cases there will be features you _don't_ care about that you want to always have enabled, only a couple will be ones you actually care to toggle.

If people want it we can add support later, just a HashSet is fine for now.

PR #1025, opened 2026-01-28
Optional Args Support

### LINKED ISSUE #1023: Default arguments for functions
Most languages support being able to name optional arguments with default values after the required arguments. It would be nice to have notation for this in Diplomat:
```rs
#[diplomat::bridge]
mod ffi {
  struct Test {
    
  }

  impl Test {
    #[diplomat::attr(auto, default_args(one, two, three=0))]
    fn test_fn(one: i32, two : i64, three : i128) {
      
    }
  }
}
```
So in the backends that support it, one and two would be user-filled. Three would have a default value of `0` (specified by the language backend). To avoid weirder edge cases like `Option` or `Result`, we should limit these to primitive types for now.

tyler-zeromatter:
Closes #1023.

Wanted to check that this implementation is a good idea. Currently `default_args()` is stored as a map from parameter names to strings, where the String is the default value. This value is then parsed into a `DefaultArgValue` enum and stored into the relevant `Param` as `default_value: Option<DefaultArgValue>` during validation (and modified).

TODO:
- [x] Expand tests
- [x] Add sample C++ implementation
- [x] Approval on implementation details
  - [x] ~~Re-work default_args to be an array~~ (actually loses some expressiveness if this is the case)
- [X] Fix up `todo!`

### DIFF
--- a/core/src/hir/attrs.rs
+++ b/core/src/hir/attrs.rs
@@ -589,7 +589,12 @@ impl Attrs {
                         }
                         "default_value" => {
                             if !support.default_args {
-                                maybe_error_unsupported(auto_found, "default_value", backend, errors);
+                                maybe_error_unsupported(
+                                    auto_found,
+                                    "default_value",
+                                    backend,
+                                    errors,
+                                );
                                 continue;
                             }
                             // Note that we do not validate that the args match here.
--- a/feature_tests/src/attrs.rs
+++ b/feature_tests/src/attrs.rs
@@ -251,18 +251,14 @@ pub mod ffi {
     }
 
     impl OpaqueArithmetic {
-        pub fn make(
-            x: i32,
-
-            #[diplomat::attr(auto, default_value=12)] y: i32,
-        ) -> Box<Self> {
+        pub fn make(x: i32, #[diplomat::attr(auto, default_value = 12)] y: i32) -> Box<Self> {
             Box::new(Self { x, y })
         }
 
         #[diplomat::attr(supports=method_overloading, rename="make")]
         pub fn make_overload(
             x: f32,
-            #[diplomat::attr(auto, default_value=14.48)] y: f32,
+            #[diplomat::attr(auto, default_value = 14.48)] y: f32,
         ) -> Box<Self> {
             Box::new(Self {
                 x: (x as i32) + 2,
@@ -428,9 +424,7 @@ pub mod ffi {
 
     #[diplomat::attr(not(supports = free_functions), disable)]
     #[diplomat::attr(auto, namespace = "nested::ns")]
-    pub fn nested_ns_fn(
-        #[diplomat::attr(auto, default_value=true)] x: bool,
-    ) -> bool {
+    pub fn nested_ns_fn(#[diplomat::attr(auto, default_value = true)] x: bool) -> bool {
         !x
     }
 
--- a/feature_tests/src/slices.rs
+++ b/feature_tests/src/slices.rs
@@ -8,9 +8,7 @@ pub mod ffi {
 
     impl MyString {
         #[diplomat::attr(auto, constructor)]
-        pub fn new(
-            #[diplomat::attr(auto, default_value='T')] v: &DiplomatStr,
-        ) -> Box<MyString> {
+        pub fn new(#[diplomat::attr(auto, default_value = 'T')] v: &DiplomatStr) -> Box<MyString> {
             Box::new(Self(String::from_utf8(v.to_owned()).unwrap()))
         }
 


core/src/hir/attrs.rs
```diff
@@ -334,6 +339,125 @@ pub struct SpecialMethodPresence {
...
+    F64(f64),
+}
+
+#[derive(Clone, Debug)]
+pub enum DefaultArgValue {
+    Bool(bool),
+    Char(u8),
+    Byte(u8),
+    Ordering(std::cmp::Ordering),
+    Int(DefaultArgIntType),
```
Manishearth: issue: we don't know the type of integer constants.

We should instead have

```rust
enum DefaultArgValue {
  Bool(bool),
  Char(char),
  Integer(i128),
  Float(f64)
}
```

No need for this complexity.

core/src/hir/attrs.rs
```diff
@@ -520,9 +644,29 @@ impl Attrs {
...
+                        "default_args" => {
+                            if !support.default_args {
+                                maybe_error_unsupported(auto_found, "default_args", backend, errors);
+                                continue;
+                            }
+                            // Note that we do not validate that the args match here.
+                            
+                            let list = attr.meta.require_list();
+                            let res = list.map_err(|e| {
+                                LoweringError::Other(format!("Expected MetaList: {e}"))
```
Manishearth: issue: people don't know what MetaList is.

`DefaultArgValueString` should parse from `attr.meta`, and produce a "default arg attribute must be a list of arguments" error message if it doesn't fit

core/src/hir/attrs.rs
```diff
@@ -334,6 +339,125 @@ pub struct SpecialMethodPresence {
...
+    Byte(u8),
+    Ordering(std::cmp::Ordering),
+    Int(DefaultArgIntType),
+    IntSize(DefaultArgIntSizeType),
+    Int128(DefaultArgInt128Type),
+    Float(DefaultArgFloatType),
+}
+
+#[derive(Default, Clone, Debug)]
+pub struct DefaultArgValueString {
```
Manishearth: issue: we should just parse directly into DefaultArgValue. We don't need this intermediate, I think, if we type this by literal types instead of actual types.

core/src/hir/attrs.rs
```diff
@@ -936,6 +1081,94 @@ impl Attrs {
                         }
                     }
                 }
+                
+                if validator.attrs_supported().default_args && !default_args.is_empty() {
```
Manishearth: all of this becomes unnecessary if we just parse it to a value the first time. The exact type is not needed.

core/src/hir/methods.rs
```diff
@@ -123,12 +125,13 @@ pub struct TraitParamSelf {
...
 
 /// A parameter in a method.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 #[non_exhaustive]
 pub struct Param {
     pub name: IdentBuf,
     pub ty: Type<InputOnly>,
     pub attrs: Attrs,
+    pub default_value: Option<DefaultArgValue>,
```
Manishearth: Hmmmmmmm. Generally I prefer for attribute-universe things to stay within the `Attrs` struct.

It makes sense for the HIR to partially validate this, but I kind of would prefer to keep the HIR restricted to the things that everyone needs.

tyler-zeromatter: Sounds good, I'll see if there's a way we can validate without storing `default_value` in the HIR

tyler-zeromatter: Actually, I think that's doable no problem, but I will make sure to move this into `Attrs`

tyler-zeromatter: Oh, I already scrapped this lol, just forgot to yank it.

Manishearth:
> This PR also requires an upgrade to askama, but I can separate that out into its own PR if that's preferred


Yes please.

feature_tests/src/attrs.rs
```diff
@@ -251,12 +245,19 @@ pub mod ffi {
     }
 
     impl OpaqueArithmetic {
-        pub fn make(x: i32, y: i32) -> Box<Self> {
+        pub fn make(
+            x: i32,
+
+            #[diplomat::attr(supports=default_args, default_arg=12)] y: i32,
```
Manishearth: use `auto` imo.

core/src/hir/attrs.rs
```diff
@@ -66,6 +66,9 @@ pub struct Attrs {
 
     /// Information on if a type declaration/impl block has custom bindings, and if so, what kind.
     pub custom_extra_code: HashMap<IncludeLocation, IncludeSource>,
+
+    /// A default argument above a parameter.
+    pub default_arg: Option<DefaultArgValue>,
```
Manishearth: nit: `default_value`, both in this code and in the attribute name

PR #1026, opened 2026-01-29
Memory Corruption for Owned Slice as Parameter (`Box<[u8]>` drop leads to segfault)

tyler-zeromatter:
This is so far just a failing test (at least on Windows) for when Nanobind is passed a string that is meant to be owned.

I think the solution here is to clone the slice right after it's been grabbed from Python, but my tests there so far have failed. Will continue to tinker with this.

### DIFF
--- a/feature_tests/cpp/tests/strings.cpp
+++ b/feature_tests/cpp/tests/strings.cpp
@@ -1,6 +1,7 @@
 #include <iostream>
 #include "../include/Opaque.hpp"
 #include "../include/OptionString.hpp"
+#include "../include/MyString.hpp"
 #include "assert.hpp"
 
 using namespace somelib;
@@ -20,4 +21,9 @@ int main(int argc, char* argv[]) {
     output = "prefix ";
     os->write_write(output).ok().value();
     simple_assert_eq("string write with result", output, "prefix hello world");
+
+    std::unique_ptr<MyString> s = MyString::new_owned("ABC");
+    simple_assert_eq("Owned string pass in", s->get_str(), "ABC");
+    s->set_str("123");
+    simple_assert_eq("Owned string pass out", s->get_str(), "123");
 }


tyler-zeromatter:
Seems like the main issue is when `Box<DiplomatStr>` is dropped. If I change `new_owned(v: Box<DiplomatStr>)` to `new_owned(v : &DiplomatStr)`, it works fine.

tyler-zeromatter:
The minimum reproducible test case in C (note that setting the length to `0` leads to no segfault):

```c
MyString_new_owned({
    "ABC", 3
});
```

Manishearth:
> It could be that Nanobind ends up doing something weird with how it interprets taking a Python string and converting it into a `std::string_slice`.

Oh. Uh, if it's converting to a `std::string_slice`, then you shouldn't be using owned slices, right? For owned slices you want it to convert to a rust-backed owned slice. `std::string_slice` is a C++-backed borrowed slice, so you're not getting any benefit of owned slices.

Feels like they should just be disabled in the nanobind backend?

tyler-zeromatter:
See my further notes, this isn't exclusive to Nanobind. C and C++ have memory leaks as well (see the above minimum reproducible test case).

Thinking on this, there must be some wonky conversion happening from `diplomat_runtime::DiplomatOwnedStrSlice -> Box<[u8]>` (even though stepping through in LLDB, the pointer converts fine, and we are able to view the data it  holds just fine?). We get a segfault when we drop `Box<[u8]>`.

For further testing, I'll remove Diplomat as the middleman entirely and investigate just through raw Rust to C ABI calls.

Manishearth:
> C and C++ have memory leaks as well (see the above minimum reproducible test case).

Well, yes, because an owned slice makes no sense for Diplomat-C++. I believe it was only added because nanobind wanted it.

PR #1033, opened 2026-01-30
Use correct edges and append array setup for Kotlin

Manishearth:
More progress on https://github.com/rust-diplomat/diplomat/issues/1003

This does not yet hook slices up to append arrays, but that change ought to be easier once this is landed.

The previous setup was incorrect: this correctly persists edges from structs.

I'm doign something slightly different from Dart. Dart _only_ uses append arrays for slice fields, and uses `struct.fieldsForLifetimeFoo` getters to get struct lifetime fields.

I think this isn't actually a better design: append arrays are consistent/simpler and can be used for all lifetimes. The resultant code here is cleaner since we're not trying to do two things at once.

### DIFF
--- a/example/kotlin/somelib/src/main/kotlin/dev/diplomattest/somelib/FixedDecimalFormatterOptions.kt
+++ b/example/kotlin/somelib/src/main/kotlin/dev/diplomattest/somelib/FixedDecimalFormatterOptions.kt
@@ -82,7 +82,6 @@ class FixedDecimalFormatterOptions (var groupingStrategy: FixedDecimalGroupingSt
         fun default_(): FixedDecimalFormatterOptions {
             
             val returnVal = lib.icu4x_FixedDecimalFormatterOptions_default_mv1();
-            
             val returnStruct = FixedDecimalFormatterOptions.fromNative(returnVal)
             return returnStruct
         }
--- a/example/kotlin/somelib/src/main/kotlin/dev/diplomattest/somelib/Lib.kt
+++ b/example/kotlin/somelib/src/main/kotlin/dev/diplomattest/somelib/Lib.kt
@@ -70,12 +70,32 @@ internal class GCSlice(val memory: Memory?, val slice: Slice) {
     fun close() {
         memory?.close()
     }
+
+    // Stick the contained memory into a list of edges. Returns `this` for convenient chaining.
+    fun into(edges: List<MutableList<Any>>): GCSlice {
+        memory?.let {
+            for(edge in edges) {
+                edge.add(memory)
+            }
+        }
+        return this
+    }
 }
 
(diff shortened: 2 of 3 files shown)


tool/src/kotlin/mod.rs
```diff
@@ -315,14 +337,83 @@ impl<'cx> ItemGenContext<'_, 'cx> {
...
+                        write!(&mut params, ")").unwrap();
+                    }
+                } else {
+                    if let Some(ref mut needs_temporary) = needs_temporary {
+                        **needs_temporary = true;
+                    } else {
+                        panic!("Struct borrow info MUST reference all lifetimes");
+                    }
+                    write!(&mut params, "arrayOf(temporaryEdgeArena)").unwrap();
+                }
```
jcrist1, line 396: I found this bit quite confusing. I think it's stemming from the optimisation. I did decide to benchmark it with a random initialisation and the optimisation is around a 3x speedup. I still feel like there might be a more declarative approach here. I don't know if this is any better really, (and I'm not sure it's even the same... it's getting late here)
```rust
            let env = || &struct_borrow_info.unwrap().use_env;

            let use_lts = struct_borrow_info
                .and_then(|i| i.param_info.borrowed_struct_lifetime_map.get(&def_lt));
            match (use_lts, needs_temporary.as_mut()) {
                (None, None) => panic!("struct borrow info must reference all lifetimes"),
                (None, Some(ref mut needs_temporary)) => {
                    ***needs_temporary = true;
                    write!(&mut params, "arrayOf(temporaryEdgeArena)").unwrap();
                }
                (Some(use_lts), None) if use_lts.len() == 1 => {
                    let lt = use_lts.first().map(|lt| env().fmt_lifetime(lt)).unwrap();
                    write!(&mut params, "{lt}AppendArray").unwrap();
                }
                (Some(use_lts), opt) => {
                    let env = env();
                    let mut maybe_comma = "";
                    write!(&mut params, "arrayOf(").unwrap();
                    for lt in use_lts.iter().map(|lt| env.fmt_lifetime(lt)) {
                        if opt.is_none() {
                            write!(&mut params, "{maybe_comma}*{lt}AppendArray").unwrap();
                        } else {
                            write!(&mut params, "{maybe_comma}{lt}Edges").unwrap();
                        }
                        maybe_comma = ",";
                    }
                    write!(&mut params, ")").unwrap();
                }
            }

```

jcrist1 (commented):
I had a quick look over the PR, and it looks good. I only have one comment about getting confused by generating the LT params for toNative.

tool/src/kotlin/mod.rs
Manishearth: Yeah I was trying to share the `maybe_comma` code which gets gnarly.

If you have ideas forcleaning this up that would be great. A single `match` statement seems like a good thing to aim for.

PR #1041, opened 2026-02-03
Nanobind overload args fix

tyler-zeromatter:
Fixes an issue where Nanobind would use the first method in its list of overloaded methods to populate information for all the other overloaded methods. This adds an associated fix with the askama upgrade, and a new test.


Walter-Reactor:
Looks like a solid fix! Nice work

Manishearth (approved):

PR #1043, opened 2026-02-04
Backend specific docs

### LINKED ISSUE #1042: Backend-specific comments
It would be really nice if backends would only forward specifically marked comments. That allows us to do things like:

```rs
/// SOME_CPP_EXAMPLE:
/// ```cpp
/// some_function_call(param_one, param_two, param_three);
/// ```
/// SOME_JS_EXAMPLE:
/// ```js
/// some_function_call(param_one, param_two, null);
/// ```
```

Or more specific comments tailored to the doc-gen system that you end up using.

tyler-zeromatter:
Closes #1042

TODO:
- [x] Add documentation
- [x] Make sure CI works

### DIFF
--- a/book/src/docs.md
+++ b/book/src/docs.md
@@ -78,10 +78,17 @@ If you want documentation that only appears in certain backends, you can use `#[
 #[diplomat::bridge]
 mod ffi {
     /// This comment will be shown everywhere.
+    /// This comment will also be shown everywhere.
     #[diplomat::docs(cpp)]
-    /// This comment will only be shown in C++
+    /// This comment will only be shown in C++.
+    /// ```cpp
+    /// void sampleCodeHere();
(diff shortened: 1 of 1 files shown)


book/src/docs.md
```diff
@@ -78,3 +70,21 @@ The full list of item kinds recognized by `rust_link` is:
  - `AssociatedTypeInTrait`
  - `AssociatedTypeInStruct`
  - `Typedef`
+
+## Backend Specific Documentation
+If you want documentation that only appears in certain backends, you can use `#[diplomat::docs(*)]` and [config language](./config.md) to specify which comments will appear in which backend:
```
Manishearth, line 75: nit: include multiple lines in the comments so it's clear that consecutive comments are grouped

Issue #1050, opened 2026-02-09
Nanobind crashes on calling Opaque ZST constructor twice


tyler-zeromatter:
Failing test in
https://github.com/rust-diplomat/diplomat/pull/1049

tyler-zeromatter:
Stepping through Nanobind, my thought is that the issue is Nanobind or Python somehow caching the returned pointer of the first constructor call (when it should be a unique pointer each time), because it somehow understand that the underlying Rust type is a ZST (because weirdly enough, this works fine when calling Opaque constructors that aren't ZST)

Manishearth:
So this should be fixed by the nanobind backend erroring when it sees this type.

tyler-zeromatter:
For that errorring behavior, we'd have to add info to the HIR to mark when an opaque return type is a ZST. I think in that case, Nanobind would still be able to have support for this behavior (instead of returning unique_ptr, we could just return a const pointer when we detect a ZST opaque)

tyler-zeromatter:
Oh, and of course this will have to account for both ZST *errors* as well as *success* types. This will not look amazing, but I'm confident it will work

PR #1051, opened 2026-02-09
Disable `Box<ZST Opaque>` in Nanobind

tyler-zeromatter:
Partial fix for #1050. Currently just throws out errors for the types and comments out the tests in Nanobind that no longer support this feature.

Will work on adding lambda functions to re-enable this behavior in a follow-up PR, but I thought this would be better to get out as it is now, to avoid crashes for anyone using Diplomat with Nanobind.

### DIFF
--- a/core/src/ast/snapshots/diplomat_core__ast__modules__tests__simple_mod.snap
+++ b/core/src/ast/snapshots/diplomat_core__ast__modules__tests__simple_mod.snap
@@ -144,6 +144,7 @@ declared_types:
       mutability: Immutable
       attrs: {}
       dtor_abi_name: OpaqueStruct_destroy
+      inner_zst: false
 declared_traits: {}
 declared_functions:
   other_test_function:
--- a/core/src/hir/snapshots/diplomat_core__hir__elision__tests__simple_mod.snap
+++ b/core/src/hir/snapshots/diplomat_core__hir__elision__tests__simple_mod.snap
@@ -529,6 +529,7 @@ TypeContext {
                 iterator: None,
                 iterable: None,
             },
+            inner_zst: false,
             dtor_abi_name: "Opaque_destroy",
         },
     ],


core/src/ast/opaque.rs
```diff
@@ -37,6 +39,7 @@ impl OpaqueType {
             methods: vec![],
             mutability,
             attrs,
+            inner_zst: strct.fields.is_empty(),
```
Manishearth, line 42: I just realized, for opaque types this is not sufficient. You won't be able to tell if something is a ZST ust by looking at the fields, since the field could be another ZST.

So we shouldn't make this a part of the AST.

tyler-zeromatter: How would you propose determining this? I think we still need to have info from the AST for the HIR to be able to lower.

I see an issue though, where someone can point to a struct outside of Diplomat that is itself a ZST, but we won't be able to tell because Diplomat will not be able to read that struct.

tyler-zeromatter: There's an easy runtime solution where we can have Rust just tell us if a struct is a ZST or not (which we would for sure want to only implement in Nanobind), but this seems much less than ideal

tyler-zeromatter: The only other thing I can think of right now is to rename this to `has_no_fields`, and we can then evaluate in the HIR to see if the inner type is one we know. Then `is_zst` can only return false negatives.

That doesn't solve the fact that Nanobind can still be passed ZST opaques unknowingly, however.

tyler-zeromatter: I think the solution will have to be in the macro, because getting this to work during Nanobind compile time would be complicated. I'm thinking something like, during compile time, we add a `const fn` to go through each Opaque and check its size in memory.

Then, if it's a ZST, we error and require a label of `diplomat::opaque(zst)` if it's not an opaque ZST in a way Diplomat can parse.

tyler-zeromatter: Actually, the other option is to override nanobind's caster for unique_ptrs to allow for the possibility of duplicates. If it is a duplicate, we'll just assume that the type is a ZST in memory.

I'll go ahead and try implementing this and seeing how it goes, but let me know if the above option works at all as well.

Manishearth: I really don't want to complicate the rest of Diplomat to support this nanobind quirk.

I think solving this in the backend is better, yeah.

tyler-zeromatter: I think the fastest solution here is to do some lambda conversions from unique_ptr to ptr* for all returned Nanobind opaque types.

There is something to be said for overriding the `stl/unique_ptr.h` implementation Nanobind gives us, but that would involve re-using some Nanobind code (currently licensed under BSD-3.0, which I don't think is a huge deal? But I don't want to be presumptuous), and could take a while. So just for speed, I think I will be focused on implementing the lambda function solution for this. It'll be messy, but it'll be a stopgap for now.

If we ever want to have actual support for returning a guaranteed unique_ptr into Python, we can revisit this.

tyler-zeromatter:
We need to go in a different direction than this PR, closing this and will hopefully open a new one later.

PR #1053, opened 2026-02-11
Nanobind ZST opaque templating

### LINKED ISSUE #1050: Nanobind crashes on calling Opaque ZST constructor twice

tyler-zeromatter:
Closes #1050. 

Took a while to find a solution that seemingly works correctly, but I feel confident about this approach. A lot of TODOs with this one:

- [x] Document/explain choices
- [X] ~~optional support~~ (Optional pointers are still pointers on the HIR side)
- [x] Lambda support
- [x] ~~Fix stringifier bug~~ (#1054)
- [x] Add askama generation for:
  - [x] Add
  - [X] ~~Add assign~~ (cannot return a value)
  - [x] Sub
  - [X] ~~Sub assign~~
  - [x] Mul
  - [X] ~~Mul assign~~
  - [x] Div
  - [X] ~~Div assign~~
  - [x] Static Getter
  - [x] Getter
  - [x] Getter + setter
  - [x] Iterator
  - [x] Iterable
  - [x] Indexer
  - [x] Fallible indexer
  - [x] Unsupported special methods
  - [x] Stringifier (as error result)
  - [X] ~~Comparison~~
- [x] Add tests (call the method twice) for:
  - [x] `result<T, E>`
    - [x] Where T is an opaque ZST
    - [x] Where E is an opaque ZST
    - [x] Where both T and E are opaque ZSTs  
  - [x] `optional<T>` 
  - [x] Constructor
  - [x] Static methods
  - [x] Mutable member methods
  - [x] Const member methods
  - [x] Add
  - [X] ~~Add assign~~
  - [x] Sub
  - [X] ~~Sub assign~~
  - [x] Mul
  - [X] ~~Mul assign~~
  - [x] Div
  - [X] ~~Div assign~~
  - [x] Getter
  - [x] Static getter
  - [x] Getter + Setter together
  - [x] Iterator
  - [x] Iterable
  - [x] Indexer
  - [x] Fallible indexer
  - [x] Stringifier
- [x] Simplify templating (not blocking)
  - [x] Give the template functions better names
- [x] Check for memory leaks
- [x] Rebase and clean-up PR

### DIFF
--- a/feature_tests/nanobind/src/include/diplomat_nanobind_common.hpp
+++ b/feature_tests/nanobind/src/include/diplomat_nanobind_common.hpp
@@ -329,16 +329,19 @@ inline std::unique_ptr<Return> maybe_alloc_zst(std::unique_ptr<Return> pointer) 
     }
 }
 
+
+// Helper for determining if we have a unique_ptr type.
 template <class T>
 struct get_unique_ptr : std::false_type {};
 
 template <class T>
 struct get_unique_ptr<std::unique_ptr<T>> : std::true_type{};
 
+// Given a Return type, map that inner type with maybe_alloc_zst (by unwrapping and re-wrapping the Return type).
 // Default implementation for plain unique_ptr types.
 template<typename Return>
 inline typename std::enable_if_t<get_unique_ptr<Return>::value, Return>
-wrap_func(Return to_wrap) {
+map_inner(Return to_wrap) {
     return maybe_alloc_zst(std::move(to_wrap));
 }
 
@@ -356,7 +359,7 @@ struct get_diplomat_result<somelib::diplomat::result<T, E>> : std::true_type {
 
 template<typename Return, typename... Args>
 inline typename std::enable_if_t<get_diplomat_result<Return>::value, Return>
-wrap_func(Return to_wrap) {
+map_inner(Return to_wrap) {
     using Result = get_diplomat_result<Return>;
     // Either the success type, the error type, or both are unique_ptrs.
     if (to_wrap.is_ok()) {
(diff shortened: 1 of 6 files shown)


tyler-zeromatter:
Thankfully, this solution compiles on the Linux CI runner, and it looks like we have failing tests already.

tyler-zeromatter:
@Manishearth mostly looking for your approval on [`maybe_alloc_zst`](https://github.com/rust-diplomat/diplomat/pull/1053/changes#diff-7a98aefab82eb1f225c3e5ceb86648a56667b3cfcb2af0dbf715db1340edc4abR319) to see if there are any potential issues there you can think of.

tool/templates/nanobind/function_base.cpp.jinja
```diff
@@ -1,22 +1,60 @@
...
+    {%- let success_is_owned_op = op.is_owned() -%}
+    {%- else -%}
+    {%- let success_is_owned_op = false -%}
+    {%- endif -%}
+    {%- if success_is_owned_op -%}
+    {{-pre_call-}}{{- caller() -}}{{- post_call -}}
+    {%- else -%}
+    {%- decl failure_is_owned_op -%}
+    {%- if let hir::ReturnType::Fallible(_s, Some(hir::Type::Opaque(op))) = m.method.output -%}
+    {%- let failure_is_owned_op = true -%}
```
walter-zeromatter: This should be able to be simplified to just be `let failure_is_owned_op = <if condition>`

tool/templates/nanobind/function_base.cpp.jinja
```diff
@@ -1,22 +1,60 @@
...
+{%- if overload_params.len() > 0 -%}
+nb::overload_cast<{% for p in overload_params -%} {{- p.type_name -}} {%- if !loop.last -%}, {% endif %} {%- endfor
+    -%}>(
+{%- endif -%}
+&{{- method_name_cpp_inner() -}}
+{%- if m.overloads.len() > 0 -%} ) {%- endif -%}
+{%- endmacro -%}
+{%- macro method_name_cpp(pre_call="std::move(maybe_op_unwrap(", post_call="))") -%}
+    {%- decl success_is_owned_op -%}
+    {%- if let hir::SuccessType::OutType(hir::Type::Opaque(op)) = m.method.output.success_type() -%}
```
walter-zeromatter, line 12: This kind of branch statement looks a little code-smelly to me, much like the one below, but given the complexity of the match statement that'd be required I'm not 100% sure how to simplify it off the top of my head

feature_tests/src/attrs.rs
```diff
@@ -505,4 +497,136 @@ pub mod ffi {
     // C++ will not generate this, since it has all features disabled by default (see lib.rs)
     #[diplomat::attr(not(feature=some_feature), disable)]
     pub struct FeatureTest();
+
+    #[diplomat::attr(not(nanobind), disable)]
+    #[diplomat::opaque]
+    pub struct OpaqueZST;
```
Manishearth, line 507: nit: document this with a link to the issue so it's clear why we have this

Manishearth (approved):
Not looking too closely at the nanobind code, seems fine.

Walter-Reactor (commented):
My only complaint would be that the structuring of the adjacent if-else blocks in the askama templates is a little hard to parse, but that really is just a nit. You really should try and address your cleanup todo note before this actually merges though

tool/templates/nanobind/function_base.cpp.jinja
tyler-zeromatter: Askama is *pretty* good at being able to evaluate rust in the template syntax, but not amazing. There might be some tricks to reduce footprint here, I'll see what I can do.

tool/templates/nanobind/function_base.cpp.jinja
tyler-zeromatter: Ah, we can't, I forgot to add a check for `failure_is_owned_op = op.is_owned()`, so this might have to stay as is. Will try to simplify

tool/templates/nanobind/function_base.cpp.jinja
tyler-zeromatter: I was able to simplify the later if statements, but not the if-else chains here

PR #1057, opened 2026-02-14
Nanobind: Use PyErr_SetObject for exceptions

### LINKED ISSUE #1054: Nanobind stringify `-> Result<(), Self>` representation can result in stack overflow
Not as big as #1050 but there's something of an insidious loop happening:

1. Python user calls stringify on an object
2. The object fails to stringify, and so Diplomat Nanobind generates an exception
3. Diplomat calls stringify on the returned object.
4. Jump to 2.

This repeats until we crash. (See https://github.com/rust-diplomat/diplomat/issues/1055 for a solution)

### LINKED ISSUE #1055: Re-work Nanobind Exception Handling
Python allows us to use an object as part of the exception value rather than a string: https://docs.python.org/3/c-api/exceptions.html#c.PyErr_SetObject

We should use this instead as it's more intuitive than reading a string exception. Would fix #1054 as well.

tyler-zeromatter:
Closes #1055, closes #1054. Allows tests to be more accurate too, which is nice.

### DIFF
--- a/feature_tests/nanobind/src/sub_modules/somelib/ResultOpaque_binding.cpp
+++ b/feature_tests/nanobind/src/sub_modules/somelib/ResultOpaque_binding.cpp
@@ -13,6 +13,7 @@ void add_ResultOpaque_binding(nb::module_ mod) {
     nb::class_<somelib::ResultOpaque> opaque(mod, "ResultOpaque", nb::type_slots(somelib_ResultOpaque_slots));
     opaque
         .def("assert_integer", &somelib::ResultOpaque::assert_integer, "i"_a)
+        .def("give_self", &somelib::ResultOpaque::give_self, nb::rv_policy::reference_internal)
         .def(nb::new_(std::move(maybe_op_unwrap(&somelib::ResultOpaque::new_))), "i"_a)
         .def_static("new_failing_bar", std::move(maybe_op_unwrap(&somelib::ResultOpaque::new_failing_bar)))
         .def_static("new_failing_foo", std::move(maybe_op_unwrap(&somelib::ResultOpaque::new_failing_foo)))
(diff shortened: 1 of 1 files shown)


tyler-zeromatter:
Even nicer is that the original error strings are preserved if you try and get the string representation, so this shouldn't be breaking with any existing Diplomat Nanobind bindings.

tyler-zeromatter:
Found a failing test as a result of this change, will investigate next week.

(It appears to be a conversion issue, the error message is: `Critical nanobind error: nanobind::detail::nb_type_put("somelib.somelib.ResultOpaque"): attempted to copy an instance that is not copy-constructible!`)

tool/templates/nanobind/common.h.jinja
```diff
@@ -111,23 +111,27 @@ namespace nanobind::detail
...
 
-            auto errorPyV = nb::cast(std::move(std::move(value).err().value()));
+            auto errorPyV = ErrCaster::from_cpp(forward_like_<V>(std::move(value).err().value()), p, cl);
             if (errorPyV.is_valid())
             {
-                PyErr_SetString(PyExc_Exception, nb::str(errorPyV).c_str());
+                PyErr_SetObject(PyExc_Exception, errorPyV.ptr());
+                // PyErr_SetObject takes ownership (https://github.com/python/cpython/blob/fa73fd473f00dd231f59e44798a3d00a46322658/Python/errors.c#L151)
+                // but Nanobind expects Python to take ownership directly. So we decref after PyErr_SetObject takes ownership, to remove Nanobind's reference:
+                Py_DECREF(errorPyV.ptr());
```
tyler-zeromatter, line 129: Looked a little more into this, here are the full details on how this works:
- The pointer to the structure has been seen before (most cases)
~~- `rvp_policy` is:~~
 ~~- `reference` - GC is flagged to never call, so I believe we can DECREF without consequences (should double check).~~
  ~~- `copy` or `move` - Nanobind calls Python constructors `PyObject_New` or `PyType_GenericAlloc`, both of which create new references that we need to DECREF to take full ownership of. This is the default for most cases.~~
  ~~- `take_ownership` - Nanobind calls `inst_new_ext` which either calls `PyObject_Init` (which INCREFs for us to then DECREF) or `PyType_GenericAlloc` which returns a new reference.~~
  ~~- `automatic` and `automatic_reference` (which are combinations of the above options).~~

~~Not sure about:~~
~~- `reference_internal`, lots of confusing keep alive logic here. Currently investigating.~~

tyler-zeromatter: UPDATE: All of the `rvp_policy`s above call `inst_new_ext` or `inst_new_int` before getting into policy specific logic. `inst_new_*` makes a new Python object (and accordingly, a ref count of 1). So this works regardless of the policy, since Nanobind will either INCREF an existing object or make a new one; then PyErr_SetObject takes direct ownership of the python object and we DECREF accordingly.

tyler-zeromatter:
Will merge once this passes checks

PR #1061, opened 2026-02-20
Nanobind Overloading Constructors

tyler-zeromatter:
Touches a lot of templating code somewhat by necessity. Overloads of `nb::new_` require that the zero param constructor is defined first.
TODO:
- [x] Rebase and cleanup commits


tyler-zeromatter:
Not sure if this closes https://github.com/rust-diplomat/diplomat/issues/234 entirely, but for any backends that support overloads and constructors (which is just Nanobind), then I think this works well.

PR #1068, opened 2026-03-05
Properly support slice borrowing in Kotlin

Manishearth:
Fixes https://github.com/rust-diplomat/diplomat/issues/1003

This follows a slightly different scheme from Dart/JS and I think it's cleaner overall: Slice memory handles can be shoved into append/edge arrays as needed using a utility type.

I think we've spent a bunch of effort in the other backends to try and avoid doing too much array manipulation, but I don't actually think it's worth it: most cases have very few lifetimes.


core/src/hir/methods/borrowing_param.rs
```diff
@@ -196,6 +196,12 @@ impl<'tcx> BorrowingParamVisitor<'tcx> {
         }
     }
 
+    /// Given a used method lifetime, return all borrowed lifetimes. None if the
+    /// lifetime isn't actually used in the method.
```
jcrist1, line 200: Is a used lifetime a lifetime parameter in the return type of the method?

Manishearth: Yes

PR #1075, opened 2026-03-06
Add C abi type documentation

tyler-zeromatter:
Expands the book to document how we use the C ABI in the C backend.

Thought this would be useful for anyone trying to understand the outputs of the C backend, or whoever's trying to build their own backend.


tyler-zeromatter:
It would also be nice to update the `authors` section of `book.toml` to be updated to `The Rust Diplomat Maintainers` or something, since I think a number of people have contributed to it at this point

PR #1089, opened 2026-03-26
Opaque Slice Support

tyler-zeromatter:
Closes #1088
A lot of TODOs with this one:

- Look closer at `&[DiplomatStrSlice]` for reference
- [x] Create rust tests and implement (create tests for C++, JS, and Nanobind)
  - [X] Opaque slice as method param 
  - [x] Mutable slices (should be disallowed for now)
  - [x] Slices of mut opaques (should be disallowed for now)
  - [x] Returned slices
  - [x] ~~Returned boxed slices~~ Don't want to worry about for now, Diplomat currently only worries about this for primitive slices.
  - [x] Static lifetimes
    - [x] Static slice of static lifetime
    - [x] Non-static slice of static lifetime 
  - [x] Test `&[Box<Opaque>]`  fails
  - [x] Slice of optional opaques
  - [x] Nothing can borrow from the slice (or the inner opaque) (no lifetime `'a` in the slice can be anywhere else)
  - [x] Input only slice of opaques
- [ ] ~~Clean up C generation~~
  - This may not be possible currently, and I think we're okay with that
  - I don't want to have to generate a `SliceView` struct for EVERY opaque. This might involve a slight refactor, but I will investigate and see.
- [x] Clean up first backend iterations
- [x] Add `supports=opaque_slices` (each language panics if they don't support it, but advertising without validation would be nice).
  - [x] Check during lowering (drop the Kotlin, demo_gen, etc. commits) 
- [x] Passing test (ready for review at this point)
- [ ] Documentation
  - [ ] C ABI
  - [ ] `slices.md`
- [x] Rebase and clean up


walter-zeromatter (approved):
Nanobind changes LGTM

tyler-zeromatter:
@Manishearth wanted to ping you on this again whenever you have a sec. I'm curious mostly if you're still comfortable with the current approach of generating a custom opaque slice type for each C opaque binding.

Happy to hash everything else out once we nail down the approach for C.

Manishearth:
I think generating a custom opaque slice type per binding is fine. I haven't really had time to look at this yet (it's in my queue, but I expect this to be more involved)

tyler-zeromatter:
After rebasing and merging, I'm not really sure why I created multiple implementations other than C++, I think it makes it messy to review. Happy to drop these commits and make them their own PRs.

tool/src/js/converter.rs
```diff
@@ -726,18 +739,11 @@ impl<'tcx> ItemGenContext<'_, 'tcx> {
...
-                            format!(
-                                r#"{spread_pre}{alloc_stmnt}diplomatRuntime.DiplomatBuf.slice(wasm, {js_name}, "{}"){alloc_end}{spread_post}"#,
-                                self.formatter.fmt_primitive_list_view(*p)
-                            )
-                        }
+                        hir::Slice::Primitive(_, p) => format!(
+                            r#"{spread_pre}{alloc_stmnt}diplomatRuntime.DiplomatBuf.slice(wasm, {js_name}, "{}"){alloc_end}{spread_post}"#,
+                            self.formatter.fmt_primitive_list_view(*p)
+                        ),
+                        hir::Slice::Opaque(_, op) => format!(r#"{spread_pre}{alloc_stmnt}diplomatRuntime.DiplomatBuf.slice(wasm, {js_name}.map((op) => {}), "u32"){alloc_end}{spread_post}"#, self.gen_js_to_c_for_type::<hir::Everywhere>(&Type::Opaque(op.clone()), "op".into(), struct_borrow_info, alloc, JsToCConversionContext::List)),
```
Manishearth, line 746: nit: comment here saying that we do not need to be fancy because lowering prohibits borrowing from opaque slices

you might want to file a followup issue and link to it here and in the lowering section

Manishearth (approved):
This looks great and is basically ready to merge with one nit

One thing I'm less happy about is the fact that every type in C gets the View type. Might be worth looking into alternate ways of doing this in a followup, e.g. using local anon struct types.

Manishearth:
I'm going to go ahead and merge this so the `.NET` PR can rebase.

Please do file the followup issues and add links once you're back!

PR #1100, opened 2026-04-08
fix panic for nanobind constructor overloading

markjm:
Hi - thanks for the project! I was looking at generating python bindings for https://github.com/unicode-org/icu4x/tree/9b075f68940cdbf98139f4573e63249619d9b464/ffi

This tool uses diplomat to create those bindings. After setup, I saw the following issue:

## The Panic

  The panic occurred in tool/src/nanobind/gen.rs line 367:
```
  _ => { panic!("Method Info for {} already exists but isn't a getter or setter!", e.method_name); }
```
```
  thread 'main' panicked at 'Method Info for from_ already exists but isn't a getter or setter!', tool/src/nanobind/gen.rs:367:20
  stack backtrace:
     0: rust_begin_unwind
               at /rustc/.../library/std/src/panicking.rs:...
     1: core::panicking::panic_fmt
               at /rustc/.../library/core/src/panicking.rs:...
     2: diplomat_tool::nanobind::gen::ItemGenContext::gen_opaque_def
               at ./tool/src/nanobind/gen.rs:367:20
     3: diplomat_tool::nanobind::generate
               at ./tool/src/nanobind/mod.rs:...
     4: diplomat_tool::main
               at ./tool/src/main.rs:...
  note: run with `RUST_BACKTRACE=full` for a verbose backtrace
```

When running

```
  /home/mmolinaro/code/diplomat/target/release/diplomat-tool \
      ffi/capi/src/lib.rs \
      nanobind \
      ffi/python/src \
      ffi/capi/config.toml
```

  On code like ICU4X's Decimal:
  ```
  #[diplomat::attr(supports = method_overloading, rename = "from")] // First "from"
  #[diplomat::attr(auto, named_constructor)]
  pub fn from_int32(v: i32) -> Box<Decimal> { ... }

  #[diplomat::attr(supports = method_overloading, rename = "from")]  // <- Second "from"
  #[diplomat::attr(auto, named_constructor)]
  pub fn from_int64(v: i64) -> Box<Decimal> { ... }  // <- PANIC HERE
  ```

On inspection, it seems like this is what happens:

  1. from_int32 renamed to "from" → Python keyword escape → "from_"
  2. Added to map with key "from_"
  3. from_int64 also renamed to "from" → "from_"
  4. Tries to insert with same key "from_"
  5. Hits .and_modify() - entry already exists
  6. match `method.attrs.special_method` sees `Some(SpecialMethod::NamedConstructor(_))`
  7. Falls through all cases to the _ catch-all
  8. panic

## Fix

The fix here is to handle `NamedConstructor` in nanobind generation code and templates to produce correct code.

I understand this could probably also be addressed by changing the attr name to be something thats not a keyword (something like `#[diplomat::attr(python, rename = "somethingelse")]`), but since this case panicked and didnt fail gracefully i figured id put up a patch.

Also added some tests demonstrating the failure case.

---

As a note, because I didnt see any related info/contributing guide - this PR was AI-tooling supported, since im not super familiar with this space, but all communication was me-generated :)


tool/src/nanobind/mod.rs
```diff
@@ -571,4 +571,124 @@ mod test {
...
+    fn test_method_overloading_python_keyword() {
+        // Test that method overloading works when methods are renamed to Python keywords
+        //
+        // Problem: ICU4X has methods like from_int32(), from_int64(), from_uint32() that
+        // are all renamed to "from" for a nice Python API. Since "from" is a Python keyword,
+        // Diplomat escapes it to "from_". This caused a panic because the code only handled
+        // Constructor overloading, not NamedConstructor.
+        //
+        // Expected Python API after fix:
+        //   d = Decimal.from_(42)       # calls from_int32
```
tyler-zeromatter: @Manishearth interested to hear your thoughts on this. I generally prefer to keep tests for backend functionality (especially if it's to do with generation) in the `feature_tests` folder, so I'd add these functions and this type to something like `feature_tests/structs.rs` or something. This way we avoid all the boilerplate that a test like this requires.

tyler-zeromatter:
Looking this over now, there's a few things I'm trying to understand about the problem, but I'll try to have some feedback by EOD today

tool/templates/nanobind/function_base.cpp.jinja
```diff
@@ -55,8 +55,10 @@ nb::overload_cast<{% for p in overload_params -%} {{- p.type_name -}} {%- if !lo
     {%- endif -%}
 
     {%- if m.overloads.len() > 0 %}
-    {#- Constructors have special overload behavior. -#}
-    {%- if let Some(hir::SpecialMethod::Constructor) = m.method.attrs.special_method -%}{%- else -%}
+    {#- Constructors and named constructors have special overload behavior. -#}
+    {%- if let Some(hir::SpecialMethod::Constructor) = m.method.attrs.special_method -%}
+    {%- else if let Some(hir::SpecialMethod::NamedConstructor(_)) = m.method.attrs.special_method -%}
+    {%- else -%}
```
tyler-zeromatter: This feels a little complicated. There's code in `gen_all_method_infos` in `gen.rs` which uses a `match special_method` to determine the overloading behavior.

Since the Nanobind backend doesn't do anything special for named constructors (Python doesn't distinguish them between regular methods as far as I'm aware), I think the fix should be as simple as adding `NamedConstructor` to the `None | Some(hir::SpecialMethod::Constructor)` grouping, which seemed to work when I did some quick tests.

If there's issues with that, happy to hear and work out a solution that covers all bases.

tyler-zeromatter (changes requested):
I like the tests, I just think this could be restructured in a much simpler way. Happy to talk it out

markjm:
Sure let me check on your comments and get back to you this evening or tomorrow

tool/src/nanobind/mod.rs
Manishearth: So I generally agree, but for nitpicky backend-specific bugs I don't want to pollute feature_tests with a lot of extra stuff.

That said, multiple overloading is a general problem that should be tested more. So yes, this should probably be in feature-tests.


I still think the nanobind test boilerplate could probably be simplified. We could have a straightforward function that you give a bunch of tokens to and it `assert_snapshot`s the output, without needing to configure everything else.

markjm: > That said, multiple overloading is a general problem that should be tested more. So yes, this should probably be in feature-tests.

🫡 

thanks for pointing me at feature_tests, as it also showed me that running those removes a now-supported "unsupported" comment in other tests

tool/templates/nanobind/function_base.cpp.jinja
markjm: couldnt get that syntax to work within the jinja templating, but can make it work with `match`, so moved to that

tool/src/nanobind/gen.rs
```diff
@@ -335,12 +335,15 @@ impl<'ccx, 'tcx: 'ccx> ItemGenContext<'ccx, 'tcx> {
                                 e.def = info.def.clone(); // when a setter exists, use it's qualifiers instead.
                                 e.param_decls = info.param_decls.clone(); // also it's params, since the getter has none by definition.
                             }
-                            None | Some(hir::SpecialMethod::Constructor) => {
-                                if matches!(e.method.attrs.special_method, Some(hir::SpecialMethod::Constructor)) ^ matches!(method.attrs.special_method, Some(hir::SpecialMethod::Constructor)) {
+                            None | Some(hir::SpecialMethod::Constructor | hir::SpecialMethod::NamedConstructor(_)) => {
+                                let e_is_constructor = matches!(e.method.attrs.special_method, Some(hir::SpecialMethod::Constructor | hir::SpecialMethod::NamedConstructor(_)));
+                                let method_is_constructor = matches!(method.attrs.special_method, Some(hir::SpecialMethod::Constructor | hir::SpecialMethod::NamedConstructor(_)));
+
+                                if e_is_constructor ^ method_is_constructor {
```
tyler-zeromatter: I think this is closer, but I would argue that we don't even need to include these extra carve-outs if a method is a `NamedConstructor`. I would just add it to the `match` union and leave it at that.

The reason we have this special flag for `Constructor` functions in Nanobind is for overloaded *constructors*, we need to know not only the method name, but type information so Nanobind can overload these methods properly (https://nanobind.readthedocs.io/en/latest/classes.html#codecell58).

I would argue it's okay to avoid catching `NamedConstructor` and populating it with this additional information, because in Python a named constructor is just the same as any other static `def some_method() -> SomeType`. We don't need this special carve-out to check if it's a constructor; in Python, unless the signature is `def __init__() -> Self`, then we don't care about whether or not it's a constructor.

TL;DR I think you could remove most of these changes and keep only `| Some(hir::SpecialMethod::NamedConstructor(..))` as part of this statement. Ideally, I'd just like to avoid touching the templates in `templates/nanobind`, since I believe changing them is unnecessary for this PR.

tyler-zeromatter: Let me know if you disagree, I haven't messed around with this too much so I'm curious if you're experiencing different issues during implementation

markjm: hm, i see. sorry it didnt quite click what you were saying last time but i see what you are saying now. i think you are totally right though. ive made that one more change (and added an additional pytest to prove these changes are also actually working in python-land too.

Ps i addressed comments in different commits, but i can squash them all if thats how you prefer PRs land

tyler-zeromatter (approved):
LGTM!

tool/src/nanobind/gen.rs
tyler-zeromatter: No need to squash, that's done automatically by Github on merging

Manishearth (approved):
(implementation seems fine)

feature_tests/src/method_overloading.rs
```diff
@@ -0,0 +1,49 @@
...
+//! keywords (like "from") which get escaped to "from_". The code detected these as duplicate
+//! entries but failed to handle `NamedConstructor` special methods, only accounting for
+//! regular `Constructor` types.
+
+#[diplomat::bridge]
+#[diplomat::attr(auto, namespace = "mylib")]
+pub mod ffi {
+
+    #[diplomat::opaque]
+    pub struct Decimal;
```
Manishearth, line 13: issue: can you call this MethodOverloading?

And also can we just make this struct and all methods conditional on `method_overloading` (`diplomat::cfg(method_overloading)`), so that we don't generate code for this on platforms that don't support it?

tyler-zeromatter:
@Manishearth I'm inclined to merge unless you have anything else

Manishearth:
Sure, go ahead

PR #1106, opened 2026-04-15
Tuples Support

### LINKED ISSUE #1101: Support for Tuples
I know this is possible, I'm debating whether or not it should be a feature in Diplomat.

Theoretically, we could do something like:

```
#[diplomat::tuplable]
pub struct SomeStruct {
  x : i32,
  y : i32,
  z : SomeOpaque
}
```

OR

```rs
fn test(a : (i32, i32)) -> (i32, i32, SomeOpaque) // In lowering to HIR, these becomes structs with the "tuplable" attribute
```

And on generation in the given backend, the struct type is converted into that language's native tuple type. For Nanobind this would be pretty easy, as we could just use the Python tuple C API without much hassle.

I think there are enough languages where this is technically supported to justify it? Python has tuples. C++ has `std::tuple`. Dart supports tuples. JS could have fixed size arrays.

Wanting this mainly just for ease of use in Python, so it's not a huge priority that this be a feature; it's just a nice-to-have to be able to support writing tuples in Rust and to expect them to match conventions in the given language.

tyler-zeromatter:
Closes #1101. Currently this is just support for C++. Some TODOs before this is ready for review:

- [x] Block refs and passing in `self` for tuplable structs
- [x] Write HIR tests for lowering
    - [x] Test function for taking in ref on tuplable struct (on non-tuplable backend)
- [x] Rebase and clean up 
- [x] Passing tests (blocked by https://github.com/rust-diplomat/diplomat/pull/1118)
  - [x] https://github.com/rust-diplomat/diplomat/issues/1119

Other TODOs (waiting on approval for implementation before doing these):
- [x] Write docs
- [x] Create tuples tracking issue
  - [x]  "Sugar defs"
  - [x] References to tuples 
  - [x] Taking tuplable structs as `self`
  - [x] Slices of `abi_compatible` tuples
- [x] Add tracking issue to HIR errors


tyler-zeromatter:
The diffs look pretty big, but that's only because we're generating the tupled structs in non-tuplable backends as regular structs.

core/src/hir/type_context.rs
```diff
@@ -480,25 +496,50 @@ impl TypeContext {
...
+                _ => unreachable!(),
+            }
+
+            let can_tuple = match st_ty {
+                TypeDef::Struct(d) => d.attrs.tuplable,
+                TypeDef::OutStruct(d) => d.attrs.tuplable,
+                _ => unreachable!(),
+            };
+            if can_tuple {
+                errors.push(LoweringError::Other("Tuplable structs cannot be passed as `self, this is currently unsupported in Diplomat.".to_string()));
```
Manishearth: suggestion: why not say "tuplable structs cannot currently have methods"

tool/src/cpp/gen.rs
```diff
@@ -449,25 +464,31 @@ impl<'ccx, 'tcx: 'ccx> ItemGenContext<'ccx, 'tcx, '_> {
             fmt: &'a Cpp2Formatter<'a>,
             type_name: &'a str,
             ctype: &'a str,
+            fields: &'a [NamedType<'a>],
             cpp_to_c_fields: &'a [NamedExpression<'a>],
             c_to_cpp_fields: &'a [NamedExpression<'a>],
+            tuple_to_c_fields: &'a [NamedExpression<'a>],
```
Manishearth: issue: This seems like a duplication: the template should be able to handle this by using `cpp_to_c_fields`, yes?

Or we should use this in `cpp_to_c_fields`, conditioned on whether the type is tuplable.

I would rather not have multiple parallel cpp_to_c things

core/src/hir/snapshots/diplomat_core__hir__elision__tests__elision_in_struct.snap
```diff
@@ -60,6 +60,7 @@ Method {
                 generate_mocking_interface: false,
                 abi_compatible: false,
                 mut_struct_ref: false,
+                tuplable: false,
```
Manishearth: issue: thinking about this more I think the attribute should be called `tuple` and everything updated to say tuple instead of tuplable

tyler-zeromatter: I agree, this makes more sense

tool/src/cpp/gen.rs
tyler-zeromatter: I'm not sure why I did it this way, I'll try to re-merge into `cpp_to_c_fields` and if there are issues I will bring them up.

Manishearth: Yeah, either we should pregenerate the version of `cpp_to_c_fields` that we need, or we do something in the template.

tyler-zeromatter: Let me know your thoughts on this new implementation, I cut `AsFFI` and `FromFFI` for tuple structs in favor of conditionally switching to `std::get<idx>(tuple)` if the struct is tuplable inside of the `cpp_to_c` method.

I guess with `AsFFI` and `FromFFI` axed, I could rename these functions to `AsFFI` and `FromFFI`, but I'd rather keep the distinction so the functionality is clear (this doesn't return a C++ struct, but a standard template tuple).

Manishearth: Yes, I think keeping the distinction is good.

tyler-zeromatter: https://github.com/rust-diplomat/diplomat/pull/1106/changes#diff-c357a50648d4d0df5a20cd3adc409723fb13690d4aa560b1555598c1dc6c3cfbR989

feature_tests/src/structs.rs
```diff
@@ -611,6 +611,56 @@ pub mod ffi {
             self.i.get_debug_str(w);
         }
     }
+
+    #[diplomat::attr(auto, tuplable)]
+    pub struct TupleStruct<'a> {
```
Manishearth, line 617: let's skip generating these types for non tuple backends

tool/src/cpp/gen.rs
```diff
@@ -943,11 +980,16 @@ impl<'ccx, 'tcx: 'ccx> ItemGenContext<'ccx, 'tcx, '_> {
         &mut self,
         cpp_struct_access: &str,
         is_in_mutable_struct: bool,
+        tuple_idx: Option<usize>,
```
Manishearth, line 981: hmm. doesn't seem like we use `cpp_struct_access` anyway.

My initial thought was we could replace `cpp_struct_access` _and_ `tuple_idx` with a single `field_getter` argument, which is `var_name`, but I think that gets more complicated. Can we just remove this and use `var_name` for `field_getter`?

Manishearth (approved):
Yeah, this is better.

tool/src/cpp/gen.rs
tyler-zeromatter: I think that works well, so I can do that.

I also have a more radical proposal, which is we could make `NamedExpression::expression` a `struct ConversionExpression(Pre : Cow<str>, Post : Cow<str>)`. Then we don't need to add these extra hoops for conversion. The template could simply call a function like: `convert.format(std::get<{idx}>(tuple))`. It would also allow more flexibility for future implementations.

tyler-zeromatter: But maybe that would be a separate PR which we could rebase on top of

Manishearth: Yeah I'd *really* love conversion to be composable. But separate PR after this lands IMO.

Manishearth: Anyway I think the only action here should be to remove the `cpp_struct_access` I think my other suggestion will make the code worse.

tyler-zeromatter: Okay, filed separate issue: #1141.

Will remove `cpp_struct_access`

tyler-zeromatter:
Writing docs and filing tuples issue now.

Manishearth (approved):

tyler-zeromatter:
Ready to merge on approval

Manishearth:
The documentation should mention that these types are not allowed to have methods, right?

Manishearth:
Or at least the methods are going to be inaccessible

tyler-zeromatter:
Let me know how that looks

book/src/attrs/tuples.md
```diff
@@ -0,0 +1,26 @@
...
+}
+```
+
+In C++, `SomeStruct` will be returned as
+
+```cpp
+std::tuple<float, float>
+```
+
+Tuples currently do not support methods (at least, ones that take in `&self`), since past the boundary a tuple is not the struct it was generated from (although we can theoretically convert again on the boundary, see the tracking issue below).
```
Manishearth: wait, I'm confused: so this supports methods that take `self`? How does that work; those won't be callable in C++, right?

I feel like we should be disallowing methods entirely, except for static ones.

tyler-zeromatter: I should clarify, it can't take in either.

I guess it's an issue of semantics. My thought was that if you have the method:

```rs
pub fn some_method(&self)
```
For tuples, that would be equivalent to
```rs
pub fn some_method(s : &SomeTuple)
```

Thinking on it though, I guess this doesn't really match how languages would treat it. I will update the issue, docs, and HIR lowering errors to reflect this.

Manishearth: > I should clarify, it can't take in either.


Got it. "Methods with a receiver" probably.

> For tuples, that would be equivalent to


I was thinking about that as a possible path. I'd rather not support that until someone asks for it because as you say it doesn't seem idiomatic.

Manishearth (approved):

book/src/attrs/tuples.md
tyler-zeromatter: It's an issue of semantics anyways, you could just write
```rs
pub fn takes_self(some_tuple : &SomeLang)
```

And in the tuplable languages it remains the same. The only difference would be for languages without tuples:

```rs
#[diplomat::cfg(supports=tuples)]
pub fn takes_self(t : &SomeTuple)

#[diplomat::cfg(not(supports=tuples))]
pub fn takes_self(&self)
```
These are syntactically different in some languages that don't support tuples, and in some cases you *might* want to unify them.

Like you said though, best to wait until someone asks for it.

Manishearth: Yep!

tyler-zeromatter:
Merging on CI pass

Manishearth (approved):

Issue #1102, opened 2026-04-10
Pure Python backend


Manishearth:
Nanobind is nice, but it has a couple downsides:

 - You must additionally use C++ in your build. All other backends allow you to take an existing Rust binary and link to it
 - It doesn't have full borrowing support, since all of that is automatically handled by nanobind. I do not particularly _trust_ the automatic handling that nanobind has. At the very least, lifetime information is lost in Rust->C++ (nanobind supports [extra annotations](https://nanobind.readthedocs.io/en/latest/ownership.html#rvp), but they're still not fully fine grained), so C++ is likely to be more conservative.
 - We're basically piling FFI libraries on top of each other

ZeroMatter has found the nanobind backend useful (and I'm very much in favor of continuing to have this backend), but I'm still interested in a pure Python backend at some point, and people can weigh tradeoffs on their own. It's probably a bit more work since you'd have to hook into Python's internal FFI APIs.

It's also possible that you need the CPython hooks on the C++ side and can't do this from just pure Python bindings.

This is a big request, mostly filing an issue to track it.

cc @tyler-zeromatter

Manishearth:
I want to highlight one thing: it is quite possible that Nanobind is objectively better in every way. @walter-zeromatter and @tyler-zeromatter   have made some interesting points over the years (and I have unfortunately forgotten the details). I just don't yet know for sure, if someone were to assess the feasability of a pure Python backend we might learn more about the exact tradeoffs involved.

But for now I'm happy to just have this be an issue I can link to when this comes up.

tyler-zeromatter:
As far as I recall, Nanobind tends to be fairly performant and smart for quick iteration; whenever we need to add a feature Nanobind is pretty smart at doing conversions from the C++->Python types.

The real issue with the Nanobind backend is when there's a mismatch in intent between the Rust->C++->Python chain. That generally involves issues with lifetimes (which C++ doesn't understand as well as Rust).

As far as I'm aware, there are two approaches:

1. We generate bindings in pure Python, using [ctypes](https://docs.python.org/3/library/ctypes.html).
  - This is likely to be the least performant option
  - It avoids having to do an extra compilation step in C++
  - Zeromatter is unlikely to use this backend if this is the approach
2. We generate C bindings, and call into CPython.
  - Slower to iterate on, but gives us more control over how we bind types
    - This would be good for features down the road that involve complicated lifetime management (slices, opaques, etc.)
[...]

PR #1108, opened 2026-04-15
Nanobind error on struct of opaque ref

### LINKED ISSUE #1107: Nanobind does not support fields of opaque references
Tests are failing in https://github.com/rust-diplomat/diplomat/pull/1106 because the Nanobind backend seemingly does not support structs holding references to opaques (i.e., `const Opaque&`)

tyler-zeromatter:
Closes #1107. This is one issue that I don't think can be resolved unless we use the solutions offered in https://github.com/rust-diplomat/diplomat/issues/1102


tyler-zeromatter:
Now it seems like the test I wrote for other backends was also caught by C++ (https://github.com/rust-diplomat/diplomat/actions/runs/24477303871/job/71532747404?pr=1108#step:5:118).

It seems like the issue is that you cannot tell C++ that what a reference *points to* has been changed, additionally without changing the *contents* of either. Rust supports this, C++ seemingly does not (at least, with lvalue references).

Since this is somewhat beyond the scope of Nanobind now, I'm curious to get your thoughts @Manishearth. Should we disable fields of references to opaques in both C++ and Nanobind? Make a new type to reflect Rust's behavior?

tyler-zeromatter:
Dart also seemingly has a similar issue (although I don't believe it's the same as C++?) https://github.com/rust-diplomat/diplomat/actions/runs/24477866677/job/71534702545?pr=1108#step:6:97

Manishearth:
I'm not really clear on what the issue is here.

Fields with references to opaques is absolutely something we wish to support. ICU4X uses it in input structs, and it has use cases for output structs as well. There's no reason that should be broken.

Fields with references to _mutable_ opaques ... yeah that might be a bit fraught. Not sure if it needs to be a problem for C++, though. And not sure what's going on with Dart.

tyler-zeromatter:
I have some thoughts on what's going on with C++, but I think I need to do some more research before I come up with a more definitive answer

tyler-zeromatter:
Ah, I think I was mistaken (I was thinking about Nanobind's issues when I wrote the above about C++). It looks to be a references of structs issue (when we clone a struct at the boundary) rather than simply a struct holding references to opaques. I'm going to investigate this a little further and will propose some solutions.

tyler-zeromatter:
Moving discussion about C++ into #1110, researching what Dart's issue is now

tyler-zeromatter:
Separate issue for Dart: https://github.com/rust-diplomat/diplomat/issues/1111

tool/src/nanobind/gen.rs
```diff
@@ -241,7 +241,14 @@ impl<'ccx, 'tcx: 'ccx> ItemGenContext<'ccx, 'tcx> {
         let field_decls = def
             .fields
             .iter()
-            .map(|field| self.gen_ty_decl(&field.ty, field.name.as_str()))
+            .map(|field| {
+                if let Type::Opaque(op) = &field.ty {
+                    if !op.is_owned() {
+                        self.errors.push_error(format!("Borrowed opaques ({type_name}::{}) are unsupported as struct fields in Nanobind. See https://github.com/rust-diplomat/diplomat/issues/1107 for more.", field.name));
```
Manishearth, line 247: issue: this seems extreme. ICU4X has these types and people are looking into using ICU4X with nanobind. Your problem is the confluence of borrowed structs and structs containing borrowed opaques.

tyler-zeromatter: Alright, closing this and will discuss more reasonable solutions in the new issues.

PR #1113, opened 2026-04-17
Demo Gen Enums: Construct from String

### LINKED ISSUE #1112: JS: Handle enum string values in struct-from-fields
Based on the behavior I'm seeing in https://github.com/unicode-org/icu4x/issues/7881

I think the solution is to call the enum constructor from within demo_gen's default renderer, but I will play around with this some.

ambiguousname:
Closes #1112

### DIFF
--- a/example/demo_gen/demo/index.mjs
+++ b/example/demo_gen/demo/index.mjs
@@ -29,7 +29,8 @@ let termini = Object.assign({
                 name: "ICU4X Fixed Decimal Grouping Strategy",
                 type: "FixedDecimalGroupingStrategy",
                 typeUse: "enumerator",
-                values: ["Auto", "Never", "Always", "Min2"]
+                values: ["Auto", "Never", "Always", "Min2"],
+                base: somelib.FixedDecimalGroupingStrategy
             },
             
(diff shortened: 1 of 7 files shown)


ambiguousname:
The feature_tests version of demo_gen doesn't have a working webpage anymore for these kinds of tests, but I was able to test this with ICU4X's demo_gen output

ambiguousname:
Clippy fixed by https://github.com/rust-diplomat/diplomat/pull/1109

robertbastian:
In JS2 you're meant to be able to use the string form of an enum wherever the enum is required. This works for enum arguments, but doesn't seem to work for enums in structs. We should fix enums in structs instead of changing the demo to not use strings. This code should work:


```js
new icu.CaseMapper()
  .titlecaseSegmentWithOnlyCaseData(
    "...abcXYZ", icu.Locale.fromString("fr"),
    icu.TitlecaseOptions.fromFields({
      leadingAdjustment: 'None',
      trailingCase: 'Unchanged'
    }))
```

robertbastian (commented):
.

PR #1115, opened 2026-04-17
JS: Call enum constructor for struct fields

### LINKED ISSUE #1112: JS: Handle enum string values in struct-from-fields

robertbastian:
Fixes https://github.com/rust-diplomat/diplomat/issues/1112

### DIFF
--- a/feature_tests/js/api/MyStructContainingAnOption.mjs
+++ b/feature_tests/js/api/MyStructContainingAnOption.mjs
@@ -32,13 +32,13 @@ export class MyStructContainingAnOption {
         }
 
         if ("a" in structObj) {
-            this.#a = structObj.a;
+            this.#a = MyStruct._fromSuppliedValue(diplomatRuntime.internalConstructor, structObj.a);
         } else {
             this.#a = null;
         }
 
         if ("b" in structObj) {
-            this.#b = structObj.b;
+            this.#b = new DefaultEnum(structObj.b);
         } else {
             this.#b = null;
         }
--- a/tool/templates/js/struct.js.jinja
+++ b/tool/templates/js/struct.js.jinja
@@ -4,7 +4,7 @@
 {%- if typescript && !is_out && !fields.is_empty() -%}
 export type {{type_name}}_obj = {
     {%- for field in fields %}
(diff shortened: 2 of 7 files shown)


feature_tests/js/api/MyStructContainingAnOption.mjs
```diff
@@ -32,13 +32,13 @@ export class MyStructContainingAnOption {
         }
 
         if ("a" in structObj) {
-            this.#a = structObj.a;
+            this.#a = MyStruct._fromSuppliedValue(diplomatRuntime.internalConstructor, structObj.a);
```
robertbastian, line 35: this also now correctly calls struct constructors, which it didn't for optional struct fields before

tool/templates/js/struct.js.jinja
```diff
@@ -60,7 +60,11 @@ structObj {%- if typescript -%} : {{type_name}}_obj {%- endif -%} {%- if is_out
 
         {%- for field in fields %}
         if ("{{field.field_name}}" in structObj) {
-            this.#{{ field.field_name }} = {%- if let Type::Struct(s) = field.field_type %} {{field.js_type_name}}._fromSuppliedValue(diplomatRuntime.internalConstructor, structObj.{{field.field_name}}) {%- else %} structObj.{{field.field_name}} {%- endif %};
+            this.#{{ field.field_name }} = 
+                {%- if let Type::Struct(s) = field.field_type %} {{field.js_type_name}}._fromSuppliedValue(diplomatRuntime.internalConstructor, structObj.{{field.field_name}})
+                {%- else if let Type::Enum(e) = field.field_type %} new {{ field.js_type_name }}(structObj.{{field.field_name}})
```
Manishearth, line 65: nit: probably would like to have a `_fromSuppliedValue` here as well. I don't think we have one for enums, but we should: the ctor is user facing and could choose to have other fancy behaviors, and the ctor internally has to do a bunch of extra work to differentiate between internal and external construction.

We should have `_fromSuppliedValue` that does the actual "what have you given me" logic, and then a constructor that calls internal constructor logic or calls `_fromSuppliedValue`.

For followup.

Manishearth (approved):

PR #1120, opened 2026-04-22
Nanobind Always Generates Structs as Mutable Referenceable

### LINKED ISSUE #1119: Nanobind Structs should always be codegenned as `mut_struct_ref`
This wasn't caught by #1118 because I never checked the case where

```rs
pub struct StructOfOpaque<'a> {
  op : &'a Opaque
}
```

Is generated for Nanobind without the `mut_struct_ref` attribute applied. The way Nanobind handles structures is that in Python, they are merely a reference to the underlying C++ structure in memory, and can be mutated at any time. So we codgen:

```cpp
// StructOfOpaque.d.hpp
struct StructOfOpaque {
  const Opaque& op;
}
// StructOfOpaque_binding.cpp
void add_StructOfOpaque_binding(nb::module_ mod) {
  // ...
  .def_rw("op", &somelib::StructOfOpaque::op)
}
```

Which is disallowed, because we are not allowed to reseat opaque types.

I think the solution (as of right now), is to ensure Nanobind *always* generates structs as if they have the `mut_struct_ref` attribute (because we are accessing this struct as if it is always a mutable reference). If we start to introduce limitations to codegen of `mut_struct_ref`, then we can revisit this issue.

tyler-zeromatter:
Closes https://github.com/rust-diplomat/diplomat/issues/1119

### DIFF
--- a/tool/src/cpp/mod.rs
+++ b/tool/src/cpp/mod.rs
@@ -14,11 +14,18 @@ pub(crate) use formatter::Cpp2Formatter;
 pub(crate) use gen::ExtraCode;
 
 #[derive(Default, Debug, Clone, serde::Serialize, serde::Deserialize)]
-pub struct CppConfig {}
+pub struct CppConfig {
+    /// Should structures always generate as if they have the `mut_struct_ref` attribute applied?
+    /// Right now, this ensures that opaques will always generate as pointers instead of references.
+    pub structs_always_mut_ref : bool,
+}
 
 impl CppConfig {
-    pub fn set(&mut self, key: &str, _value: toml::Value) {
-        panic!("C++ does not support any backend-specific configs, found {key}");
+    pub fn set(&mut self, key: &str, value: toml::Value) {
+        match key {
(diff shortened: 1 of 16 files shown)


tool/src/cpp/mod.rs
```diff
@@ -14,11 +14,22 @@ pub(crate) use formatter::Cpp2Formatter;
 pub(crate) use gen::ExtraCode;
 
 #[derive(Default, Debug, Clone, serde::Serialize, serde::Deserialize)]
-pub struct CppConfig {}
+pub struct CppConfig {
+    /// Should structures always generate as if they have the `mut_struct_ref` attribute applied?
+    /// Right now, this ensures that opaques will always generate as pointers instead of references.
```
Manishearth, line 19: nit: mention nanobind

Manishearth (approved):
Not a fan but if we need it this is the best way to do it.

tyler-zeromatter:
Yeah, I don't love that this requires altering the codegen for C++, but I'm not sure of any other way to handle this. Probably another point in favor of #1102, this dependency on C++ is not good for the long term health of the backend.

Manishearth (approved):

PR #1122, opened 2026-04-22
Update Nanobind Struct Constructors to use `nb::new_`

### LINKED ISSUE #1121: Nanobind: Compiler error on fallible structure constructor with opaque
For a constructor for a structure that returns `Result<(), Box<Opaque>>`. The issue is something to do with
```cpp
.def("__init__",std::move(maybe_op_unwrap([](somelib::ns::RenamedStructWithAttrs* self, bool a, uint32_t b){ auto tmp = somelib::ns::RenamedStructWithAttrs::new_fallible(a, b);
                if(tmp.is_ok()) {
                    *self = std::move(tmp).ok().value();
                } else {
                    nb::cast(tmp); // This will raise a python error with the contents of the error type
                }})), "a"_a, "b"_a)
```
There's something going wrong with calling `nb::cast(tmp)`, which should just output the error correctly. Investigating now.

tyler-zeromatter:
Closes https://github.com/rust-diplomat/diplomat/issues/1121. This makes codgenned constructors consistent across types, and I have yet to encounter an error with this method.

Not sure why things were split up like this in the first place, so I'm checking in with @walter-zeromatter to make sure that everything is on the level.

Also adds a shared `set_py_err` helper function for potentially future implementations.


walter-zeromatter:
Looks solid to me. I think I did this originally when there were more limitations around declaring constructors in some contexts, but now that the behavior matches better, this is a solid consolidation.

walter-zeromatter (approved):

Manishearth (approved):
Quick skim seems fine, for more in-the-weeds nanobind stuff I'm unlikely to do in depth reviews

PR #1127, opened 2026-04-24
Nanobind Fix Bound Vector Check

### LINKED ISSUE #1124: Update Nanobind `diplomat::span` casting logic
If we create a type in Rust that mimics Python's vector type:

```rs
struct SomeVec(Vec<inner>);

impl SomeVec {
  pub fn __len__() -> usize;
  #[diplomat::attr(auto, indexer)]
  pub fn get(i : usize) -> inner;
  #[diplomat::attr(auto, iterable)]
  pub fn iter();
}
```

Then if we request a slice of the inner type:

```rs
pub fn some_fn(i : &[inner]);
```

```py
a = somelib.SomeVec(["""..."""])
# Nanobind tends to recognize this:
some_fn(a)
```

This leads to segfaulting. Need to do some more investigation, although this isn't currently a high priority.

tyler-zeromatter:
Closes #1124.

### DIFF
--- a/feature_tests/c/include/PrimitiveStructVec.h
+++ b/feature_tests/c/include/PrimitiveStructVec.h
@@ -27,9 +27,12 @@ DiplomatPrimitiveStructView PrimitiveStructVec_as_slice(const PrimitiveStructVec
 
 DiplomatPrimitiveStructViewMut PrimitiveStructVec_as_slice_mut(PrimitiveStructVec* self);
 
-PrimitiveStruct PrimitiveStructVec_get(const PrimitiveStructVec* self, size_t idx);
+typedef struct PrimitiveStructVec_get_result {union {PrimitiveStruct ok; }; bool is_ok;} PrimitiveStructVec_get_result;
+PrimitiveStructVec_get_result PrimitiveStructVec_get(const PrimitiveStructVec* self, size_t idx);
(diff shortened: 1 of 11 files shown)


tyler-zeromatter:
Seems like MSVC is working fine, need to investigate the issue with Ubuntu

tyler-zeromatter:
Ah, the compile defines are slightly different, this is a new bug:

```py
    a = somelib.PrimitiveStruct(0, True, ' ', 0, 0, 0)
    b = somelib.PrimitiveStruct(10., False, ' ', 0, 0, 0)
    c = somelib.PrimitiveStruct(20., False, ' ', 0, 0, 0)
    l = [a, b, c]
    assert l[0].x == 0
    assert l[1].x == 10
    assert l[2].x == 20

    from_slice = somelib.PrimitiveStructVec.take_in_slice(l)
    for i in from_slice:
         print(i)
```

This doesn't work (regardless of platform), because something about Python accessing the iteration API is broken.

Either this is a binding bug, or (more likely), this doesn't fully implement the spec Python expects.

tyler-zeromatter:
Blocked on https://github.com/rust-diplomat/diplomat/issues/1128

Manishearth (approved):

PR #1129, opened 2026-04-25
Fix C++ codegen using wrong view type for Slice::Strs with UTF-16 encoding

markjm:
The C++ code generator hardcodes `DiplomatStringView` for all `Slice::Strs` types regardless of encoding. For UTF-16 string slices (`&[DiplomatStr16Slice]`), this produced incorrect `reinterpret_cast` to `DiplomatStringView*` instead of `DiplomatString16View*`.

Fix by destructuring the encoding from `Slice::Strs(encoding)` and using `fmt_str_view_name(encoding)` to select the correct view type.

Adds a `new_from_utf16` test method to the `MyString` feature test to exercise this code path.

---

Found when generating codegen for icu4x

https://github.com/unicode-org/icu4x/blob/3b55a3ee7e2b879a83065ceb7d1540e44688b548/ffi/capi/src/list.rs#L173

```rust
        #[diplomat::rust_link(icu::list::ListFormatter::format, FnInStruct)]
        #[diplomat::rust_link(icu::list::ListFormatter::format_to_string, FnInStruct, hidden)]
        #[diplomat::rust_link(icu::list::FormattedList, Struct, hidden)]
        #[diplomat::attr(not(supports = utf8_strings), rename = "format")]
        #[diplomat::attr(supports = utf8_strings, rename = "format16")]
        pub fn format_utf16(&self, list: &[DiplomatStr16Slice], write: &mut DiplomatWrite) {
            let _infallible = self
                .0
                .format(
                    list.iter()
                        .map(|a| potential_utf::PotentialUtf16::from_slice(a))
                        .map(writeable::adapters::LossyWrap),
                )
                .write_to(write);
        }
```

### DIFF
--- a/tool/src/cpp/gen.rs
+++ b/tool/src/cpp/gen.rs
@@ -1021,19 +1021,22 @@ impl<'ccx, 'tcx: 'ccx> ItemGenContext<'ccx, 'tcx, '_> {
                 )
                 .into()
             }
-            Type::Slice(Slice::Struct(b, ref st)) => format!(
-                "{{reinterpret_cast<{}{}*>({cpp_name}.data()), {cpp_name}.size()}}",
-                if b.mutability().is_mutable() {
+            Type::Slice(Slice::Struct(b, ref st)) => {
+                let mutability = if b.mutability().is_mutable() {
                     ""
                 } else {
                     "const "
-                },
-                self.formatter.namespace_c_name(
+                };
+                let c_name = self.formatter.namespace_c_name(
                     st.id().into(),
-                    &self.formatter.fmt_type_name_unnamespaced(st.id())
+                    &self.formatter.fmt_type_name_unnamespaced(st.id()),
+                );
+                format!(
+                    "{{reinterpret_cast<{mutability}{c_name}*>({cpp_name}.data()), {cpp_name}.size()}}",
+
                 )
-            )
-            .into(),
+                .into()
+            }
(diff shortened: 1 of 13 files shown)


tool/src/cpp/gen.rs
```diff
@@ -990,37 +990,58 @@ impl<'ccx, 'tcx: 'ccx> ItemGenContext<'ccx, 'tcx, '_> {
...
-            Type::Slice(Slice::Struct(b, ref st)) => format!("{{reinterpret_cast<{}{}*>({cpp_name}.data()), {cpp_name}.size()}}",
-                if b.mutability().is_mutable() { "" } else { "const " },
-                self.formatter.namespace_c_name(st.id().into(), &self.formatter.fmt_type_name_unnamespaced(st.id()))
-            ).into(),
+                let str_view = self.c.formatter.fmt_str_view_name(encoding);
+                format!(
+                    "{{reinterpret_cast<const {str_view}*>({cpp_name}.data()), {cpp_name}.size()}}"
+                )
+                .into()
+            }
```
markjm, line 1023: these lines are the only real change in this file, but the GHA is failing on rust fmt of the whole file - lmk how youd like me to handle that!

Manishearth: I'm surprised the old code didn't format. This is fine, I can review this.

Manishearth: No, there's something weird going on with the formatter.

Manishearth: Decided to fix up some other things that got worse here, and pushed.

PR #1130, opened 2026-04-27
Nanobind require `Result` or `Option` Indexers

### LINKED ISSUE #1128: Nanobind must always Raise `IndexError` on invalid access for `indexer`
As per [Python's sequences documentation](https://docs.python.org/3/reference/datamodel.html#sequences), you *MUST* raise an `IndexError`. The same goes for dictionaries, but you must raise `KeyError`.

If this behavior isn't implemented, it leads to issues like those found in https://github.com/rust-diplomat/diplomat/pull/1127, where Python is accessing things it shouldn't through iteration. You can sidestep this in bound types by implementing `__iter__`, but this is a trap that we should tell new users about.

I think we should require that all indexers in Nanobind be of type `Result<...>` or `Option<...>` (per #927). We set a [Lookup Error (or a subclass that implements it)](https://docs.python.org/3/library/exceptions.html#LookupError), and add the additional `E` return type as an argument to that exception.

tyler-zeromatter:
Closes https://github.com/rust-diplomat/diplomat/issues/1128


tyler-zeromatter:
Not sure if we want a config to disable this behavior, but for right now if Python assumes that every class that implements `__getitem__` is a sequence, then we should account for that (and enforce it) with this error.

Manishearth (approved):

tyler-zeromatter:
Will merge after checks

PR #1131, opened 2026-04-29
Partial comparison support

### LINKED ISSUE #1103: Support for partial comparison bindings
I'm dealing with structs containing `f32` in Diplomat, which do not support full comparisons. In Python, I can hand write the bindings `__eq__(&self, other : &Self) -> bool`, but I'd like to avoid that if possible.

Python handles cases when comparison functions return `None`, and C++ could be modified to support `std::optional<Comparison>`. 

I propose rewriting `SpecialMethod::Comparison` to take `SpecialMethod::Comparison(bool)`, where `bool` determines whether the comparison is nullable. We'd add a `nullable_comparisons` flag for each backend to mark whether they support `Option<std::cmp::Ordering>` as a return type:

```rs

#[diplomat::attr(auto, comparison)]
#[diplomat::cfg(supports=nullable_comparison)]
pub fn compare(&self, other : &Self) -> Option<std::cmp::Ordering>
```

tyler-zeromatter:
Closes https://github.com/rust-diplomat/diplomat/issues/1103.

Just to note, this currently includes partial comparisons as information inside of `SpecialMethod::Comparison`. They can theoretically be made as part of their own flag. This would ensure `#[diplomat::attr(auto, ...)]` works and removes the mutation in lowering code, at the cost of not being as concise when writing bindings.

TODOs:

- [ ] Docs

### DIFF
--- a/core/src/hir/attrs.rs
+++ b/core/src/hir/attrs.rs
@@ -269,8 +269,10 @@ pub enum SpecialMethod {
     Setter(Option<String>),
     /// A stringifier. Must have no parameters and return a string (DiplomatWrite)
     Stringifier,
-    /// A comparison operator. Currently not universally supported
-    Comparison,
+    /// A comparison operator. Currently not universally supported.
+    ///
+    /// bool is true if the comparison is partial (i.e., [`Option<std::cmp::Ordering>`]).
+    Comparison(bool),
     /// An iterator (a type that is mutated to produce new values)
     Iterator,
     /// An iterable (a type that can produce an iterator)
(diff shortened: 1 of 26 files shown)


tyler-zeromatter:
Just as a heads up @Manishearth, the 3 open PRs are all the features I'm interested in, I wouldn't expect any more big PRs for a while.

Anything else will most likely be bugfixes as I uncover more bugs with the Nanobind backend.

Sorry to inundate your inbox, I will hold off on anything more until these are merged and you have some more time to actually review. Just thought I should get these three out of the way since they're important features that I'd like to test internally (utilizing the PRs CI to check for any obvious mistakes).

Manishearth (approved):
quite straightforward, nice!

Manishearth:
Yep, and I appreciate having these opened early so at least you can make progress if I notice things!

PR #1140, opened 2026-05-02
Fix nanobind stub type for diplomat::result to show success type instead of 'result'

markjm:
The `type_caster` for `diplomat::result<T, E>` used `const_name("result")` as its `Name`, causing nanobind's stubgen to emit `-> result` in .pyi stubs. Since the caster unwraps on success (returning `T`) and raises on error, the Python-visible return type is `T`. Use `Caster::Name` to forward the inner type's name so stubs correctly show e.g. `-> int` or `-> MyType`.

Concretely, this fixes a stubgen issue in icu4x from

```python
class ZonedDateTime:

    @staticmethod
    def full_from_string(v: str, calendar: Calendar, iana_parser: IanaParser, _offset_calculator: VariantOffsetsCalculator) -> result: ... # result is just an undefined token
```
to

```python
class ZonedDateTime:

    @staticmethod
    def full_from_string(v: str, calendar: Calendar, iana_parser: IanaParser, _offset_calculator: VariantOffsetsCalculator) -> ZonedDateTime: ...
```

There is not currently any testing infrastructure for the types nanobind introspects and creates from our bindings, so i instead opted for a test which shows that the intermediate step of `__doc__` which I believe is what nanobind uses to create these stubs

### DIFF
--- a/feature_tests/nanobind/src/include/diplomat_nanobind_common.hpp
+++ b/feature_tests/nanobind/src/include/diplomat_nanobind_common.hpp
@@ -133,7 +133,7 @@ namespace nanobind::detail
         bool is_ok;
         Py_ssize_t size;
         using Caster = make_caster<U>;
-        static constexpr auto Name = const_name("result");
+        static constexpr auto Name = Caster::Name;
 
         static handle from_cpp(somelib::diplomat::result<T, E> value, rv_policy p, cleanup_list *cl) noexcept
         {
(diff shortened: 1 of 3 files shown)


tyler-zeromatter:
Ah, I didn't realize the names were used in stub generation. I wonder if there's any way in Python to mark a function as returning an exception? But I guess that would be out of the scope of this PR

tyler-zeromatter (approved):

markjm:
As far as there is not a real great way to type exceptions in most python typecheckers 🤷

tyler-zeromatter:
Maybe this is just something we could optionally include as an extra docstring for each Result type in the generation:

```py
'''
Throws FailType on failure.
'''
def will_throw() -> SuccessType:
    pass
```

PR #1147, opened 2026-05-07
C++: Composable conversions

### LINKED ISSUE #1141: C++: Composable Conversions
Per discussion here: https://github.com/rust-diplomat/diplomat/pull/1106#discussion_r3197622556

`NamedExpression`'s conversions should be composable, to reduce clutter across conversion code generation.

tyler-zeromatter:
Closes https://github.com/rust-diplomat/diplomat/issues/1141.

TODOs:

- [ ] Rebase on top of main and ensure functionality with Nanobind

### DIFF
--- a/tool/src/cpp/gen.rs
+++ b/tool/src/cpp/gen.rs
@@ -419,11 +419,9 @@ impl<'ccx, 'tcx: 'ccx> ItemGenContext<'ccx, 'tcx, '_> {
         let cpp_to_c_fields = def
             .fields
             .iter()
-            .enumerate()
-            .map(|(idx, field)| {
+            .map(|field| {
                 self.gen_cpp_to_c_for_field(
                     is_in_mut_struct,
-                    idx,
                     field,
                     namespace.clone(),
                 )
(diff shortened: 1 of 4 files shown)


tool/src/cpp/gen.rs
```diff
@@ -83,10 +83,36 @@ pub struct MethodInfo<'a> {
...
+}
+
+impl<'a> ConversionExpression<'a> {
+    fn empty() -> Self {
+        Self {
+            fmt_str: "{0}".into()
+        }
+    }
+
+    fn fmt(&self, inner_name : &str) -> String {
```
Manishearth, line 98: hmm, I don't think this is what I had in mind.

I'll have to think about this more.

Manishearth (changes requested):
Yeah now that I think about this I think this makes the code messier. I think there are ways to make the API here more composable but doing a simple templating engine isn't it.

tyler-zeromatter:
Alright, I need to tinker with this internally some more, then.

PR #1164, opened 2026-05-29
Add .NET backend

irvingoujAtDevolution:
## Summary

Adds a `.NET` target language to `diplomat-tool` that generates idiomatic C# bindings against the Diplomat C ABI via P/Invoke. Devolutions has been carrying an out-of-tree .NET backend against a pre-HIR fork for years; this PR brings a fresh HIR-era implementation upstream so consumers like picky-rs and IronRDP can move off the fork.

Invoke via `diplomat-tool dotnet <out>`.

## Generated layout

Two layers per Rust type:

- **Raw layer** (`{namespace}.Raw`): `[DllImport]` externs on a `partial struct`, mirrors the C ABI directly with pointer types and the `Diplomat*` runtime structs.
- **Idiomatic layer** (`{namespace}`): `public partial class T : IDisposable` for opaques (holds the raw handle in `_inner`, has a finalizer), `public partial struct T` for value structs. Exposes PascalCase members, throws on `Result<_, E>`, returns `T?` for pointer-nullable `Option`.

Runtime helpers (`DiplomatSliceU8` / `U32`, mutable variants, `DiplomatWriteable`) are emitted once per generation as Askama templates so the namespace stays project-specific.

## Supported HIR shapes

- Opaques (incl. `#[diplomat::opaque_mut]`), structs (primitive fields only — non-primitive fields panic with a descriptive message during codegen), enums.
- Method I/O for primitives, opaques (incl. `Option<Box<T>>` and `Option<&T>`), structs, enums, `&DiplomatStr`, `&[u8]` / `&mut [u8]`, `&[u32]` / `&mut [u32]`, `&mut DiplomatWriteable`.
- `Result<T, E>` where `E` is opaque / enum / struct / unit `()`. Auto-generates `{Name}Exception` classes per error type (configurable `exception_trim_suffix`). Unit errors throw the built-in `InvalidOperationException`.
- `Option<Box<T>>` / `Option<&T>` for opaque `T` (pointer-nullable on the wire — no wrapper struct).
- `Option<T>` for primitives / structs / enums via a tagged `DiplomatOption{T}` wrapper.

## iOS support

Generated `[DllImport]` references are wrapped in `#if __IOS__` to resolve to `lib{name}.framework/lib{name}` (framework bundle layout) on iOS, with the bare `{name}` form elsewhere. Reverse callbacks in `DiplomatWriteable` (`Flush`, `Grow`) carry `[MonoPInvokeCallback(typeof(...))]` under the same guard so Mono/Xamarin AOT can pin them. Replays a known-good fix Devolutions has carried out-of-tree.

## Not yet covered

`attr_support()` is conservative — these flags are `false` so the HIR validator rejects them at lowering rather than the backend panicking during generation:

- Callbacks (template scaffold exists at `tool/templates/dotnet/callback.cs.jinja`, gated off)
- Accessors / static accessors, stringifiers, comparators, iterators / iterables, indexing
- Constructors / named / fallible constructors
- Traits, custom errors, mocking interface
- Static slices, owned slices
- Method overloading: claimed true, but real coverage is partial

Primitive slices only lower for `&[u8]` / `&[u32]` and their mutable counterparts today; other element types hit a descriptive `unimplemented!()` panic in the slice arm. Picky-rs and IronRDP — the consumers driving this backend — exercise only the supported subset.

## Test plan

- [x] `cargo check -p diplomat-tool` clean
- [x] Regenerate picky-rs (`Devolutions.Picky`) bindings end-to-end — 261 files emitted, all compile against picky's hand-written addons
- [x] Spot-check generated C# for one struct, one opaque, one Result, one Option case — layout matches Rust `repr(C)`
- [ ] Reviewers: please flag any HIR shape you'd expect a downstream consumer to need that's currently in the "not covered" list — I'd rather extend coverage in follow-ups than ship a backend that silently lowers things wrong


Manishearth:
Wonderful, thank you! I'll try and have a look, but I'm rather behind on Diplomat reviews right now.

Manishearth:
> attr_support() is conservative — these flags are false so the HIR validator rejects them at lowering rather than the backend panicking during generation:

Thank you, this is the right call. Diplomat's attribute system is _explicitly_ so that backends do not need to support everything (even if they could). So yes, be conservative here, and don't feel pressured to add feature support until you need it.

Manishearth:
Nothing the copilot review brought up seems major and you should feel free to ignore it, but addressing it would be good regardless.

irvingoujAtDevolution:
Hi @Manishearth , thanks for quick reviewing, I still have pretty dirty git stashes/branches etc... all over the place, and I'll clean up and get it ready for review soon.

irvingoujAtDevolution:
Ready for review!

irvingoujAtDevolution:
@Manishearth It's ready now, thanks!

Manishearth:
Please make sure fmt/clippy passes, see CI.

Manishearth:
I plan to review this though it will take some time.

irvingoujAtDevolution:
@Manishearth Let me know if you need anything from myside

Manishearth (changes requested):
This should also generate `feature_tests`. That's where most functionality is.

This should contain tests for the code (feature_tests has a bunch of extra folders for per-language tests). It's fine if you can't figure out how to CI this on our CI, but I think it should be possible to test with dotnet on Linux.

@irvingoujAtDevolution has the output of this impl been tested out in your use cases yet? or are you waiting for this to merge first? Becuase I cannot review .NET ffi code in depth the best test of this will be actual usage.

tool/src/dotnet/gen/opaque.rs
```diff
@@ -0,0 +1,89 @@
+//! Opaque-type codegen.
```
Manishearth, line 1: nit: this file should just be opaque.rs, it doesn't have any child modules so it doesn't need a folder.

tool/templates/dotnet/opaque.raw.cs.jinja
```diff
@@ -0,0 +1,25 @@
...
+using {{ namespace }};
+using {{ namespace }}.Diplomat;
+
+namespace {{ namespace }}.Raw;
+
+[StructLayout(LayoutKind.Sequential)]
+public partial struct {{ name }}
+{
+#if __IOS__
+    private const string NativeLib = "lib{{ dylib_name }}.framework/lib{{ dylib_name }}";
```
Manishearth: nit: should this be a shared constant somewhere?

tool/templates/dotnet/option.raw.cs.jinja
```diff
@@ -0,0 +1,26 @@
+using System;
+using System.Runtime.InteropServices;
+
+namespace {{ namespace }}.Raw;
+
+using {{ namespace }};
+
+[StructLayout(LayoutKind.Sequential)]
+public partial struct {{ option_struct_name }}
```
Manishearth: question: are we unable to use some generic union type for this? It's fine, just feels like more codegen per type

tool/templates/dotnet/DiplomatWriteable.cs.jinja
```diff
@@ -0,0 +1,157 @@
...
+/// Caller-provided writer that Rust appends UTF-8 bytes into.
+/// Use <see cref="ToUnicode"/> after a Rust call returns to extract the
+/// written content as a managed string. Dispose to free the underlying
+/// unmanaged buffer.
+/// </summary>
+// Mirrors the Rust `DiplomatWrite` repr(C) layout. The `byte growFailed`
+// field corresponds to Rust's `grow_failed: bool`; Sequential layout pads
+// it to the next 8-byte boundary so `flush`/`grow` land at offsets 40/48.
+[StructLayout(LayoutKind.Sequential)]
+public struct DiplomatWriteable : IDisposable
```
Manishearth: issue: this should probably be internal not public?

tool/templates/dotnet/method_body.cs.jinja
```diff
@@ -0,0 +1,125 @@
...
+            {{ statement }}
+{%- endfor %}
+            {
+{%- endif %}
+{%- if method.return_type.is_write() %}
+            {# Benoit-compat surface: caller supplies the writer, we never
+               allocate or dispose internally. Method returns void; the writer
+               fills with the formatted string.
+
+               The signature is `ref DiplomatWriteable writeable` so the writer
```
Manishearth: issue: I don't think we should have `DiplomatWriteable` on the public API. Perhaps a higher level interface type that you pass in, but DiplomatWriteable is supposed to be a low level FFI type. 

We should only generate the API that returns a string here.

tool/templates/dotnet/method_body.cs.jinja
```diff
@@ -0,0 +1,125 @@
...
+               won't be moved by the GC. #}
+{{ bi }}fixed (DiplomatWriteable* writeablePtr = &writeable)
+{{ bi }}{
+{%- if let Some(info) = method.error_info %}
+{{ bi }}    var result = Raw.{{ name }}.{{ method.name }}({{ method.inputs.raw_call_args }}{% if !method.inputs.raw_call_args.is_empty() %}, {% endif %}writeablePtr);
+{{ bi }}    if (!result.IsOk)
+{{ bi }}    {
+{%- if info.error.is_unit() %}
+{{ bi }}        throw new {{ info.exception_name }}("FFI function failed with unit error");
+{%- else if info.error.is_opaque() %}
```
Manishearth: nit: I'm finding a lot of these if blocks harder to read due to a lack of indentation. The `{%-` should mean we can indent these anyway. Can we do that?

tool/templates/dotnet/method_body.cs.jinja
Manishearth: If this is a compat thing, I'd love to learn more!

example/dotnet/Generated/RawDataProvider.cs
```diff
@@ -0,0 +1,25 @@
...
+public partial struct DataProvider
+{
+#if __IOS__
+    private const string NativeLib = "libdiplomat_example.framework/libdiplomat_example";
+#else
+    private const string NativeLib = "diplomat_example";
+#endif
+
+    [DllImport(NativeLib, EntryPoint = "icu4x_DataProvider_new_static_mv1", CallingConvention = CallingConvention.Cdecl)]
+public static unsafe extern DataProvider* NewStatic();
```
Manishearth: issue: these raw APIs shouldn't be public I think

Manishearth (commented):
Overall looks quite good! Code is well structured and this makes good use of the templating system

tool/src/dotnet/gen/method.rs
```diff
@@ -0,0 +1,1180 @@
...
+}
+
+// ─────────────────────────────────────────────────────────────────────────────
+// Method info
+// ─────────────────────────────────────────────────────────────────────────────
+
+/// Pre-processed view of a single HIR `Method`. Carries both raw-layer and
+/// idiomatic-layer render data — templates pick the side they want.
+#[derive(Clone)]
+pub(super) struct MethodInfo<'ctx> {
```
Manishearth, line 407: question: one thing: I don't see any handling of borrowing here.

Is that intentional? Borrow handling is definitely trickier so it's fine to figure out later but we might want to document it.

irvingoujAtDevolution:
Thanks for the review, though it is not used by upstream in production, I do have an experiemental branch on IronRdp, which currently uses what we have on this branch, and it works great
https://github.com/Devolutions/IronRDP/tree/test/diplomat-hir-dotnet-avalonia-clean

tool/src/dotnet/gen/opaque.rs
irvingoujAtDevolution: Done, flattened this to `tool/src/dotnet/gen/opaque.rs`; the `opaque/mod.rs` module folder is no longer tracked.

tool/templates/dotnet/opaque.raw.cs.jinja
irvingoujAtDevolution: Done, I added a shared generated `DiplomatNativeLib.Name` constant and raw externs now use that instead of repeating the native library string.

tool/templates/dotnet/option.raw.cs.jinja
irvingoujAtDevolution: Updated. This is raw/internal ABI only; the public C# API exposes native nullable types (`T?`) instead of
  `DiplomatOption*`.

tool/templates/dotnet/DiplomatWriteable.cs.jinja
irvingoujAtDevolution: We used to depends on the old diplomat fork that uses public `DiplomatWriteable` with in place mutation,however we will keep it align with what this repo expect here. Updated.

tool/templates/dotnet/method_body.cs.jinja
irvingoujAtDevolution: updated, unified to single method

tool/templates/dotnet/method_body.cs.jinja
irvingoujAtDevolution: Updated. This was mirroring the old Diplomat .NET backend, where generated bindings exposed public `DiplomatWriteable` writer-style overloads in downstream users like IronRDP. I agree that should stay raw/internal for this backend, so the public API now returns `string` and keeps `DiplomatWriteable` internal.

example/dotnet/Generated/RawDataProvider.cs
irvingoujAtDevolution: updated

tool/src/dotnet/gen/method.rs
irvingoujAtDevolution: we are looking for support it further in subsequent PR, let's try keep this one smaller for review purpose.

irvingoujAtDevolution: updated comments, if you want, I can create a issue

irvingoujAtDevolution:
We currently have a lot disabled feature test, which we are looking for support in subsequent PRs. Thanks!

tool/src/dotnet/gen/method.rs
Manishearth: My issue is with supporting an inherently unsound backend from the get go. I'd feel a bit more comfortable if it rejected any APIs with borrows across a function boundary (and you can disable APIs like that with `#[diplomat::cfg(not(dotnet))]` in the tests)

Manishearth:
> We currently have a lot disabled feature test, which we are looking for support in subsequent PR

I don't understand what you mean here. Are you saying that a lot of the feature tests needed to be disabled so you didn't enable them here, opting for a future PR? That's fine I guess, but I'd still want an end to end test that can be run that shows this _works_.

Manishearth:
Okay, if the issue is that you are missing a lot of the features, I think my _preference_ is that you try and enable the tests anyway and be pretty liberal in what tests you disable. If you don't want to do that, I think I'd still like to see an end-to-end test in this PR.

Since I am not a .NET expert, I really would like to see that all of this _works_ end-to-end (and have some tests for that. Then we can land and come up with a plan for the rest. Doing this using a partially-disabled feature_tests would be ideal, but doing it with the example also seems fine.

tool/src/dotnet/gen/method.rs
Manishearth: But don't consider this to be a hard blocker for initial landing. This is a preference of mine.

irvingoujAtDevolution:
> > We currently have a lot disabled feature test, which we are looking for support in subsequent PR
> 
> I don't understand what you mean here. Are you saying that a lot of the feature tests needed to be disabled so you didn't enable them here, opting for a future PR? That's fine I guess, but I'd still want an end to end test that can be run that shows this _works_.

Hi @Manishearth , sorry that I phrased if poorly. whtat I actually meant is that we do have tests, but because this PR is not fully implementing every feature that dotnet could support, so I disabled them.

irvingoujAtDevolution:
> Okay, if the issue is that you are missing a lot of the features, I think my _preference_ is that you try and enable the tests anyway and be pretty liberal in what tests you disable. If you don't want to do that, I think I'd still like to see an end-to-end test in this PR.
> 
> Since I am not a .NET expert, I really would like to see that all of this _works_ end-to-end (and have some tests for that. Then we can land and come up with a plan for the rest. Doing this using a partially-disabled feature_tests would be ideal, but doing it with the example also seems fine.

Ah, I think there was some miscommunication. The .NET feature_tests path is enabled in this PR: we generate `feature_tests/dotnet/Generated` and run the xUnit tests via `test-dotnet-
  feature`.

  What I meant is that some individual shared feature-test APIs are disabled for the dotnet backend with `#[diplomat::attr(dotnet, disable)]`, because those HIR shapes are not supported
  yet. The subset we do support is enabled and tested end-to-end.

irvingoujAtDevolution:
I'll add more tests to boost your confidence.

tool/src/dotnet/gen/method.rs
irvingoujAtDevolution: Yes, I agree. My apologies.

This backend was initially migrated from an older diplomat and a fork of .NET backend that did not model Rust lifetime soundness very rigorously; it mostly relied on caller discipline. In practice, the caller was responsible for ensuring that a borrowed type did not outlive its C# wrapper.

The current .NET backend supports borrowed inputs, including opaque handles, strings, and primitive slices. For synchronous functions without callbacks, this is relatively contained: the generated wrapper pins or converts the input, and the borrowed reference is only valid for the duration of the call.

The more concerning case is borrows that cross the function boundary. We already reject direct borrowed opaque returns such as `&T`, `&mut T`, and `Option<&T>`, because the current C# wrapper assumes ownership and would be unsound in those cases.

What remains is owned opaque returns whose Rust type carries a lifetime, for example `Box<T<'a>>`, where the returned object may still borrow from an input or from `self`. C# cannot enforce that lifetime relationship in the type system today.

For this PR, I added generated XML documentation on methods where the HIR borrow analysis detects such an output lifetime. The documentation explicitly states that the caller must keep the borrowed backing storage alive and undisposed while the returned value is in use.

I think this matches the current backend behavior without pretending it is fully lifetime-safe. A future PR can either reject these lifetime-carrying owned returns for .NET, or implement proper owner-edge tracking similar to the managed backends that retain borrowed-from objects.

irvingoujAtDevolution:
@Manishearth Thanks for the review.

I think I misunderstood the lifetime concern at first. I added more tests and tightened the policy a bit.

Main changes:

1. Added more .NET tests for the stuff this PR supports:
   - opaque lifecycle / dispose: [`OpaqueTests.Dispose_IsIdempotent`](feature_tests/dotnet/Tests/OpaqueTests.cs#L49)
   - primitive return: [`OpaqueTests.ReturnsUsize_Is412`](feature_tests/dotnet/Tests/OpaqueTests.cs#L27)
   - result/error paths: [`ResultOpaqueTests`](feature_tests/dotnet/Tests/ResultOpaqueTests.cs#L10)
   - borrowed byte slice input: [`Float64VecTests.NewF64BeBytes_PinsByteSliceAndExposesValues`](feature_tests/dotnet/Tests/Float64VecTests.cs#L9)
   - rename / ABI rename / attrs: [`AttrTests`](feature_tests/dotnet/Tests/AttrTests.cs#L10)

2. Added backend unit tests for the lifetime edge cases:
   - borrowed opaque error rejected: [`borrowed_opaque_error_is_rejected`](tool/src/dotnet/mod.rs#L504)
   - direct borrowed opaque return rejected: [`borrowed_opaque_return_is_rejected`](tool/src/dotnet/mod.rs#L529)
   - `Box<T<'a>>` borrowing from a temporary string/slice input rejected: [`lifetime_carrying_owned_return_borrowing_slice_input_is_rejected`](tool/src/dotnet/mod.rs#L554)
   - `Box<T<'a>>` borrowing from `self` / opaque wrapper gets lifetime docs: [`lifetime_carrying_owned_return_borrowing_opaque_gets_warning`](tool/src/dotnet/mod.rs#L581)

3. Fixed the actually-unsound case: if a returned object borrows from a C# string/slice input, we now reject it instead of only warning. The issue is that those inputs are temporary buffers inside the generated wrapper, so the caller can't keep them alive anyway. That logic is in [`borrowed_output_lifetime_warning`](tool/src/dotnet/gen/method.rs#L697), with the diagnostic here: [`method.rs`](tool/src/dotnet/gen/method.rs#L726).

So now the split is:
- borrowed inputs are supported for call-scoped sync calls
- direct borrowed opaque returns are rejected
- returns borrowing from temporary string/slice inputs are rejected
- returns borrowing from `self` / opaque wrappers still generate, but with lifetime docs

irvingoujAtDevolution:
@Manishearth let me know what you think! Thanks.

tool/src/dotnet/gen/method.rs
Manishearth: I'm happy to have this be documented for now, that is a quick solution! We should file a followup for this, though.

Manishearth (approved):
Great! And thank you for adding CI as well.

I think this is landable, and you should file followups for:

 - Adding proper borrowing support
 - For each of the issues marked disable in https://github.com/rust-diplomat/diplomat/pull/1164/changes/654018928a79b5d2b0a84c9ae978c2e4215a296a. Ideally, update the code to mention those issues.
 - Anything else I've missed.

Manishearth:
Thanks for your work! There's definitely more work to be done here but this is good enough to land for now, and I'm happy to let that happen in smaller, more scoped PRs.

Since you're adding a new backend that is not supported by the primary Diplomat maintainers (similar to how Kotlin was), I want to make sure you understand that support for this backend will be best-effort from us.

_By and large_, modern Diplomat HIR does not change often, and we try not to change it in ways that will cause a delta in codegen. BUT that could theoretically happen, and in that case we'll try and fix any issues.  If we can't, we might temporarily disable CI and cc you for help.

One of the reasons I wanted there to be end to end tests was that so the rest of us can at least _try_ to poke at this backend when we need to. When I say support is best effort it doesn't mean we are fine breaking things, just that that might end up happening if we can't keep things working. Having tests makes it easier for us to try :smile: 

In case of much larger architectural changes, we may not be able to guarantee migrating .NET. This is _extremely_ unlikely: HIR was the main large architectural change we had to do and we explicitly designed it so that we will never have to do it again. Furthermore, I suspect LLMs would be helpful for such migrations in the future.

If there are other maintainers from your side that you would like cc'd (is Benoit still involved), do let me know.

Manishearth:
Oh, also, we discuss Diplomat on the Unicode Slack. email me at manish@unicode.org if you'd like an invite to it.

irvingoujAtDevolution:
Thank you! 
I'll create subsequent PR to address these issues.
Also, feel free to @CBenoit and me for any issues.

CBenoit:
Thank you very much for reviewing!! @Manishearth 
And thank you @irvingoujAtDevolution, it will be great to be able to update to newer Diplomat versions 🎉 
I think, mostly Irving will, but as Devolutions we’ll continue iterating to fix the remaining issues.
Soundness issues about borrowing is a long standing problem we already had in the previous backend that was mostly inspired by the C++ backend at the time, but now that we have an easier to work with model (HIR) and other quality reference backends it should be easy enough to fix, hopefully!
I’m still interested to be cc'd when necessary. Let us know! Thank you!

PR #1169, opened 2026-06-09
Use `proc_macro2::Span` for expanded error handling in AST

### LINKED ISSUE #1028: Wrap `proc_macro2::Span` in AST for improved erroring
Follows from discussion in #1024.

Per discussion there, we'll wait until ICU4X has MSRV 1.88.

(Or we can simply feature gate the wrapped type in `ast::Ident` if we really want to)

tyler-zeromatter:
Closes https://github.com/rust-diplomat/diplomat/issues/1028.

Some TODOs:

- [x] Fix Lifetime span generation
- [x] Fix RustLink generation?
- [x] Span tracking snapshot tests (with `insta` for both?)
- [x] Get passing tests
- [x] Drop new crate deps

### DIFF
--- a/core/src/ast/idents.rs
+++ b/core/src/ast/idents.rs
@@ -1,13 +1,38 @@
-use proc_macro2::Span;
 use quote::{ToTokens, TokenStreamExt};
 use serde::{Deserialize, Serialize};
 use std::borrow::{Borrow, Cow};
 use std::fmt;
+use std::ops::Range;
+
+/// Equivalent to `proc_macro2::Span`.
+#[derive(Hash, Eq, PartialEq, Serialize, Clone, Debug)]
+pub struct Span {
+    /// The row where we are in a file (Alternatively, the offset by the number of newline characters)
+    pub start_line: usize,
+    /// The column where we are in a file (Alternatively, the offset by newline characters plus this amount)
+    pub column_line: usize,
+    /// The range in bytes of the span. Used mostly to determine length.
+    pub range: Range<usize>,
+    /// Equivalent to `proc_macro2::Span::file`, but we have to allow for internal testing or
+    /// spaces where we simply don't know where we've acquired this Span from.
+    pub span_location: SpanLocation,
+}
+
+#[derive(Hash, Eq, PartialEq, Serialize, Clone, Debug)]
+#[non_exhaustive]
+pub enum SpanLocation {
+    /// For testing or when accessing SpanLocation is not possible.
+    /// Will panic on trying to render an error (we need to know where a file is located to properly display).
+    None,
+    /// An absolute path to a rust file.
+    FilePath(String),
+    /// Does not exist in a file, full source text that is used locally.
+    LocalSource(String),
+}
 
 /// An identifier, analogous to `syn::Ident` and `proc_macro2::Ident`.
-#[derive(Hash, Eq, PartialEq, Serialize, Clone, Debug, Ord, PartialOrd)]
-#[serde(transparent)]
-pub struct Ident(Cow<'static, str>);
+#[derive(Eq, Serialize, Clone, Debug)]
+pub struct Ident(Cow<'static, str>, Option<Span>);
 
 impl Ident {
     /// Validate a string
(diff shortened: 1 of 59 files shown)


tyler-zeromatter:
To prove that this fails correctly, here's the CI run: https://github.com/rust-diplomat/diplomat/actions/runs/27242799756/job/80450142590?pr=1169#step:5:72

Manishearth:
Can we use `annotate-snippets` instead of `miette`? miette is good but i'm not sure if it's still maintained and it has far more dependencies.

Also `diplomat-core` should absolutely not have another dependency in macro mode. If we have to introduce a dummy Span type to make it work, so be it.

I think we should make it optional and make the "good" version of `panic_with_report` (etc) behind a separate feature. This way even hir users don't _have_ to use it either.

Could we have one PR to just add good span tracking with the right panic API, but the panic API just formats things better? And then we make it look good using annotate-snippets separately. Even better would be to split it further so we have a PR that _just_ adds span tracking but doesn't use it yet, now that you have an end-to-end proof of concept.

tyler-zeromatter:
Yeah, I'm happy to incorporate just Span tracking with this PR, and we can increment with other features at a later date.

tyler-zeromatter:
Need to handle some of the TODOs above, then I will mark this as ready for review.

tyler-zeromatter:
So, one problem I'm experiencing with insta tests: proc_macro2 does not output accurate span information. I'd really like to be able to test `Span` information with our existing insta tests, so there seems to be a pretty simple workaround:

```rs
macro_rules! parse_quote_stringify {
   ($($tt:tt)*) => { syn::parse_str(quote::quote!($($tt)*)).to_string()) };
}
```

Converting to and back from a string creates the proper span information for testing. Will implement with this for all of our AST tests, so I don't have to make separate tests just for span information.

tyler-zeromatter:
`stringify!` also works, although it seems to be arbitrarily chopping up newlines (although I might switch to this because I believe it's slightly faster than converting to/from strings)

tyler-zeromatter:
The alternative is to make tests that just cover what we expect from `Span`, which I think might be what I ultimately go with since a lot of these parsing functions are finnicky and subject to change (especially when it comes to whitespace)

Manishearth:
I think I'm okay not testing span info in our insta tests and testing it some other way.

core/src/ast/idents.rs
```diff
@@ -1,52 +1,115 @@
...
+    /// Equivalent to `proc_macro2::Span::file`, but we have to allow for internal testing or
+    /// spaces where we simply don't know where we've acquired this Span from.
+    pub span_location: SpanLocation,
+}
+
+#[derive(Hash, Eq, PartialEq, Serialize, Clone, Debug)]
+#[non_exhaustive]
+pub enum SpanLocation {
+    /// For testing or when accessing SpanLocation is not possible.
+    /// Will panic on trying to render an error (we need to know where a file is located to properly display).
```
Manishearth: issue: should we panic? won't this be a problem for spans from macros? We should render `<location unknown>` or something

Manishearth (approved):

core/src/ast/idents.rs
tyler-zeromatter: Part of the reason we waited for MSRV 1.88 is that Rust stabilized spans from proc macros in 1.88: https://blog.rust-lang.org/2025/06/26/Rust-1.88.0/

tyler-zeromatter: I think I might have misread your concern. For `SpanLocation`, proc_macros also have `input.span().file()` stabilized in 1.88, which allows us to set filepaths as the location:

https://doc.rust-lang.org/stable/proc_macro/struct.Span.html#method.file

tyler-zeromatter: Currently I have the macro use `input.span().file()` in this PR, although I think it should be `input.span().local_file()` based on the docs. Updating

tyler-zeromatter: `local_file` returns an `Option`, so a `None` location shouldn't panic anymore; we now have to account for the fact that `#[diplomat::bridge]` might be used in a place where proc macro can't determine the origin of. Updated the docstring and marking this as resolved.

Manishearth (approved):

PR #1173, opened 2026-06-12
Add Clang lifetime annotations to C++ backend

### LINKED ISSUE #390: Consider adding lifetime annotations to CPP2
There are multiple solutions to lifetimes in C++:
* `ABSL_ATTRIBUTE_LIFETIME_BOUND`
* `[[clang::lifetimebound]]`
* and maybe more

Manishearth:
Fixes https://github.com/rust-diplomat/diplomat/issues/390

Pretty straightforward, we have Fancy Lifetime Infra, and clang needs Basic Lifetime Infra, so it's very simple to get the few bits of information we actually need out of it.

### DIFF
--- a/tool/src/cpp/gen.rs
+++ b/tool/src/cpp/gen.rs
@@ -6,6 +6,7 @@ use crate::config;
 use crate::read_custom_binding;
 use crate::ErrorStore;
 use askama::Template;
+use diplomat_core::hir::borrowing_param::ParamBorrowInfo;
 use diplomat_core::hir::CallbackInstantiationFunctionality;
 use diplomat_core::hir::IncludeLocation;
 use diplomat_core::hir::IncludeSource;
@@ -28,6 +29,7 @@ pub struct NamedType<'a> {
     pub(crate) type_name: Cow<'a, str>,
     /// Default value (for method params, but could eventually be for structs).
     pub(crate) default_value: Option<Cow<'a, str>>,
+    pub(crate) lifetimebound: bool,
 }
 
 impl<'a> NamedType<'a> {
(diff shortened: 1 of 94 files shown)


tyler-zeromatter:
Awesome! Will take a look in just a sec

tool/src/cpp/gen.rs
```diff
@@ -507,6 +509,20 @@ impl<'ccx, 'tcx: 'ccx> ItemGenContext<'ccx, 'tcx, '_> {
...
             .namespace_c_name(id, method.abi_name.as_str());
+
+        let mut this_borrowed = false;
+        let mut visitor = method.borrowing_param_visitor(self.c.tcx, false);
+
+        if let Some(param_self) = method.param_self.as_ref() {
+            let info = visitor.visit_param(&param_self.ty.clone().into(), "this");
+            if !matches!(
+                info,
+                ParamBorrowInfo::NotBorrowed | ParamBorrowInfo::TemporarySlice
```
tyler-zeromatter, line 520: nit: I'm worried that `not matches!(...)` is somewhat brittle if we ever add to or change `ParamBorrowInfo`. Maybe we could add a helper `is_borrowed` function to the HIR?

That also would allow some code re-use here.

tool/templates/cpp/runtime.hpp.jinja
```diff
@@ -18,6 +18,14 @@
 #include <array>
 #endif
 
+#ifndef DIPLOMAT_LIFETIME_BOUND
+#if defined(__clang__)
+#define DIPLOMAT_LIFETIME_BOUND [[clang::lifetimebound]]
+#else
+#define DIPLOMAT_LIFETIME_BOUND
```
tyler-zeromatter: Suggestion: MSVC also has a similar attribute: https://learn.microsoft.com/en-us/cpp/code-quality/c26815?view=msvc-170

tyler-zeromatter: I believe `[msvc::lifetimebound]` works interchangeably with `[clang::lifetimebound]`

tyler-zeromatter (commented):
Looks good! I have one suggestion and one nit in my review comments.

In addition, a nice-to-have would be a test to ensure that clang produces the right warnings when using a method improperly.

tyler-zeromatter (approved):

tool/src/cpp/gen.rs
Manishearth: I think this is tricky: TemporarySlice is also "borrowed". This is talking about different types of borrows and it's very use case specific to figure out which ones you care about.

tyler-zeromatter: That makes sense, I think this setup is fine then.

tool/templates/cpp/runtime.hpp.jinja
Manishearth: Good call!

PR #1174, opened 2026-06-13
agent-driven: Fix Dart fallible struct constructors

### LINKED ISSUE #1123: Dart Issues with Fallible Struct Constructors
There are a number of issues here, not sure what the cause for each is.

I have a test struct from https://github.com/rust-diplomat/diplomat/pull/1122, which looks like:

```rs
    #[diplomat::cfg(supports=fallible_constructors)]
    pub struct FallibleOpaqueConstructor {
        x: i32,
    }

    impl FallibleOpaqueConstructor {
        pub fn ctor() -> Result<Self, Box<ResultOpaque>> {
            Err(Box::new(ResultOpaque(10)))
        }
    }
```

Dart generates the function

```dart
  factory FallibleOpaqueConstructor({int? x}) {
    final result = _FallibleOpaqueConstructor_ctor();
    final dart = FallibleOpaqueConstructor._fromFfi(result);
    if (x != null) {
      dart.x = x;
    }
    return dart;
  }
```

There are at least two issues here, as far as I can understand.

First, `{int? x}` should not be generated as an argument, as this constructor takes no arguments.

Secondly, Dart complains: `The argument type '_ResultFallibleOpaqueConstructorFfiOpaque' can't be assigned to the parameter type '_FallibleOpaqueConstructorFfi'.` That's because the variable `result` returns a `_ResultFallibleOpaqueConstructorFfiOpaque`, while `_fromFfi` takes `_FallibleOpaqueConstructorFfi`, so there's some conversions that need to be done here.

Manishearth:
(This entire PR, including the body, is agent-driven. I have yet to look at it, I just asked Gemini to fix the issue and it came up with something plausible. I'll have a closer look at some point, but it _seems_ plausible if you want to tkae a look)

This change fixes the Dart backend to correctly handle fallible constructors for structs. Previously, the generator would repurpose the zero-argument constructor to allow setting fields, but it assumed the FFI function returned the struct directly, leading to compile errors when it actually returned a Result FFI struct.

We now check if the constructor is fallible, and if so, we generate the error handling code (throwing the error) and extract the success value from the Result union before applying field overrides and returning.

Fixes #1123


Manishearth:
Reviewed this carefully, and it doesn't seem to have other effect. I refactored out the return handling so we wouldn't duplicate it. Would still love another review from others.

feature_tests/src/result.rs
```diff
@@ -99,7 +99,7 @@ pub mod ffi {
     }
 
     // FIXME: https://github.com/rust-diplomat/diplomat/issues/1123
```
tyler-zeromatter, line 101: Now that this issue is fixed, these TODOs should be removed.

tool/src/dart/mod.rs
```diff
@@ -1053,66 +1084,85 @@ impl<'cx> ItemGenContext<'_, 'cx> {
...
-    fn gen_c_to_dart_for_return_type(
+    /// Generates the return handling code for a return type, allowing customization
+    /// of how the success value is wrapped.
+    ///
+    /// - `result_ty`: The return type to handle.
+    /// - `lifetime_env`: The lifetime environment.
+    /// - `err_ffi_expr`: The FFI expression for the error value (e.g. "result.union.err").
+    /// - `fallback_action`: The Dart statement to execute on error fallback when there is no specific error type to convert and throw (e.g. "return null;" or "throw Object();").
+    /// - `wrap_success`: A closure that takes the success Dart expression and returns the wrapped statement.
+    fn gen_c_to_dart_for_return_type_with_post_process<F>(
```
tyler-zeromatter, line 1095: This is unrelated to the content of the PR itself, but I'm not necessarily a fan of the fact that these functions are formatting strings so heavily; it's less flexible and feels like it defeats the purpose of using askama.

I would try to move a lot of this templating logic into our askama templates at some point.

tyler-zeromatter (commented):
I think this implementation works, one small suggestion and one larger issue (unrelated to the content of this PR).

PR #1175, opened 2026-06-13
agent-driven: Fix Dart borrowing in structs by enabling struct_refs and fixing FFI passing

### LINKED ISSUE #1111: Dart Structs do not support references of opaques
The struct
```rs
    pub struct StructBorrowsOpaque<'a> {
        op : &'a Opaque,
    }
```

Fails to compile in Dart:
```dart
  _StructBorrowsOpaqueFfi _toFfi(ffi.Allocator temp, {core.List<core.List<Object>> aAppendArray = const []}) {
    final struct = ffi.Struct.create<_StructBorrowsOpaqueFfi>();
    struct.op = op._ffi;
    return struct;
  }
```
Dart warns that `aAppendArray` is never used.

Manishearth:
Fixes https://github.com/rust-diplomat/diplomat/issues/1111

Fully agent driven, haven't looked at this yet

- Enable `struct_refs` for Dart backend.
- Implement GC keeping-alive logic for opaque fields in structs: append them to lifetime `AppendArray`s in `_toFfi` to prevent premature GC.
- Fix FFI struct passing: pass borrowed structs (`&self`, `&mut self`, `&Struct`) by pointer instead of by value to match C FFI expectations.
- Enable `ImmutableStructOfOpaque` feature test for Dart.
- Add Dart test case to verify `ImmutableStructOfOpaque` functionality.


Manishearth:
@ambiguousname I'm not really as familiar with the struct ref code. I tried to review this and realized I need more context. If you feel like taking a closer look that would be great, otherwise I'll look later.

ambiguousname:
I'll take a look in a few minutes

Manishearth:
No rush! I figured I could see what would happen if I threw the AI at older bugs and the results were promising

feature_tests/src/structs.rs
```diff
@@ -601,7 +601,7 @@ pub mod ffi {
     }
 
     // FIXME: https://github.com/rust-diplomat/diplomat/issues/1111
-    #[diplomat::cfg(all(not(dart), supports=struct_refs))]
+    #[diplomat::cfg(supports=struct_refs)]
```
ambiguousname, line 618: Suggestions: If this now compiles, this FIXME can be removed (and possibly the FIXME above should be removed and tested as well)

tool/src/dart/mod.rs
```diff
@@ -844,7 +845,11 @@ impl<'cx> ItemGenContext<'_, 'cx> {
             SelfType::Struct(s) => {
                 let id = s.id();
                 let type_name = self.formatter.fmt_type_name(id);
-                format!("_{type_name}Ffi").into()
+                if s.owner().is_owned() {
+                    format!("_{type_name}Ffi").into()
+                } else {
+                    format!("ffi.Pointer<_{type_name}Ffi>").into()
```
ambiguousname, line 882: Makes sense, we're passing in a reference to a struct so we must pass in a pointer to that struct.

tool/src/dart/mod.rs
```diff
@@ -855,7 +860,16 @@ impl<'cx> ItemGenContext<'_, 'cx> {
...
             SelfType::Enum(ref e) if is_contiguous_enum(e.resolve(self.tcx)) => "index".into(),
-            SelfType::Struct(..) => format!("_toFfi({allocator})").into(),
+            SelfType::Struct(ref s) => {
+                let id = s.id();
+                let type_name = self.formatter.fmt_type_name(id);
+                let conversion = format!("_toFfi({allocator})");
+                if s.owner().is_owned() {
+                    conversion.into()
+                } else {
+                    format!("() {{ final ptr = {allocator}<_{type_name}Ffi>(); ptr.ref = {conversion}; return ptr; }}()").into()
```
ambiguousname, line 901: Also makes sense to me, like in JS and C++ we have to clone the struct to an FFI safe pointer.

This works for `struct_refs`, although the implementation will have to be slightly different for any `mut_struct_refs` support.

I can't comment on any dart specifics, but the rough logic seems sound to me.

tool/src/dart/mod.rs
```diff
@@ -881,8 +895,16 @@ impl<'cx> ItemGenContext<'_, 'cx> {
...
-                self.gen_dart_to_c_for_struct_type(dart_name, struct_borrow_info, alloc.unwrap())
+            Type::Struct(ref s) => {
+                let id = s.id();
+                let type_name = self.formatter.fmt_type_name(id);
+                let conversion = self.gen_dart_to_c_for_struct_type(dart_name, struct_borrow_info, alloc.unwrap());
+                if s.owner().is_owned() {
+                    conversion
+                } else {
+                    let alloc = alloc.unwrap();
+                    format!("() {{ final ptr = {alloc}<_{type_name}Ffi>(); ptr.ref = {conversion}; return ptr; }}()").into()
```
ambiguousname, line 941: Same as above, makes sense.

tool/templates/dart/struct.dart.jinja
```diff
@@ -77,6 +77,26 @@ final class {{type_name}}
     {%- endif %}
     struct.{{field.name}} = {{field.dart_to_c}};
     {%- endfor %}
+    {%- for l in lifetimes.all_lifetimes() %}
+    {%- let ltname = lifetimes.fmt_lifetime(l) %}
+    {%- for field in self::iter_fields_with_lifetimes_from_set(fields, l) %}
```
ambiguousname, line 82: I believe this is technically a separate fix (https://github.com/rust-diplomat/diplomat/issues/1111), but I think it's fine to include here.

I think the logic here roughly makes sense to me, we're filtering for fields that use a given lifetime `l` and adding them to that lifetime's array entries.

As a minor nit: I think the template code could be simplified into one for loop (wrapped by checks for if the field is optional and contains a value).

ambiguousname (approved):
Not sure about the dart specifics, but the struct ref logic roughly makes sense to me here.

PR #1177, opened 2026-06-17
AST Location errors second pass

tyler-zeromatter:
Continued progress on #111.

This addresses `ast/lifetimes.rs` and `ast/macros.rs`, with a few other files changed. I've also included some more advanced error features (like being able to point to different spans for additional context).

### DIFF
--- a/core/src/ast/types.rs
+++ b/core/src/ast/types.rs
@@ -12,9 +12,12 @@ use super::{
     OpaqueType, Path, RustLink, Struct, Trait,
 };
 use crate::{
-    Env, ast::{
-        Function, idents::{FromWithSpan, IntoWithSpan, SpanLocation}, logging::create_simple_report
-    }
+    ast::{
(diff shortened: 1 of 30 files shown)


core/src/ast/types.rs
```diff
@@ -1023,10 +1024,9 @@ impl TypeName {
                     if let syn::PathArguments::AngleBracketed(type_args) =
                         &p.path.segments.last().unwrap().arguments
                     {
-                        assert!(
-                            type_args.args.len() > 1,
-                            "Not enough arguments given to Result<T,E>. Are you using a non-std Result type?"
-                        );
+                        if type_args.args.len() != 2 {
```
tyler-zeromatter, line 1027: Wanted to check that changing this assert statement is okay. Is there ever a case where the number of generic args would be 3 or more?

Manishearth: Yeah

Manishearth (approved):

PR #1180, opened 2026-06-23
dotnet: GC.KeepAlive after native calls to fix mid-call finalization UAF

irvingoujAtDevolution:
## Problem

A generated opaque wrapper holds a raw `Raw.T*` and frees it via `Destroy` in `Dispose`/the finalizer. An instance call reads the pointer out, then calls native:

```csharp
var result = Raw.T.Method(AsFFI(), ...);
```

After `AsFFI()` the wrapper (`this`) is no longer referenced, so under optimized codegen the JIT marks it collectible at that point. The GC can then finalize it — running `Destroy`, freeing the Rust object — **while the native call is still in flight**. That's a use-after-free. It's the object-lifetime pitfall documented by Microsoft: [Best practices for native interop](https://learn.microsoft.com/dotnet/standard/unsafe-code/best-practices) ("Assumptions about object lifetimes (finalizers, GC.KeepAlive)").

## Fix

Emit `GC.KeepAlive(x)` after each native call that passes an opaque pointer — the receiver and/or any opaque parameters — keeping the wrapper rooted for the call's duration.

## Why not SafeHandle

SafeHandle prevents this **only when the SafeHandle itself is passed to the P/Invoke**, so the marshaller does `DangerousAddRef`/`DangerousRelease` around the call. This backend passes a bare `Raw.T*` (read via `DangerousGetHandle`), which the marshaller never sees — so a SafeHandle wrapper would add cost without engaging the mechanism that fixes the race. `GC.KeepAlive` is the correct, minimal fix here.

Verified empirically: with a full SafeHandle wrapper in place, removing `GC.KeepAlive` still reproduces the UAF (`drops = 1`).

## Test

**`GcRaceTests`** — runtime reproducer. A probe's native method sleeps and reports how many probes were finalized during the call; `>= 1` means the receiver was freed mid-call. RED without `GC.KeepAlive`, GREEN with it. The test project sets `<Optimize>true</Optimize>` so the race reproduces under the **default** `dotnet test` (the JIT only drops the receiver at last-use with optimized IL; a Debug build roots it and hides the bug).

### DIFF
--- a/example/dotnet/Generated/DataProvider.cs
+++ b/example/dotnet/Generated/DataProvider.cs
@@ -10,34 +10,7 @@ namespace Somelib;
 
 public partial class DataProvider: IDisposable
 {
-    /// <summary>
-    /// SafeHandle, not raw pointer + finalizer, for robust once-only release.
-    /// The native call takes a bare pointer the marshaller can't root, so the
-    /// generated <c>GC.KeepAlive(this)</c> — not this finalizer — is what stops
-    /// the GC freeing it mid-call (MS object-lifetime pitfall:
-    /// https://learn.microsoft.com/dotnet/standard/unsafe-code/best-practices).
-    /// No per-call <c>DangerousAddRef</c>: concurrent Dispose stays the caller's
-    /// problem, as with any <c>IDisposable</c>.
-    /// </summary>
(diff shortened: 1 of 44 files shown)


Manishearth (approved):
Seems fine.

PR #1181, opened 2026-06-23
dotnet: keep-alive edges for lifetime-carrying owned opaque returns

irvingoujAtDevolution:
> ⚠️ **DRAFT / experimental — do not merge.** Part 2 of a stacked pair, **stacked on #1180** (SafeHandle migration). Until #1180 merges, this PR's diff also contains #1180's commit; the net-new change here is the keep-alive edges. Under active review.

## What
When a method returns an owned `Box<Child<'a>>` that borrows from `self` or an opaque parameter, the generated child wrapper now stores strong references (`object[] _edges`) to the borrowed-from wrappers, so the GC cannot collect (and `Destroy`) a parent while the child is alive — mirroring the Dart/Kotlin/JS edge mechanism.

## Details
- Child still **owns** its own box (`ownsHandle: true`); edges only root the parents.
- `OpaqueParam` edges handled, including transitivity (e.g. `diamond_left` → `{ left, bottom }`).
- `SliceParam` and `StructLifetime` borrowing-returns, and **any non-opaque return that carries edges**, are rejected with a diagnostic (deferred / unsound to plumb today).
- Adds a constructible `BorrowParent`/`BorrowChild` fixture + a GC test (`Child_KeepsParent_AliveAcrossGc`) proving red→green.

## Residual (documented, not fixed)
Keep-alive edges defend against **GC collection** only. C# has no borrow checker, so explicitly `Dispose()`-ing (or `using`-scoping) a borrowed-from wrapper while the borrower is still in use frees the native memory early — a use-after-free this approach cannot statically prevent. Documented in the backend module doc.

## Verified
`cargo fmt`/`clippy` clean; 42/42 .NET feature tests (incl. the new GC test); dotnet unit tests pass.

### DIFF
--- a/example/dotnet/Generated/DataProvider.cs
+++ b/example/dotnet/Generated/DataProvider.cs
@@ -10,7 +10,40 @@ namespace Somelib;
 
 public partial class DataProvider: IDisposable
 {
-    private unsafe Raw.DataProvider* _inner;
+    /// <summary>
+    /// Owns the native <c>Raw.DataProvider*</c> handle. Deriving from
+    /// <c>SafeHandle</c> (instead of holding a raw pointer + a hand-written
+    /// finalizer) gives a once-only, thread-safe release and — through its
+    /// critical finalizer — prevents the GC from freeing the pointer while a
+    /// native call that reads it is still in flight.
+    /// </summary>
+    internal sealed unsafe class DataProviderHandle : SafeHandle
+    {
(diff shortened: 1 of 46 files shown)


irvingoujAtDevolution:
Superseded by #1186 — a clean reimplementation of keep-alive edges off the merged `main`. This branch was stacked on the old SafeHandle approach (since reverted in #1180 in favor of GC.KeepAlive), so it would reintroduce SafeHandle against current main. Closing in favor of #1186.

Issue #1166, opened 2026-06-08
Opaque Slices: Remove per-C-type View structs


tyler-zeromatter:
We support slices on structs *and* opaque types now, in addition to primitive types. Not every type is sliced, but currently the C/C++ backends can't know if an HIR type is in a slice unless we visit every function before visiting types.

It would be nice to have a solution that avoids having to generate all code in headers just in case.

tyler-zeromatter:
Of the many solutions, here are two possible ones:
1. Rework generation code to check which types are used in slices before generating those types
  - Pros: Re-usable types in headers for users. Opens the door to other kinds of generation where we might need to know how a type is used in other functions to generate its definition.
  - Cons: Quite a bit of work in re-writing generation. Might require two passes (which runs counter to the purpose of the HIR).
2. Use anonymous structs.
  - Pros: Fairly easy to implement.
  - Cons: C users would have to re-implement the struct themselves to use it. I also think C++ compilers complain about this, last time I experimented with this.

Manishearth:
> Cons: Quite a bit of work in re-writing generation. Might require two passes (which runs counter to the purpose of the HIR).


A thing you can do is have the generation code keep track of this and at the end of it make a single file with all of the objects. We used to do this for results, but it got ugly.

> Cons: C users would have to re-implement the struct themselves to use it.

What do you mean? Just the methods? We could define macros.

> I also think C++ compilers complain about this, last time I experimented with this.

Yeah, this is the thing I'd like to see some more investigation on.

But it's not a big deal.

tyler-zeromatter:
> What do you mean? Just the methods? We could define macros.

Yeah, for each anonymous struct you'd have to make your own definition. Macros would be useful in this case.

> Yeah, this is the thing I'd like to see some more investigation on.

I can research this again, I'll come back with any complaints I get from C++.

tyler-zeromatter:
For C/C++ compiler errors:

C gives a warning `anonymous struct declared inside parameter list will not be visible outside of this definition or declaration`

Both GCC and Clang do not support anonymous structs in functions for C++:
- Clang: https://github.com/llvm/llvm-project/blob/6c6338b3438704cddc1cf3bc6676316856b64907/clang/test/SemaCXX/type-definition-in-specifier.cpp#L37
[...]

PR #1182, opened 2026-06-23
Add Typing Usage Info Struct

tyler-zeromatter:
Part of the solution discussed in https://github.com/rust-diplomat/diplomat/issues/1166#issuecomment-4654211517

Once this gets approval, I'll add an implementation for selectively including slice structs based on the presence of slices which use that type.

### DIFF
--- a/core/src/hir/defs.rs
+++ b/core/src/hir/defs.rs
@@ -109,6 +109,7 @@ pub struct ResultUsageInfo<P: TyPosition> {
 
 #[derive(Debug, Clone)]
 #[non_exhaustive]
+// InputOnly allows for traits and callbacks, which creates a large byte difference (this type is > 900 bytes).
 #[allow(clippy::large_enum_variant)]
 pub enum ResultUsage {
     /// Result is used as the return of a callback:
--- a/core/src/hir/defs.rs
+++ b/core/src/hir/defs.rs
@@ -108,6 +108,40 @@ pub struct TypingUseInfo {
     pub optioned : bool
 }
 
+pub(super) trait TypeUsage {
+    fn set_usage(&mut self, usage : super::TypingUseInfo);
+}
+
+impl TypeUsage for StructDef<super::Everywhere> {
+    fn set_usage(&mut self, usage : TypingUseInfo) {
+        self.usage = usage;
+    }
+}
+
+impl TypeUsage for StructDef<super::OutputOnly> {
+    fn set_usage(&mut self, usage : TypingUseInfo) {
+        self.usage = usage;
+    }
+}
+
+impl TypeUsage for OpaqueDef {
+    fn set_usage(&mut self, usage : TypingUseInfo) {
+        self.usage = usage;
+    }
+}
+
+impl TypeUsage for EnumDef {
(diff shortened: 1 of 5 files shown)


core/src/hir/lowering.rs
```diff
@@ -206,6 +208,17 @@ impl<'ast> LoweringContext<'ast> {
         self.lower_all(ast_defs, Self::lower_function)
     }
 
+    pub(super) fn update_usage<'a, Ast: super::TypeUsage + 'a>(
```
Manishearth, line 216: nit: docs please

Manishearth (approved):

core/src/hir/defs.rs
```diff
@@ -109,6 +109,7 @@ pub struct ResultUsageInfo<P: TyPosition> {
 
 #[derive(Debug, Clone)]
 #[non_exhaustive]
+// InputOnly allows for traits and callbacks, which creates a large byte difference (this type is > 900 bytes).
 #[allow(clippy::large_enum_variant)]
 pub enum ResultUsage {
```
Manishearth, line 114: Given that this will be in a Vec we should box the ResultUsageInfos to avoid large copies

tyler-zeromatter: I'm wondering if there's a way to do this as a set, I will tinker

tyler-zeromatter: I really want to believe this is possible, but boy are callbacks making this more difficult than necessary.

tyler-zeromatter: The main issue is that callbacks can have attributes, and attributes touch a whole host of other things (which callbacks don't even use) that would need to be marked as `PartialEq`, `Eq`, and `Hash` for this to work properly. On top of that, I don't believe having a callback as the result of an `InputOnly` type is even possible. It's just a restriction of how `Type<InputOnly>` is set up.

It's theoretically possible to set up hashing/comparisons and panicking on Callbacks (since we don't expect them), but it would have to be localized to `ResultUsageInfo`, which is messy.

tyler-zeromatter: I'm wondering if I should just generate a unique ID for each callback on construction (so any callback will never be equivalent to anything other than itself). I think this might make sense since each callback is defined uniquely inside a parameter (although it will definitely need a disclaimer). Or maybe just skip attributes when hashing/comparing them?

I think being able to hash and compare types would be helpful for future devs, but I'm also just happy to abandon this for now if you'd like this in sooner than later @Manishearth

Manishearth: Yes, we should consider registering callbacks with an id instead.

But for now it's not a big deal.

tyler-zeromatter: I'll wrap the vec types in a `Box` for now and just make a separate draft PR to implement my `HashSet` changes instead, with callbacks having unique IDs (so each callback is unique to the parameter it is implemented at).

tyler-zeromatter: Filed an issue in https://github.com/rust-diplomat/diplomat/issues/1184, although I have a draft PR ready to go whenever this gets merged

PR #1186, opened 2026-06-25
dotnet: keep-alive edges for lifetime-carrying owned opaque returns

### LINKED ISSUE #1185: dotnet backend: lifetime soundness for lifetime-carrying owned opaque returns
## Problem

When a method returns an **owned opaque** whose Rust type carries a lifetime — `Box<Child<'a>>` that borrows from `self` or an opaque parameter — the generated C# `Child` wrapper holds a raw pointer into the parent's native memory but **no managed reference** to the parent wrapper. C# has no borrow checker, so:

- The GC can collect (and finalize → `Destroy`) the parent while a borrowing child is still alive → use-after-free.
- (C#-specific second hole: a caller can explicitly `Dispose()` the parent while the child is alive.)

The .NET backend originally emitted only an XML-doc warning for these returns, relying on caller discipline. This was flagged in the .NET backend PR (#1164, discussion on `tool/src/dotnet/gen/method.rs`); a follow-up was requested there.

## Resolution

Implement **keep-alive edges** (as the Dart/Kotlin/JS backends do): the returned wrapper retains strong references to the borrowed-from wrappers (computed from the HIR lifetime analysis) so the GC cannot collect a parent while a borrowing child is alive. Borrow shapes that can't be made sound this way — slice/string-borrowed returns, struct-lifetime borrows, fallible borrowing returns — are rejected with a diagnostic rather than emitted.

### Residual (accepted, documented)

Explicitly `Dispose()`-ing a parent while a borrowing child is in use is still a use-after-free and remains the caller's responsibility, as with most `IDisposable` types. Edges close the GC hole, not the explicit-disposal hole. Reference-counting outstanding borrows was considered and deliberately rejected (implicit, error-prone in many settings).

irvingoujAtDevolution:
Closes #1185.

## Problem

A method returning an owned opaque `Box<Child<'a>>` that borrows from `self` or an opaque parameter generates a C# `Child` wrapper holding a raw pointer into the parent's native memory, but **no managed reference** to the parent. The GC can then collect (and finalize → `Destroy`) the parent while a borrowing child is still alive — a use-after-free. (Previously only an XML-doc warning was emitted; flagged in #1164.)

## Fix — keep-alive edges

The returned wrapper now retains strong references to the borrowed-from wrappers in an `object[] _edges` field, computed from the HIR lifetime analysis (`borrowing_param_visitor` → `LifetimeEdgeKind`), so the GC cannot collect a parent while a borrowing child is alive. This is the same approach the Dart/Kotlin/JS backends use. Cleared on `Dispose`.

## Rejected, not silently emitted

Borrow shapes that can't be made sound this way produce a diagnostic instead of unsound code:
- returns borrowing from slice/string parameters (only pinned/converted for the call);
- struct-lifetime borrows (no struct edge plumbing yet);
- fallible borrowing returns (edges aren't threaded into the error/exception arm).

## Residual (documented, accepted)

Explicitly `Dispose()`-ing a parent while a borrowing child is in use is still a use-after-free and remains the caller's responsibility — as with most `IDisposable` types. Edges close the GC hole, not the explicit-disposal hole. Reference-counting outstanding borrows was considered and deliberately rejected (implicit, error-prone). Documented on the edge constructor.

## Tests

- `BorrowSoundnessTests.Child_KeepsParent_AliveAcrossGc` — a live child keeps its borrowed-from parent alive across a forced GC. Deterministic and non-crashing: it asserts via `WeakReference` *before* dereferencing the child. RED without edges, GREEN with.
- `ViewFrom_ParamNamedThis_KeepsParentAlive` — regression for a Rust parameter literally named `this` (the edge must render `@this`, not the receiver keyword `this`, which wouldn't compile in the generated static method).

The fixture (`BorrowParent`/`BorrowChild`) is `#[diplomat::attr(not(dotnet), disable)]` to avoid churning the other backends' generated output.


tool/src/dotnet/gen/method.rs
```diff
@@ -712,50 +755,74 @@ impl<'ctx, 'tcx> ItemGenContext<'ctx, 'tcx> {
...
             lifetime_warning,
+            keep_alive_edges,
             error_info,
             option_info,
             property_accessor,
         })
     }
 
-    fn borrowed_output_lifetime_warning(&self, method: &'tcx Method) -> Option<bool> {
+    fn borrowed_output_keep_alive_edges(&self, method: &'tcx Method) -> Option<Vec<String>> {
```
Manishearth, line 765: nit (future): Please add docs for this function at some point, explaining what it does.

But you might need to restructure this to support structs and slices anyway

Manishearth (approved):

PR #1189, opened 2026-06-25
dotnet: borrowed opaque returns via a non-owning wrapper

### LINKED ISSUE #1187: dotnet backend: flat keep-alive edges can't represent non-owning (borrowed) opaque returns
Follow-up from #1186 (keep-alive edges), recording a design limit for future work.

## Context

The .NET backend stores keep-alive edges as a single flat `object[] _edges` on the returned wrapper, vs. the per-lifetime edge fields the Dart/Kotlin/JS backends use. For the shapes .NET currently supports — an **owned** opaque return (`Box<Child<'a>>`) that borrows from `self` or an opaque parameter — this is functionally equivalent: the array just roots the borrowed-from wrappers so the GC can't collect (and finalize → `Destroy`) them.

## The limit

.NET currently **rejects** borrowed opaque returns (`&T`, `&mut T`, `Option<&T>`). Dart/Kotlin support them, and their per-lifetime `selfEdge` does double duty: a populated self-edge also marks the returned wrapper as **non-owning** (`owned = false` / don't attach the finalizer), so the wrapper never frees a pointer it doesn't own.

The flat `_edges` model carries **no ownership signal**. So if/when .NET lifts the borrowed-opaque-return rejection, the flat edges won't be enough — it will also need:

- a non-owning wrapper variant (an `owned` flag, or simply not running `Destroy` for borrowed returns), and
- possibly per-lifetime edge tracking, if struct-lifetime borrows or borrowed-return threading are added at the same time.

## Not a current bug

Today the borrowed-return path is closed (rejected at codegen with a diagnostic), so the flat model is sound for everything .NET accepts. This issue just records the constraint so the borrowed-return work doesn't naively reuse the flat model.

irvingoujAtDevolution:
Closes #1187.

## What

Borrowed opaque returns — `&T` / `&mut T` / `Option<&T>` — in the .NET backend, via a non-owning wrapper.

## How

The raw handle lives in a `RustHandle<T>` that carries the Rust destructor for **owned** returns and **null for borrowed** ones — so `Dispose`/the finalizer free the pointer only when C# owns it (no separate `_owned` flag). Borrowed returns are built with `RustHandle.Borrowed(...)`; the keep-alive `_edges` (from #1186) root the borrowed-from owner so the borrow stays valid. A pre-pass (`borrowed_return_targets`) marks every opaque ever returned `&T`, so it gets the non-owning path even when it carries no lifetime of its own. Owned returns are unchanged.

## Tests

New `BorrowedReturnTests` cover the runtime contract end-to-end:
- the borrowed value is readable;
- interior aliasing — mutating the owner is observed through an outstanding borrow (proves it's a reference, not a copy);
- a borrowed `Dispose` is a no-op (no double-free);
- `_edges` keeps the owner alive across a forced GC.

Also enables the `get` indexer, which now lowers as a plain `Get(...)` method.

## Deferred — rejected with a clear diagnostic, not silently mis-compiled

- Fallible borrowing returns (`Result<&T, E>`) — needs per-arm edge analysis.
- Borrowing from slice/string params — those buffers are only pinned for the call, so there's no edge to root yet.

### DIFF
--- a/example/dotnet/Generated/RustHandle.cs
+++ b/example/dotnet/Generated/RustHandle.cs
@@ -1,0 +1,50 @@
+using System;
+
+namespace Somelib.Diplomat;
+
+#nullable enable
+
+/// <summary>
+/// Frees a Rust-owned <typeparamref name="T"/> by calling its native
+/// destructor. Held by an owned <see cref="RustHandle{T}"/> so the handle can
+/// release the pointer without knowing the concrete type.
+/// </summary>
+internal unsafe delegate void RustDestructor<T>(T* ptr) where T : unmanaged;
+
+/// <summary>
+/// A raw pointer that remembers who's responsible for freeing it. An owned
+/// handle carries the Rust destructor and runs it when you release the handle;
+/// a borrowed handle carries none, so releasing it does nothing — Rust still
+/// owns the memory and will free it itself. That's the whole trick: the
+/// pointer itself says "free me" or "leave me alone", so the wrapper doesn't
+/// need a separate flag bolted onto the class.
+/// </summary>
+internal readonly unsafe struct RustHandle<T> where T : unmanaged
+{
+    private readonly T* _ptr;
+    private readonly RustDestructor<T>? _free;
+
+    private RustHandle(T* ptr, RustDestructor<T>? free)
+    {
+        _ptr = ptr;
+        _free = free;
+    }
+
+    /// <summary>The C# side owns the pointer: <see cref="Release"/> frees it.</summary>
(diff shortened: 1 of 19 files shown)


Manishearth:
> Fallible borrowing returns (Result<&T, E>) — needs per-arm edge analysis.


Look at what the other backends do, i think they structure their FromFFI code to require edge lists and those lists get constructed in the arms.

But yes, this complicates things, good call to handle later.

Manishearth (approved):
Hmm, I think this is fine, but you probably want to use the preexisting `borrowing_param` infrastructure for tracking ownership and borrowing in the long run.

https://docs.rs/diplomat_core/latest/diplomat_core/hir/borrowing_param/index.html

PR #1190, opened 2026-06-26
Add nanobind documentation

twantonie:


Manishearth (approved):

twantonie:
I'll try and see if I can fix the pipeline issues after the weekend

tyler-zeromatter:
Taking a look now

feature_tests/nanobind/test/test_opaque_zsts.py
```diff
@@ -43,7 +43,7 @@ def test_success_result(self):
     
     def test_fail_result(self):
         c = somelib.ns.RenamedOpaqueZST.fail_zst(True)
-        assert(c)
+        assert c is None, "A unit Ok value is represented as None"
```
tyler-zeromatter, line 46: Nit: Not sure why this needs to be changed? But I guess this is fine.

Unrelated, but a test of `__nb_signature__` (or wherever Nanobind stores the docstrings) would be useful.

feature_tests/nanobind/pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["scikit-build-core >=0.4.3", "nanobind >=2.0.0"]
+requires = ["scikit-build-core >=0.4.3", "nanobind ==2.4.0"]
```
tyler-zeromatter, line 2: Why version 2.4.0 specifically?

feature_tests/nanobind/CMakeLists.txt
```diff
@@ -72,5 +72,24 @@ target_include_directories(somelib PUBLIC "src/include" "${nanobind_ROOT}/ext/ro
...
+  DEPENDS somelib
+  MARKER_FILE py.typed
+)
+
 # Install directive for scikit-build-core
-install(TARGETS somelib LIBRARY DESTINATION somelib)
\ No newline at end of file
+install(TARGETS somelib LIBRARY DESTINATION somelib)
+install(FILES
+  ${CMAKE_CURRENT_BINARY_DIR}/somelib.pyi
```
tyler-zeromatter: Suggestion: I'm thinking it might be nice to track the generated `.pyi` files on GitHub so CI can check for type changes in the stubs.

tool/src/nanobind/formatter.rs
```diff
@@ -25,9 +29,50 @@ impl<'tcx> PyFormatter<'tcx> {
...
+        let mut docs = docs
+            .to_markdown(DocsTypeReferenceSyntax::SquareBrackets, self.docs_url_gen)
+            .trim()
+            .to_string();
+        if let Some(deprecated) = attrs.deprecated.as_ref() {
+            if !docs.is_empty() {
+                docs.push('\n');
+                docs.push('\n');
+            }
+            let _ = writeln!(&mut docs, ".. deprecated:: {deprecated}");
```
tyler-zeromatter, line 47: It would be nice to be able to mark types with `@warnings.deprecated` in the future, but this is fine for now.

tyler-zeromatter (commented):
Looks good to me with a few minor nits/suggestions.

Ideally I'd like to understand why this PR locks to Nanobind 2.4.0 before I give approval.

feature_tests/nanobind/pyproject.toml
twantonie: That is the version that is fetched by the [CI](https://github.com/rust-diplomat/diplomat/blob/main/Makefile.toml#L665), so that is one that is known to work. 

I think I might be able to properly fix the nanobind hack that references internal header files, but that would also rework the C++ generation as well. The idea would be to instead of having unique_ptr's for the classes, you would have an internal pointer that behaves very similar to a unique pointer. But that's another pull request of it's own.

feature_tests/nanobind/test/test_opaque_zsts.py
twantonie: Previously this function returned monostate which is a value, now it returns None which is not a value and the assertion would otherwise fail

feature_tests/nanobind/CMakeLists.txt
twantonie: I agree that it would be nice (and I tried doing this), but it's quite the headache since you need to build the python module in the gen-nanobind-feature step

feature_tests/nanobind/pyproject.toml
twantonie: After some more research I don't think I can implement the above idea. This is mainly due to the fact that you allow references to opaque types, which is a dangerous footgun to users of this library.

twantonie:
Fixed up line endings, so hopefully the pipeline will pass now

feature_tests/nanobind/CMakeLists.txt
tyler-zeromatter: That can be a future improvement, then.

twantonie: If the pipeline passes, the generated .pyi files will be tracked. It just adds quite a bit of additional compilation, but if that doesn't worry you I'll leave it in

tyler-zeromatter: Sounds good. I don't think we're worried about additional compilation in the CI right now, but I'll let @Manishearth weigh in on that.

tyler-zeromatter (approved):
LGTM!

I think at some point we should try to upgrade our Nanobind version so we can use `nanobind_add_stub` and avoid all this custom tooling, but that can be a separate PR.

PR #1194, opened 2026-06-29
dotnet: remove borrowed_return_targets pre-pass; all opaques unconditionally use RustHandle

irvingoujAtDevolution:
Follow-up to #1189, prompted by @Manishearth's review.

As per the Rust/Postgres (not related here, it's just I like it) community's long-standing correctness-over-performance practice, I replaced the `borrowed_return_targets()` global pre-pass — which scanned every type and method in the TCX to decide which opaques needed the non-owning-wrapper machinery — with the simpler invariant: **all opaque types always use `RustHandle<T>` + `_edges`**.

The pre-pass was fragile: it only caught direct `&T` returns, silently missed `Result<&T, E>` and any future borrowed-return shape, and was a parallel tracking mechanism sitting alongside the existing per-method `borrowing_param` visitor. The two mechanisms answered different questions about the same domain.

Cost: 8 bytes per instance (the `_edges` field; zero heap allocation when unused — `System.Array.Empty<object>()` is a shared global), and one extra indirection through `_destroy` per Dispose call. The `has_edges` conditional in the template is gone entirely, so the generated shape is now uniform across all opaques.

### DIFF
--- a/example/dotnet/Generated/DataProvider.cs
+++ b/example/dotnet/Generated/DataProvider.cs
@@ -10,7 +10,15 @@ namespace Somelib;
 
 public partial class DataProvider: IDisposable
 {
-    private unsafe Raw.DataProvider* _inner;
+    private unsafe RustHandle<Raw.DataProvider> _inner;
+
+    /// <summary>
+    /// Roots the wrappers this value borrows from so the GC cannot finalize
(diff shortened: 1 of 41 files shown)


Manishearth (approved):

Manishearth:
Yeah this makes more sense.

PR #1199, opened 2026-07-01
Separate out Nanobind stubgen task

tyler-zeromatter:
It's annoying to rebuild in CMake every time the files change when I'm iterating on generation, so this task separates out generating the stubs from the bindings to make iteration faster.

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -682,9 +682,29 @@ script = '''
 mkdir vendor
 cd vendor
 nanobind_exists = is_path_exists nanobind
-if not nanobind_exists
+if not ${nanobind_exists}
     exec git clone https://github.com/wjakob/nanobind.git --depth=1 -b "v2.4.0" --recurse-submodules -q
 end
+# Read Nanobind version:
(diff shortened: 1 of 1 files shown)


tyler-zeromatter:
Just added some checks to verify the Nanobind version and a fix for the error message complaining about Nanobind not existing when it does

Makefile.toml
```diff
@@ -581,23 +581,39 @@ generate_generic example kotlin somelib "--config kotlin.scaffold=false" /src/ma
 
 [tasks.gen-nanobind-feature]
 category = "Code generation"
+dependencies = ["gen-nanobind-bindings", "gen-nanobind-stubs"]
```
Manishearth: Won't this run the two in parallel? That would be flaky, right?

You can have bindings be a dependency of stubgen or invoke cargo make here I think.

tyler-zeromatter: Ah, yeah I think there's a better way to organize this. Let me re-work

Manishearth (approved):

PR #1201, opened 2026-07-02
dotnet: owned opaque returns borrowing slice params via pinned ReadOnlyMemory

irvingoujAtDevolution:
## Motivation

Zero-copy parse APIs of the shape

```rust
pub fn decode(data: &'a [u8]) -> Result<Box<SecurityList<'a>>, Box<Error>>
```

were rejected by the .NET backend: slice params are only pinned via `fixed (...)` for the
duration of the call, so a returned value that borrows from one would dangle once the GC
moved or collected the backing array.

## Design

When `BorrowingParamVisitor` classifies a `&[u8]`/`&[u32]` param as `BorrowedSlice` (its
lifetime is used by an **owned** opaque success return), the lowering changes:

- The idiomatic param becomes `ReadOnlyMemory<byte>` / `ReadOnlyMemory<uint>` (arrays convert
  implicitly, so naive call sites are unchanged). This communicates the retention and lets
  performance-sensitive callers avoid GC pinning entirely (POH arrays, native
  `MemoryManager`-backed memory, sub-ranges via `AsMemory(offset, len)`).
- The buffer is pinned into a new `DiplomatPinnedMemory` holder (`ReadOnlyMemory.Pin()` →
  `MemoryHandle`) and rooted on the returned wrapper via the existing keep-alive `_edges`
  mechanism.
- `Dispose()` unpins holder edges **strictly after** `_inner.Release()` — Rust's `Drop` may
  still read the borrowed buffer. The holder deliberately has **no finalizer**, so it can
  never unpin ahead of the owning wrapper's destroy in the same GC cycle.
- Exception-safe throughout: the pin is created inside a `try`; a `catch` disposes each pin
  and rethrows (so a failed P/Invoke or a partial multi-pin failure can't leak a pin), and
  `DiplomatPinnedMemory.Pin` disposes the handle if wrapping it throws.

Generated shape:

```csharp
public static Parsed Parse(ReadOnlyMemory<byte> data)
{
    unsafe
    {
        DiplomatPinnedMemory? dataPin = null;
        try
        {
            dataPin = DiplomatPinnedMemory.Pin(data);
            var result = Raw.Parsed.Parse(new DiplomatSliceU8 { Ptr = (byte*)dataPin.Pointer, Len = (nuint)data.Length });
            if (!result.IsOk)
            {
                throw new ParseErrorException(new ParseError(result.Err));
            }
            return new Parsed(result.Ok, new object[] { dataPin });
        }
        catch
        {
            dataPin?.Dispose();
            throw;
        }
    }
}
```

## Conditional emission (netstandard2.0 floor)

`DiplomatPinnedMemory` depends on `System.Buffers.MemoryHandle` / `ReadOnlyMemory<T>`, which
need the `System.Memory` package on the netstandard2.0 / .NET Framework floor the backend
supports. A **run-level** flag emits the helper and the `Dispose` unpin-sweep only when the
generation actually produces a pinned-slice edge, so projects that never use the feature
don't inherit the dependency. The flag is computed before rendering (a pin edge lands on the
*returned* type, which may render before the pinning method's declaring type), and when set
the sweep is applied uniformly to every opaque — this does **not** reintroduce the per-type
pre-pass #1194 removed.

## Rejected (with diagnostics)

- borrowed (non-owned) opaque returns that borrow a slice — the wrapper never runs the Rust
  destructor, so unpinning on its `Dispose` would race the still-live Rust object
- borrowed string params (`&'a str`)
- error-arm slice borrows (a thrown exception has no owner to unpin)
- `Option`-wrapped pinned returns (a `null` result would leak the pin)
- mutable borrowed slices (`ReadOnlyMemory` can't hand Rust a mutable view)

Temporary (non-borrowed) slices keep the cheap call-scoped `fixed` lowering unchanged.

## Caller contract

The generated XML remarks state that the buffer stays pinned until the returned value is
disposed and must not be mutated while it is in use (Rust holds `&'a [u8]` and assumes
immutability — the same implicit contract the C/C++ bindings already carry).

## Tests

Tool unit tests cover the fallible/infallible decode shapes, unpin-on-Dispose ordering
(`Release()` before unpin), cross-type pinned returns (a factory on one type returning a
pinned instance of another), multi-pin methods, `ReadOnlyMemory<uint>`, and the rejection
diagnostics. Runtime tests (`PinnedSliceTests.cs`) cover round-trip, GC-pressure survival
while pinned, sub-range `AsMemory`, dispose semantics, and disposing a **live** pinned handle
on the throw path. Feature-test and example bindings regenerated.


tool/templates/dotnet/DiplomatPinnedMemory.cs.jinja
```diff
@@ -0,0 +1,61 @@
...
+/// </summary>
+/// <remarks>
+/// This uses <c>System.ReadOnlyMemory&lt;T&gt;</c> and
+/// <c>System.Buffers.MemoryHandle</c>. Those are built in on .NET Core 2.1+ /
+/// .NET Standard 2.1+, but on .NET Standard 2.0 and .NET Framework they ship in
+/// the <c>System.Memory</c> NuGet package — if you target those and use a
+/// method that borrows a slice, add a reference to <c>System.Memory</c> or this
+/// file will not compile.
+/// </remarks>
+internal sealed unsafe class DiplomatPinnedMemory : IDisposable
```
Manishearth, line 24: I recommend looking at how Kotlin does this for slices

Manishearth (approved):
This is fine, but I really really recommend trying to reuse the existing Diplomat scaffolding for borrowing. Look at the Kotlin, Dart, and JS backends. The scaffolding already has special understanding of how slices work.

Kotlin also has nice library abstractions around this.

tool/templates/dotnet/DiplomatPinnedMemory.cs.jinja
irvingoujAtDevolution: Hi @Manishearth, thanks for the review. Want to make sure I read the suggestion right — from the codebase, Kotlin copies the slice into a fresh native allocation
```
fun borrow(byteArray: ByteArray): GCSlice {
    val mem = allocateGarbageCollectedMemory(byteArray.size.toLong())
    val slice = copy(byteArray, mem)
    return GCSlice(mem, slice)
}
```
I'm not a JVM expert, but my understanding is: mem is a malloc'd block outside the JVM heap, the array is copied into it, and it's freed through GC reachability — the Memory object rides in the returned wrapper's edge list until both are unreachable. That copy looks like a tradeoff forced by the JVM (JNA docs: arrays are only valid within a single call), whereas C/C++ always pass the reference through without copying.

Since .NET specifically supports long-lived pinning (Memory<T>.Pin()), I think it makes more sense to preserve the reference semantic here. The edge model is the same as Kotlin's — the pin holder sits in the wrapper's edges the way GCSlice's Memory does — just with a pin inside instead of a copy.

It is also performance critical for our usage, https://github.com/Devolutions/IronRDP
where we try to limit allocation as much as possible.

Manishearth: Right, I was more talking about how the `.borrow()` method is on the arena type and that gets passed around and used in edges. It's not a code behavior thing, it's a code structure thing, it should not affect perf.

But perhaps .NET doesn't need the edge arena type. I don't fully understand how `Memory.Pin()` avoids a copy here: Does .NET allow native code to directly access buffers? Even for strings?

JVM/Dart/JS all do copies here because at least for some array types their internal representation is opaque to native code.

irvingoujAtDevolution:
Yes — .NET lets native code access managed buffers directly. Primitive arrays are [blittable](https://learn.microsoft.com/en-us/dotnet/framework/interop/blittable-and-non-blittable-types): same representation in managed and unmanaged memory, to the point that the default P/Invoke marshaller pins them instead of copying. [`ReadOnlyMemory<T>.Pin()`](https://learn.microsoft.com/en-us/dotnet/api/system.readonlymemory-1.pin) makes that stable past the call — "the garbage collector will not move the memory until the returned MemoryHandle is disposed". So for `&[u8]`/`&[u32]` the pointer Rust gets is the caller's actual array, no copy.

Strings too, funnily enough — the marshaller ["passes the callee a direct pointer to managed strings in the internal Unicode buffer"](https://learn.microsoft.com/en-us/dotnet/framework/interop/copying-and-pinning) instead of copying. The copy we do for strings is just the UTF-16→UTF-8 transcode the bridge needs, not a GC thing.

Structure-wise I think we ended up close to Kotlin already: `DiplomatPinnedMemory.Pin(data)` plays the `.borrow()` role and the holder is what lands in the wrapper's edges — it just holds a pin handle instead of a malloc+copy. We still need the holder type since something has to own the unpin and run it after the Rust dtor.

tool/templates/dotnet/DiplomatPinnedMemory.cs.jinja
irvingoujAtDevolution: please see https://github.com/rust-diplomat/diplomat/pull/1201#issuecomment-4879002757

Manishearth:
> The copy we do for strings is just the UTF-16→UTF-8 transcode the bridge needs, not a GC thing.


Yes, and if you want to hang on to a borrow of this string you need to pin the copy for the GC.

The Kotlin copy isn't a GC thing either: it's because you need to transcode things.

> Structure-wise I think we ended up close to Kotlin already: DiplomatPinnedMemory.Pin(data) plays the .borrow() role and the holder is what lands in the wrapper's edges — it just holds a pin handle instead of a malloc+copy. We still need the holder type since something has to own the unpin and run it after the Rust dtor.

That makes sense, thanks!

PR #1207, opened 2026-07-06
dotnet: render stringifiers as ToString() overrides

irvingoujAtDevolution:
Flips `stringifiers` on for .NET. A `#[diplomat::attr(auto, stringifier)]` method now renders as `public override string ToString()` on the idiomatic wrapper instead of a plain named method — raw layer unchanged. The old name-based stopgap (any parameterless write method literally named `to_string`) runs through the same predicate now, so there's one mechanism, not two.

Shapes C# can't express fail loud instead of emitting broken code: static stringifiers, fallible/nullable ones, and two methods both claiming the `ToString()` override all report a diagnostic and abort generation.

Regen produced zero diff — both feature-test stringifiers were already named `to_string`, so the output is byte-identical, just attr-driven now.

Tests: 4 new tool unit tests (render/replace + all three rejections), 2 new xunit facts including virtual dispatch through an `object`-typed reference.

### DIFF
--- a/feature_tests/dotnet/Tests/StringifierTests.cs
+++ b/feature_tests/dotnet/Tests/StringifierTests.cs
@@ -1,0 +1,31 @@
+using Somelib;
+using Xunit;
+
+namespace Somelib.FeatureTests;
+
+public class StringifierTests
+{
+    [Fact]
+    public void Stringifier_OverridesObjectToString()
+    {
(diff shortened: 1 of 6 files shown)


irvingoujAtDevolution:
closing the draft — stringifiers/`ToString()` isn't something we're pushing right now. can reopen later if we pick it back up.

Issue #228, opened 2022-08-24
Make `hir::LoweringError` have more specific variants


QnnOkabayashi:
When I wrote the HIR lowering code, I didn't know every kind of error that could occur, so the `hir::LoweringError` enum only has an `Other(String)` variant. Once the lowering is in a more finalized state, we should ctrl-F through the lowering code and add variants for each kind of error that can occur.

Manishearth:
Also potentially try to use something like annotate-snippets or miette for errors

PR #1209, opened 2026-07-07
HIR location errors first pass

tyler-zeromatter:
This adds a `LoweringError::OtherWithLoc` type for `LoweringError`, which takes in an `AstReport` to show where the error happens in the HIR.

TODO:
- [ ] Add one test for location errors in the HIR. This doesn't need to be as comprehensive as the AST location error tests, but at least one will show that lowering errors can print with locations.


tyler-zeromatter:
Closing this because I'd like to also address https://github.com/rust-diplomat/diplomat/issues/228 in the same stroke; otherwise I think trying to combine reporting information in the `LoweringError` enum with later variants will be tricky.

PR #1215, opened 2026-07-09
dotnet: owned Box<[u8]> return via zero-copy RustVec

irvingoujAtDevolution:
Adds .NET support for returning `Box<[u8]>` without an implicit copy.

`RustVec` owns the native allocation, works with `using`, provides scoped zero-copy access through `WithSpan(...)`, and makes copies explicit with `Clone()`. Raw P/Invoke stays in the raw layer; the public wrapper is safe.

Tests: `cargo fmt --all --check`, Clippy with `-D warnings`, C# formatter/analyzers, and CI.


core/src/hir/lowering.rs
```diff
@@ -1563,6 +1563,29 @@ impl<'ast> LoweringContext<'ast> {
...
+            // guaranteed layout, unlike the `DiplomatOwnedSlice<u8>` repr(C)
+            // shape an infallible return is converted to.
+            ast::TypeName::PrimitiveSlice(None, prim, _stdlib)
+                if self
+                    .attr_validator
+                    .attrs_supported()
+                    .owned_byte_slice_returns
+                    && context == TypeLoweringContext::Method
+                    && !in_result_option
+                    && matches!(
```
tyler-zeromatter: I would migrate the `self.attr_validator.attrs_supported().owned_byte_slice_returns` check into the block so we can return a specific `LoweringError` to tell users to `#[diplomat::cfg(supports=owned_byte_slice_returns)]`.

This avoids a generic "Owned slices cannot be returned" error (which is now not helpful in this scenario.

tyler-zeromatter: Additionally, I would update the "Owned slices cannot be returned" error to specify that some backends can now return `Box<[u8]>`.

tyler-zeromatter (commented):
Mostly looks good to me, the only nit I have is that I would like to see the error messages updated in `lowering.rs` to be more specific now that some backends can return `Box<[u8]>`.

I gave the C# implementation a quick glance and it seems to be fine, but it's not really my backend so I'm trusting the backend authors here that it is correct.

core/src/hir/lowering.rs
irvingoujAtDevolution: Thanks, updated!

tyler-zeromatter:
Doing a second review pass now

tyler-zeromatter (approved):
LGTM!

I think the .NET backend may want to eventually switch to using [insta](https://insta.rs/) for its lowering error tests, like the rest of the backends, but that's beyond the scope of this PR.

tyler-zeromatter:
@Manishearth unless you want to take a look, I'm happy to merge

Manishearth:
I'm still on vacation so go ahead

PR #1217, opened 2026-07-10
feat: `diplomat-runtime` - Add `no_std` support for `wasm32`

### LINKED ISSUE #1212: Support `no_std` builds for `wasm32-unknown-unknown`
Just documenting this while working on another issue. No clue if there is any value or interest in resolving it :)

After adding the `#[cfg(debug_assertions)]` to the `panic_handler()` method and setting `#![no_std]` in the `lib.rs`, the runtime crate compiled successfully for `no_std`.

I don't think a `std` feature is necessary here since the only relevant context for the current gating against `no_std` is that `panic_handler()` method which is gated behind debug? (_and the module itself states the `wasm_glue.rs` is intended as private_) 🤷‍♂️

**History:**
- July 2021 - [Support building the diplomat runtime without requiring `std`](https://github.com/rust-diplomat/diplomat/pull/37) (except for `wasm32`):

  https://github.com/rust-diplomat/diplomat/blob/f53af7e6554e7ff4e7053ce1ecec2ae228d61010/runtime/src/lib.rs#L1

- Oct 2025 - [Avoid `wasm_glue` for WASI targets](https://github.com/rust-diplomat/diplomat/pull/977):

  https://github.com/rust-diplomat/diplomat/blob/f53af7e6554e7ff4e7053ce1ecec2ae228d61010/runtime/src/lib.rs#L25-L26

---

The only remaining blockers for building `wasm32-unknown-unknown` seem to be due to the following:

There is a `panic_handler()` added here via a debug assertions constraint:

https://github.com/rust-diplomat/diplomat/blob/f53af7e6554e7ff4e7053ce1ecec2ae228d61010/runtime/src/wasm_glue.rs#L23-L33

That constraint isn't carried over here on the method however:

https://github.com/rust-diplomat/diplomat/blob/f53af7e6554e7ff4e7053ce1ecec2ae228d61010/runtime/src/wasm_glue.rs#L35-L55

Where `String` is not `alloc::string::String`, and the method parameter is `std::panic::PanicHookInfo` (_`core::panic::PanicInfo` was a apparently a compatible type [prior to Rust 1.81 changes](https://github.com/rust-lang/rust/pull/115974) but instead now [expects `std::panic::PanicHookInfo`](https://github.com/rust-diplomat/diplomat/pull/501)_).

---

A panic handler and global memory allocator would still need to be provided for such a build, but `diplomat` doesn't necessarily have to be responsible for that?

For example the `icu_capi` crate at ICU4X (which uses `diplomat` as a dependency) provides [this support via opt-in features](https://github.com/unicode-org/icu4x/blob/d0e721bf145e38953df94d2d22098d6e9f30c423/ffi/capi/src/lib.rs#L57-L65):

**`src/lib.rs`:**
```rust
#[cfg(all(not(feature = "std"), feature = "looping_panic_handler"))]
#[panic_handler]
fn panic(_info: &core::panic::PanicInfo) -> ! {
    loop {}
}

#[cfg(all(not(feature = "std"), feature = "libc_alloc"))]
extern crate libc_alloc;
```

**`Cargo.toml`:**
```toml
[features]
# Adds a looping panic handler (useful for no_std targets)
looping_panic_handler = []
# Uses malloc as the global allocator (useful for no_std targets)
libc_alloc = ["dep:libc_alloc"]

[dependencies]
[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
# Logging uses diplomat_runtime bindings in wasm, we only need this for native
simple_logger = { workspace = true, optional = true }
libc_alloc = { workspace = true, features = ["global"], optional = true }
```

In that case the allocator wouldn't be sufficient, and one that is compatible with `no_std` in general would be needed instead (_doesn't add much more complexity_).

polarathene:
Presently the `diplomat-runtime` crate is `no_std` for all targets other than `wasm32-unknown-unknown`. This PR should keep the existing functionality during debug builds, but enable support for building without `std` on release builds.

This is only required for the `wasm_glur.rs` module, which itself feature gates the module functionality behind a `log` feature (only requires `alloc`) and the remainder behind a `debug_assertions` constraint for a panic handler hook (that requires `std`).

Resolves: https://github.com/rust-diplomat/diplomat/issues/1212

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -1,4 +1,4 @@
-#![cfg_attr(not(any(target_arch = "wasm32")), no_std)]
+#![no_std]
 
 //! The [Diplomat](https://rust-diplomat.github.io/diplomat/) runtime crate.
 //!
--- a/runtime/src/wasm_glue.rs
+++ b/runtime/src/wasm_glue.rs
@@ -6,7 +6,7 @@
 #[cfg(debug_assertions)]
 extern crate std;
 #[cfg(debug_assertions)]
-use alloc::{ boxed::Box, string::String };
+use alloc::{boxed::Box, string::String};
 
 // https://blog.rust-lang.org/2026/04/04/changes-to-webassembly-targets-and-handling-undefined-symbols/#what-is-going-to-break-and-how-to-fix
 #[link(wasm_import_module = "env")]
--- a/runtime/src/wasm_glue.rs
+++ b/runtime/src/wasm_glue.rs
@@ -1,11 +1,17 @@
 //! Wasm-specific glue code
 //!
 //! This is a private module
-use alloc::format;
+
+// `std` is only required to support a `panic_handler()` hook in debug builds:
+#[cfg(debug_assertions)]
+extern crate std;
+#[cfg(debug_assertions)]
+use alloc::{ boxed::Box, string::String };
 
 // https://blog.rust-lang.org/2026/04/04/changes-to-webassembly-targets-and-handling-undefined-symbols/#what-is-going-to-break-and-how-to-fix
 #[link(wasm_import_module = "env")]
 extern "C" {
+    #[cfg(debug_assertions)]
     fn diplomat_throw_error_js(ptr: *const u8, len: usize);
 
     #[cfg(feature = "log")]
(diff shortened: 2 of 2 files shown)


Manishearth (approved):
Looks like the code does not do anything without `debug_assertions`, so yeah this makes sense.

Manishearth:
I'm surprised, I thought we unconditionally handled panics on wasm, but I guess we don't.

PR #1220, opened 2026-07-15
Add `Error` and `Exception` to enums/classes which are errors/exceptions in Dart

mosuem:
While Dart allows to throw any `Object`, this is a convention. Also it allows for more fine grained control on what to catch.


core/src/hir/attrs.rs
```diff
@@ -69,6 +69,11 @@ pub struct Attrs {
     /// For #[diplomat::attr()]. If true, the struct should be treated as a tuple in function arguments and return values by the language.
     pub tuple: bool,
 
+    /// From #[diplomat::attr()]. Mark an error type as an unrecoverable bug/precondition failure (Error in Dart/panic).
+    pub bug: bool,
```
robertbastian: I'm not convinced that Dart's `Error`/`Exception` architecture generalises like this. 

We already have `attr(*, error)`, do we really need three different attributes for this?

core/src/hir/attrs.rs
```diff
@@ -69,6 +69,11 @@ pub struct Attrs {
...
+    /// From #[diplomat::attr()]. Mark an error type as a catchable runtime failure (Exception in Dart/catchable).
+    pub catchable: bool,
```
robertbastian: I don't think "catchable" is a good term for implementing an exception trait. In Dart, everything is throwable and catchable

Java uses the `custom_error` bit to implement the exception interface

tool/templates/dart/enum.dart.jinja
```diff
@@ -6,7 +6,13 @@
...
 enum {{type_name}}
    {%- if let Some(it) = special.iterable %} with core.Iterable<{{it}}> {%- endif %}
-   {%- if special.comparator -%} implements core.Comparable<{{type_name}}> {%- endif %} {
+   {%- if is_bug && special.comparator %} implements core.Error, core.Comparable<{{type_name}}> {
+   {%- else if is_bug %} implements core.Error {
+   {%- else if is_error && special.comparator %} implements core.Exception, core.Comparable<{{type_name}}> {
+   {%- else if is_error %} implements core.Exception {
+   {%- else if special.comparator %} implements core.Comparable<{{type_name}}> {
+   {%- else %} {
+   {%- endif %}
```
robertbastian: this is error-prone because you need to check all combinations. you're skipping some because they're not reachable with current rust code

instead of this, calculate a `Vec` of interfaces in the Rust code, and then in the template merge them with ", "

tool/src/dart/mod.rs
```diff
@@ -171,16 +171,24 @@ impl<'cx> ItemGenContext<'_, 'cx> {
...
             },
         )
     }
 
-    fn gen_enum(&mut self, ty: &'cx hir::EnumDef, type_name: &str) -> String {
+    fn gen_enum(&mut self, ty: &'cx hir::EnumDef, id: TypeId, type_name: &str) -> String {
+        let is_error = ty.attrs.custom_errors
+            || ty.usage.results.iter().any(|r| match &**r {
+                hir::ResultUsage::Output(info) => info.err.as_ref().and_then(|e| e.id()) == Some(id),
+                hir::ResultUsage::Input(info) => info.err.as_ref().and_then(|e| e.id()) == Some(id),
```
robertbastian: I don't think we should automatically implement `Exception` just because something appears in a `Result`. We sometimes use `Result` to return unused data in case of an error, however that doesn't make that data an `Exception`.

Clients like ICU4X should just mark their error types as errors (well, their "exception" types as errors and their "error" types as "bugs")

core/src/hir/attrs.rs
mosuem: I just added a `bug` field now.

Manishearth: Hmm. Can this be `error_is_program_bug` or something? This seems very Dart-specific so if we pick a name it should either mention dart or be sufficiently precise.

robertbastian: This should also not say "panic" in the docs, because this is not a Rust panic (Diplomat doesn't handle those).

I'm still not convinced that we should be making this distinction in Diplomat/Dart that just doesn't exist in Rust. In the context of https://github.com/unicode-org/icu4x/pull/8231 I don't consider `MismatchedCalendarError` to be fundamentally different from all the other errors.

robertbastian (commented):
.

core/src/hir/attrs.rs
mosuem: > Hmm. Can this be `error_is_program_bug` or something? This seems very Dart-specific so if we pick a name it should either mention dart or be sufficiently precise.

Happy to rename, I thought we might want to keep diplomat Dart-agnostic.

> I don't consider MismatchedCalendarError to be fundamentally different from all the other errors.

I think the difference is that this is a user error, a user can avoid this error by writing correct code. It cannot happen through external factors outside of the users control, or cannot be checked beforehand. So it is an error and not an exception in my view.

robertbastian: I think then something like `precondition_violation` would be a more appropriate name, and we need to document the precondition on functions that return errors like this.

Still, this should not draw parallels to a panic.

mosuem: Done

feature_tests/dart/lib/src/ErrorEnum.g.dart
```diff
@@ -3,7 +3,7 @@
 
 part of 'lib.g.dart';
 
-enum ErrorEnum {
+enum ErrorEnum implements core.Exception {
```
robertbastian: one of these should be marked `precondition_violation` to actually test your code

tool/templates/dart/enum.dart.jinja
```diff
@@ -42,6 +42,11 @@ enum {{type_name}}
   @override
   int get hashCode => 42; // Cannot get hash from Rust, so a constant is the only correct impl
   {%- endif %}
+  {%- if is_precondition_violation %}
```
robertbastian: this makes it clearer that the `get stackTrace` method is for the `core.Error` class, and avoids the extra bool field

```suggestion
  {%- if implements.contains("core.Error") %}
```

tool/templates/dart/opaque.dart.jinja
```diff
@@ -5,9 +5,9 @@
 @core.Deprecated('{{ deprecated }}')
 {% endif -%}
 final class {{type_name}}
-  {%- if let Some(it) = special.iterable %} with core.Iterable<{{it}}> {%- endif %} implements ffi.Finalizable
-  {%- if special.comparator -%}, core.Comparable<{{type_name}}> {%- endif %}
-  {%- if let Some(it) = special.iterator %}, core.Iterator<{{it}}> {%- endif %} {
+   {%- if let Some(ext) = extends_class %} extends {{ext}}{% endif %}
```
robertbastian: why does the enum `implements` the class, but the struct `extends` it?

mosuem: Dart `enum`s [can't extend classes](https://github.com/dart-lang/sdk/issues/53853#issuecomment-1778040294) unfortunately, Dart `final class`es can.

robertbastian: but the struct could also `implements` it? what's the advantage of using `extends`?

feature_tests/dart/lib/src/ErrorEnum.g.dart
```diff
@@ -3,12 +3,15 @@
...
 part of 'lib.g.dart';
 
-enum ErrorEnum implements core.Exception {
+enum ErrorEnum implements core.Error {
   // ignore: public_member_api_docs
   foo,
   // ignore: public_member_api_docs
   bar;
 
+
```
robertbastian: nit: whitespace

tool/templates/dart/opaque.dart.jinja
robertbastian: nit: indentation

tool/src/dart/mod.rs
```diff
@@ -209,13 +220,30 @@ impl<'cx> ItemGenContext<'_, 'cx> {
...
             is_contiguous: is_contiguous_enum(ty),
+            implements,
             special,
         }
         .render()
         .unwrap()
     }
 
     fn gen_opaque_def(&mut self, ty: &'cx hir::OpaqueDef, type_name: &str) -> String {
+        let special = self.gen_special_method_info(&ty.special_method_presence);
```
robertbastian: suggestion: make `gen_special_method_info` return the `implements` set, which should be a `BTreeSet<&'static str>`, so this code isn't duplicated.

you can then specialize the opaque code path with `implements.add("ffi.Finalizable")`, and `let extends_class = implements.take("core.Error")`

tool/templates/dart/opaque.dart.jinja
mosuem: `extends` means we don't have to override the `StackTrace` getter.

tool/src/dart/mod.rs
mosuem: I'm not sure if I got it right - is this what you meant?

robertbastian: yes

robertbastian (commented):
lgtm, but I'd like @Manishearth to approve

Manishearth:
I think I have a slight preference for `error_is_precondition_violation` but I don't care that much.

PR #1221, opened 2026-07-16
Rework cpp

twantonie:
This body of work significantly changes what the C++ interface looks like. These changes were made to bring the generated C++ interface more in line with other higher level languages at no additional overhead. This should actually improve overhead since a std::unique_ptr with custom deleter takes up more space than a [single pointer](https://stackoverflow.com/a/13460653).

To improve ergonomics and remove raw pointer management all together, three new structs were introduced. OpaquePointer which acts like a unique_ptr without the overhead, Ref which is used in places a reference is required but a normal C++ reference can't be used and Optional which acts like a std::optional, but for OpaquePointer and Ref where it boils down to a single pointer. Very similar to Option<Box<>> in rust. With these in place, no raw pointer management needs to happen. The nanobind destructor hack can go away. The interface clearly indicates an optional pointer or not (before you could just get an empty unique_ptr). And we can finally have C++ constructors.


twantonie:
This is quite a significant change, mostly done by AI. I have only reviewed the tests changes deeply. If you like the look of these changes, I will review the tool generation side of things and get it all working for nanobind.

Manishearth:
Hmm, this changes C++ public API, right? We're not currently doing that unless it can be done cleanly as a config option (and even then I'm worried about the maintenance burden of two parallel ways of doing this type of thing).

We may do C++ breaking changes eventually (timed with ICU4X 3.0, potentially), but not soon.

twantonie:
Yes, this changes the C++ public API significantly. I fully understand wanting it moved under a config option such that current users of the library remain unaffected. With that in place, would you consider merging these changes for the benefits it can provide. Even though it increases the maintenance burden?

I've looked at alternate ways to fix the typing in nanobind for Opaque Optionals, but the alternatives are to either have another post step to manually fix up the typing. Or generate our own python type signatures. The first is unpleasant, the second is also a lot of maintenance work.

Manishearth:
I have further thoughts but will post them when I'm back from vacation next week.

twantonie:
@Manishearth, I'd like to hear your further thoughts. I've stalled development on this for now, since it's usable for me as is. But I'd prefer for this to be merged in rather than having to maintain a fork forever

Manishearth:
Yep, I'm still working through my queue of items from three weeks of vacation. This is on my list, I just also have a lot of big things to look at. Hopefully will get to it tonight or tomorrow but no promises. My position is still leaning "no" but I'll explain my rationale when I get a chance.

Until then I have some questions I can ask.

Generally I like to look at things from a problems-first perspective: show me the problems you are trying to solve. Is the problem mostly just the lack of distinction between nullable and non nullable at the type layer? What problems does that cause for C++ itself? I understand it causes problems for nanobind. Are your interests primarily from a nanobind angle or are you using C++ as well?

twantonie:
No worries. 

The nullable vs non-nullable was the straw that broke the camel's back for me. For C++ the problem is is that you need to check every unique_ptr you receive since you have no way of knowing whether it contains a value. Even though from a Rust perspective it should always contain a value unless specifically stated otherwise. For python it's almost worse, since you're declaring you're always returning a value, but one day it suddenly might not be, so the python API might quietly break on you.

At the moment I'm mostly concerned with the python side of things, since that's most likely what's going to be used, but I've done a lot of C++ programming and I thought the API could be better. Some of the other problems that I came across were:
- No more relying on internal nanobind header files, meaning we should always be able to update when there is a minor release
- The inability to have constructors with the C++ API
- Having to use raw pointers for parts of the C++ API

Manishearth:
Okay. So a couple thoughts.

I'll start out by saying that the `unique_ptr` design was very intentional. I spent a bunch of time researching C++ idioms before doing this, and it seemed like the idiomatic way of doing optional unique pointers is to just use a `unique_ptr` (and for references to use a raw pointer).

The CPP1 backend actually did things closer to this way: but it made borrows harder, and you can't do things like `std::shared_ptr<Opaque>` [^1] without incurring multiple layers of indirection. 


Note that Chromium uses [`Managed<TemporalType>`](https://source.chromium.org/chromium/chromium/src/+/main:v8/src/objects/js-temporal-objects.h;l=42?q=js-temporal-&ss=chromium), which is equivalent to a `shared_ptr`. So that itself makes this direction a bit of a non-starter: there are active users who care about the properties of the unique_ptr design.

You've solved the borrows problem with the `Ref` type, but I'm not sure if we want to introduce our own smart pointer universe here. Many C++ codebases do their own smart pointers and mixing stdlib with custom smart pointers often works but mixing two different custom smart pointer universes can be painful.

Constructors is a bit of a loss but it is also mostly idiomatic for this type of code.

Basically, a lot of these issues boil down to idiomatic C++ being suboptimal.


I think a potentially interesting path would be to have this as a _wrapper_. Maybe something the c++ backend does automatically (but `#ifdef`'d off), or maybe another backend called `cpp-wrap` that produces smarter output and uses the current C++ backend as a base. I'm not sure. This would probably be harder than tweaking the existing backend (if you want to copy over signatures) though there may be a way to share most of the code, I'm not sure. I'm generally worried about complicating the C++ backend.

There may also be a way to use the reflection functionality that C++ has to copy over signatures automatically.

I haven't thought too deeply about this, all of this is stuff in the realm of needing a proper proposal with API before I can be sure. Generally I want to see proposals before large refactors, especially LLM refactors done by people who haven't yet worked much on the codebase[^2].

Overall I'm not yet convinced of the benefit of this complexity when looking just at the C++ backend. For nanobind, I'd prefer this be solved in the nanobind backend as much as possible, which brings me to my next topic:

> I've looked at alternate ways to fix the typing in nanobind for Opaque Optionals, but the alternatives are to either have another post step to manually fix up the typing. Or generate our own python type signatures. The first is unpleasant, the second is also a lot of maintenance work.

FWIW, a Diplomat design principle is that individual backends do not foist much complexity on other backends. The attribute system and HIR were designed with this in mind, and it enables people to work on Diplomat backends without having to deal with fixing other backends, some of which may be lower quality or deal with things the backend maintainer just doesn't know about.

This is also why I felt comfortable merging the mostly-LLM-coded .NET backend without deep review: if a project is finding that good enough for their use, that works for me, and it won't get in my way maintaining other backends if they end up having problems.

So "a lot of maintainence work for nanobind" vs "a lot of maintainence work for C++" are meaningfully different. Nanobind is a bit special in that it calls the C++ backend, so that separation breaks down a bit. I'm happy to accept _some_ C++ complexity for nanobind but not too much.


 [^1]: This works even with opaque types, the refcount goes in a different allocation.

Very much in favor with you and @tyler-zeromatter and others figuring out the best path forward for fixing this problem in nanobind in the other thread. If the conclusion ends up being to make changes to the C++ backend that's a signal, though I'm still reluctant to making changes this big.

> No more relying on internal nanobind header files, meaning we should always be able to update when there is a minor release


Say more? Also, are you a ZeroMatter person using nanobind or from a different project? (Out of curiosity)


> This should actually improve overhead since a std::unique_ptr with custom deleter takes up more space than a [single pointer](https://stackoverflow.com/a/13460653).

I don't think this is true: this isn't a `unique_ptr` with a custom deleter (a separate generic arg), this is a `unique_ptr` with a custom `delete`, which is on the type itself. The pointers here are regular `std::unique_ptr<T>` with no additional generics.

 [^2]: Not a dig or anything. I have found LLMs to be useful, even for large projects, however I have repeatedly found that LLM contributions from a project maintainer are miles better than those from people new to the project. LLMs have made code cheap: but maintenance and code review is still a cost, and LLMs don't necessarily know all of our constraints. When someone approaches with a large LLM PR that I need to review, getting it up to scratch is often a more arduous process than if I had used the LLM myself to do it, at which point I'm going to weigh that PR against the cost of me doing it myself with LLMs, and often decide that it's not worth the effort. The .NET backend is different: it's a new backend and the people contributing it are maintaining it, so as long as their LLM-coded backend impl obeyed certain easy to check principles I didn't need to put too much effort into reviewing it.

twantonie:
Thanks for giving me such a detailed response. From the context you've provided, it makes a lot more sense why you went with the unique_ptr approach. (and I made too many assumptions without reading the code as diligently as I should have with regards to the custom deleter). 

> I haven't thought too deeply about this, all of this is stuff in the realm of needing a proper proposal with API before I can be sure. Generally I want to see proposals before large refactors, especially LLM refactors done by people who haven't yet worked much on the codebase[2](https://github.com/rust-diplomat/diplomat/pull/1221#user-content-fn-2-182fddb341237ae9447a15d4d48f8cca).

In this case I thought a show don't tell might be the easier approach. (and it was a bit of an exercise for me into what an LLM is capable of)

> Say more? Also, are you a ZeroMatter person using nanobind or from a different project? (Out of curiosity)

I work for a completely separate company. I'm using diplomat to expose hardware control via a shared library.

I think we should just discard this work. Maybe improve some documentation on the C++ side of things so the next person will have a better understanding of why the C++ interface is the way it is. 

I still think we can do better on the optional side of opaque pointers though. The idiomatic C++ way assumes you generally don't use raw pointers (and rather use references) unless you specifically say that something is optional and then you use the pointer. The problem with diplomat is is that everything is already a pointer, so an additional layer is necessary to show when you can expect a pointer to be null and when not.

I think @tyler-zeromatter suggestion of using something like `maybe_null<std::unique_ptr<T>> would communicate intent much better. This should probably also be done for raw pointers that can be optional as well. To not change the API, we could put it behind a configuration option, and then change the default over after sufficient warning. I can probably look at doing that, but I'll be going on holiday for a month next week, so it might be awhile before I finish it.

Manishearth:
Sounds good. `maybe_null` sounds like something that can be done behind a toggle, too, it wouldn't be an extremely invasive change.

Please make sure to have a small end-to-end proposal for a change like that before writing the PR.

> The idiomatic C++ way assumes you generally don't use raw pointers (and rather use references) unless you specifically say that something is optional and then you use the pointer. The problem with diplomat is is that everything is already a pointer, so an additional layer is necessary to show when you can expect a pointer to be null and when not.


I'm not sure what you mean here. Diplomat consistently uses references (not pointers) for non nullable borrows.

This is a problem for `unique_ptr` (where we have no such distinction), but not for the reference/raw pointer situation. I don't see that as a problem in need of fixing.

feature_tests/cpp/tests/callback.cpp
```diff
@@ -114,22 +114,21 @@ int main(int argc, char *argv[])
...
     auto a = Opaque::from_str("This is a test value.").ok().value();
-    auto ptr = a.get();
     {
-        auto str = o.test_option_opaque([ptr]() {
-            return ptr;
+        auto str = o.test_option_opaque([&a]() {
+            return diplomat::Optional(a.as_ref());
         });
         simple_assert_eq("Test opaque string passing", str, "\"This is a test value.\"");
     }
```
twantonie, line 122: This is what I'm referring to. You return a pointer, in this case the pointer is an optional, so can be a nullptr. Generally you would run into a segfault were you to return a nullptr in this case

feature_tests/cpp/tests/slices.cpp
```diff
@@ -7,29 +8,19 @@ using namespace somelib;
...
+    diplomat::span<MyString> in(arr.data(), arr.size());
     simple_assert_eq("Slice of opaques", MyString::slice_of_opaques(in), "Test String end.");
 
-    const MyString* optional_arr[] = {
-        a.get(), nullptr, b.get()
-    };
-    diplomat::span<const MyString*> optional_in(optional_arr, 3);
+    std::vector<diplomat::Optional<MyStringRef>> optional_vec{diplomat::Optional(arr[0].as_ref()), diplomat::Optional<MyStringRef>(std::nullopt), diplomat::Optional(arr[1].as_ref())};
+    diplomat::span<diplomat::Optional<MyStringRef>> optional_in(optional_vec.data(), optional_vec.size());
     simple_assert_eq("Optional slice of opaques", MyString::optional_slice_of_opaques(optional_in), "Some(MyString(\"Test\")) None Some(MyString(\" String \")) ");
```
twantonie, line 18: Here as well

PR #1225, opened 2026-07-21
dotnet: support both UTF-8 and UTF-16 strings, zero-copy

irvingoujAtDevolution:
Adds UTF-16 string support to the .NET backend alongside the existing UTF-8 path, and closes a related gap: borrowed string/slice returns (`&'a str` / `&'a [u8]` / `&'a [u32]`), which `lower_return` had no `Type::Slice` arm for at all.

Inputs split three ways by what Rust's own safety contract actually requires. `&DiplomatStr16` and `&DiplomatStr` (unvalidated UTF-8, no caller-side validity requirement) both pin zero-copy — the latter is reshaped to `byte[]`/`ReadOnlyMemory<byte>` and reuses the exact `&[u8]` codegen path, since that's really all it is. A validated `&str` still copies (only a transcode from a real `System.String` can guarantee well-formed UTF-8), but the copy is routed through a named `Diplomat.Utf8.Clone(...)` helper instead of an inlined `Encoding.UTF8.GetBytes` call, so it's a visible, named step rather than something buried in marshalling.

Borrowed string/slice returns get a new `DiplomatBorrowedSpan<T>`, rooted with the same keep-alive edges a borrowed opaque return already uses. It deliberately doesn't expose a `Span`-returning property — a bare `Span<T>` carries no reference back to the edges keeping the borrow's owner alive, so a caller who only kept the span would be left holding a dangling pointer the moment the owner got collected. `WithSpan(...)` scopes access to a callback instead (same shape as `RustVec`'s), and `Clone()` is the explicit, separate copy. Wrapping a borrowed span in `Result`/`Option` isn't supported yet — the result/option bridging was never exercised for this shape, so it's rejected rather than risking broken codegen.

Regen: un-disabled the 4 feature-test methods that were `#[diplomat::attr(dotnet, disable)]`'d for this (`MyString::borrow`, `OpaqueMutexedString::dummy_str`, `Utf16Wrap::from_utf16`, `Utf16Wrap::borrow_cont`). Existing UTF-8 fixtures also changed: `&DiplomatStr` params flipped from `string` to `byte[]` wherever they appear.

Tests: 38 tool unit tests (dotnet), 84 xunit facts against the built native library — new coverage for empty-string zero-copy pinning, `Clone`, and GC-survival of borrowed views under stress.

### DIFF
--- a/book/src/backends/dotnet.md
+++ b/book/src/backends/dotnet.md
@@ -43,6 +43,38 @@ Consumers are not required to call `Dispose()`; a finalizer is the documented la
 cleanup path for owned handles. Native calls are followed by `GC.KeepAlive(this)` to
 prevent the finalizer from running mid-call and freeing memory a P/Invoke is still using.
 
+## String encoding
+
+The backend supports both UTF-8 and UTF-16 strings, zero-copy wherever the C# and Rust
+representations line up:
+
+* `&DiplomatStr16` params and returns: a C# `string` is already a flat UTF-16 buffer, so
+  these are always zero-copy — pinned directly with `fixed` (or, if the return value
+  borrows it, via `ReadOnlyMemory<char>` + the same pinning holder slices use).
+* `&DiplomatStr` params and returns (unvalidated UTF-8 — Rust places no validity
+  requirement on the caller): treated exactly like `&[u8]`, so these are also zero-copy —
+  `byte[]` / `ReadOnlyMemory<byte>` pinned directly, no transcoding.
+* `&str` params (validated UTF-8 — Rust requires the caller to guarantee well-formed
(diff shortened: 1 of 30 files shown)


Manishearth:
@ambiguousname are you interested in giving this a review? I'm still on vacation for a week. Not a big deal if not.

But it would be nice to get this into the release if it happens before I'm back.

ambiguousname:
I can take a look this weekend, I think

ambiguousname (approved):
Looks good to me.

This roughly makes sense; though C# FFI isn't my expertise, this makes sense to me and tests were passing on my end.

irvingoujAtDevolution:
@Manishearth can you help me merge this one too? thanks!

Manishearth:
Needs merge conflict resolution

Manishearth (approved):

PR #1226, opened 2026-07-24
dotnet: fixed-width bool tag so Result/Option marshal on .NET Framework

irvingoujAtDevolution:
.NET Framework can't marshal a by-value struct return that has a `bool` field — `bool` isn't blittable — so every by-value `DiplomatResult`/`DiplomatOption` return throws `MarshalDirectiveException` on net48 (`Method's type signature is not PInvoke compatible`). Struct `bool` fields hit the same wall. Kotlin already sidesteps this by storing struct-field bools as `Byte`.

This adds a one-byte [`DiplomatBool`](https://github.com/irvingoujAtDevolution/diplomat/blob/dotnet-blittable-bool-tag/tool/templates/dotnet/DiplomatBool.cs.jinja) and uses it for the Result/Option tag and struct `bool` fields. Rust and the public C# API stay `bool` — the conversion lives in the generated raw layer. Params stay `bool` (those marshal fine on net48).

Tests: feature tests now multi-target `net8.0;net48` — 76/76 on both.

### DIFF
--- a/example/dotnet/Generated/DiplomatBool.cs
+++ b/example/dotnet/Generated/DiplomatBool.cs
@@ -1,0 +1,21 @@
+using System.Runtime.InteropServices;
+
+namespace Somelib.Diplomat;
+
+// Blittable one-byte stand-in for `bool`. .NET Framework P/Invoke can't return a
+// struct by value if any field is non-blittable, and `bool` is non-blittable, so
+// Result/Option tags and struct bool fields use a byte (these bindings target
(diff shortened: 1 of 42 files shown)


Manishearth (approved):

irvingoujAtDevolution:
Hi! can you merge it please, I don't have the access looks like.
@Manishearth

Manishearth:
I'm just returning from vacation, will be slow to catch up

PR #1230, opened 2026-07-27
dotnet: render HIR accessors as C# properties

irvingoujAtDevolution:
Switches the .NET backend to HIR's accessor attributes, so `#[diplomat::attr(auto, getter)]` /
`setter` render as C# properties — the same mechanism Dart, JS and Nanobind use.

Before this PR the backend had `accessors = false` and synthesized properties from its own
`getters_prefix` / `setters_prefix` config, keyed off how a method happened to be *named*. No other
backend has such a config, and it meant `set_*` methods could never become properties at all. That
config is deleted.

### The accessor is the property

Its body is generated inline, so no second member is emitted — exactly as in Dart and JS. Each
method is routed as it is built, into the property it backs or into the plain method list. There is
no post-processing pass, no privacy flag, and nothing to hide. (This addresses the review comment on
the earlier revision: the separate post-process routine is gone.)

C# needs one step Dart does not: a getter and a setter cannot be two members of the same name
(CS0102), so a matching pair merges into one property carrying both accessors. A setter with no
getter becomes a write-only property, which is what a Rust config object usually wants.

### A getter and its setter always have one type

Marshalling and the exposed C# type are separate concepts here: `AccessorMarshal` is per-accessor,
`PropertyType` is shared, and a pair may only merge when their `PropertyType`s are equal.

Each string encoding stays its own marshal, because they are not interchangeable — `&str` transcodes
and demands valid UTF-8, `&DiplomatStr` is opaque bytes with no validity contract, `&DiplomatStr16`
pins a C# string with no allocation, and a `DiplomatWrite` getter has Rust append UTF-8 to a buffer
we then decode. All four present as `string`, and that is the one place the mapping collapses; it is
written out one arm at a time so the collapse is reviewable rather than assumed.

That is also what fixes the disagreement this PR would otherwise have: a `&DiplomatStr` parameter
marshals like `&str` inside a property, while keeping its zero-copy `byte[]` shape everywhere else. A
property cannot read as `string` and write as `byte[]`. Every other backend already maps
`&DiplomatStr` to its string type (`std::string_view`, Dart `String`, JS `string`) — the `byte[]`
shape was .NET's outlier.

When no single type can serve both sides, the property is refused with a diagnostic naming both C#
types, both Rust methods, and the marshal each side chose:

```
[.NET backend] property `Data` would need two types: `DiplomatBorrowedSpan<byte>` from getter `data`
(a borrowed view over Rust-owned memory), and `byte[]` from setter `set_data` (a managed array Rust
reads during the call). A C# property has a single type. Make the two Rust signatures agree, or
disable one of them for .NET.
```

The C# types alone would not explain a refusal when both spell `string`, which is why the marshal is
named too.

### No member can collide

Two Rust methods claiming the same `get` or `set` are refused, as is a property whose name matches a
method, a struct field, or a member Diplomat always generates (`AsFFI`, `FromFFI`, `Dispose`). Each
of those is CS0102, so the alternative is C# that does not compile.

### Naming and layering

* `fmt_accessor_name` puts the property name in the formatter, beside `fmt_method_name` and
  `fmt_type_name`. `#[diplomat::rename]` applies *after* case conversion so its value survives
  verbatim (`#[rename = "UTCTime"]` stays `UTCTime`). Dart case-folds after renaming; this backend
  deliberately does not, because its surface has to keep matching already-shipped bindings.
* A setter's value renders as the accessor's implicit `value` — the Rust parameter name is not in
  scope in a property body. The raw layer keeps the Rust name: a P/Invoke declaration is a plain
  method however the idiomatic layer presents it.
* `unsafe` stays inside the accessor rather than on the property, so the idiomatic surface stays
  safe, as it is for every generated method.
* The property block and the doc-comment block are each emitted from one template included by both
  the opaque and struct impls, instead of being copy-pasted into both. The copies had already
  drifted: properties lost the `<returns>` tag the methods loop emits, which is why
  `OpaqueThinVec.First` gains one here. Doc assembly moved to Rust so a property says each thing
  once — two accessors sharing an error type documented the same exception twice.
* `static_accessors` stays off, matching Dart and JS.

Plain methods render byte-identically to before.

### Generated output

```csharp
// MyString.cs — get_str returns a DiplomatWrite, set_str takes &DiplomatStr
public string Str
{
    get { unsafe { /* writer allocated, filled, decoded, disposed */ } }
    set { unsafe { /* value transcoded via Diplomat.Utf8.Clone, pinned, passed */ } }
}

// OpaqueThinVec.cs — a setter with no getter
public string FirstC
{
    set { unsafe { /* ... */ } }
}
```

### Coverage

The accessor attributes already in `feature_tests` now produce properties, so the generated bindings
and the C# tests exercise them, and `set_first_c` gains a setter attribute for the write-only shape.

That still left the feature this PR adds as the least-tested thing in it: **every generated setter in
the tree was `&DiplomatStr`**, so no other marshal's setter half was ever compiled or run. The new
`PropertyMarshals` carries both halves of each one, and `PropertyTests.cs` round-trips every row
through the generated C# on **net8.0 and net48**:

| Property | Getter → setter | Covers |
| --- | --- | --- |
| `Number` | `u32` → `u32` | primitive |
| `Choice` | `DefaultEnum` → `DefaultEnum` | enum |
| `Point` | `PrimitiveStruct` → `PrimitiveStruct` | struct by value, via `AsFFI` / `FromFFI` |
| `Held` | `Box<Opaque>` → `&Opaque` | owned out, borrowed in — one C# type, two ownerships |
| `Utf8Text` | `DiplomatWrite` → `&str` | validated UTF-8, transcoded |
| `Utf16Text` | `DiplomatWrite` → `&DiplomatStr16` | UTF-16, pinned in place |

Writing through the UTF-16 setter and reading back through the UTF-8 getter checks the central claim
— that every text encoding presents as the same `string` — at runtime rather than in a comment. A
separate test reads a struct property, which nothing exercised before, and another checks that the
opaque getter hands back a value independent of a later write through the setter.

`PropertyMarshals` is `#[diplomat::attr(not(dotnet), disable)]`: .NET is the only backend that merges
two accessors into one member, so it is the only one that can get the pairing wrong. No other
backend's generated output changes.

Unit tests cover the rest — each text marshal's own marshalling, each non-text marshal pairing into
one member, nullable agreement and disagreement, one `<exception>` and one `<returns>` per property,
type disagreement, differing byte-slice marshals, duplicate getters and duplicate setters, collisions
with a method / a struct field / `Dispose` / `AsFFI`, a rename surviving verbatim, a static accessor
staying a method, a getter that returns nothing, helper types pulled in by an accessor, and the raw
layer keeping its parameter name.

Two notes for review:

* The unit-test harness now calls itself `dotnet` rather than `dotnet_test`, so
  `#[diplomat::attr(dotnet, …)]` means the same thing in a test as it does to a user. Nothing could
  exercise a positive `dotnet` cfg before.
* One diagnostic is unreachable today and says so in a comment instead of carrying a test that
  pretends otherwise: a callback is the only parameter shape with no property type, and HIR rejects
  callbacks for this backend first. It is the guard for when `callbacks` lands.

The whole mechanism is documented in the backend's book page.


tool/src/dotnet/gen/method.rs
```diff
@@ -839,13 +842,39 @@ impl MethodInfo<'_> {
...
+        .collect();
+
+    for method in methods {
+        if wired.contains(method.name.as_str()) {
+            method.visibility = "private ";
+        }
+    }
+}
+
 pub(super) fn collect_properties(methods: &[MethodInfo<'_>]) -> Vec<PropertyInfo> {
```
Manishearth: why not set the privacy boolean here? I don't think you need a separate post process routine

irvingoujAtDevolution:
Sorry, let me put this back to darft for a sec

Manishearth (approved):
Much larger diff than expected, but looks like it's to handle more things.

PR #1231, opened 2026-07-29
dotnet: ban problematic borrowed-span return paths

irvingoujAtDevolution:
Ban problematic borrowed-span return paths until support is explicitly added.

- Reject borrowed span returns that borrow a slice/string parameter (no Dispose path to unpin).
- Reject `'static` string/slice returns for now.


Manishearth (approved):
needs fmt

irvingoujAtDevolution:
Updated

Manishearth (approved):

PR #1236, opened 2026-08-04
Switch diplomat-runtime to 0.15.1

Manishearth:
To improve the experience for users of multiple Diplomat libraries, this switches diplomat-runtime over to 0.15.1 to avoid multiple copies in a single binary. At some point we shall update everyone to 1.0. This would need to be done coordinated between ICU4X and Temporal.

https://github.com/rust-diplomat/diplomat/issues/958

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -399,7 +399,7 @@ dependencies = [
 
 [[package]]
 name = "diplomat-runtime"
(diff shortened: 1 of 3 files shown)


tyler-zeromatter (approved):
LGTM

PR #1242, opened 2026-08-05
Improve some diplomat-runtime docs

Manishearth:
Bunch of docs cleanups, added an overview section, and added some clarity on the slice types.

PR mostly made by getting gemini to do what improvements it saw, and then cherry picking. I ended up redoing the type overview section myself.


ambiguousname (approved):
The additions here make sense to me; I think the module docstring now clearly outlines what types we provide and their general use case

I currently can't speak to anything that might be missing which we'd want to explain to runtime users (I can review more thoroughly tomorrow if you remind me), but these additions look good

sffc (commented):
Good improvements

PR #1244, opened 2026-08-05
dotnet: make IDisposable opt-in for opaques

irvingoujAtDevolution:
Opaque wrappers now use finalizer cleanup by default. Types that need deterministic release can opt in with `#[diplomat::attr(dotnet, manually_disposable)]`.

This keeps the existing edge lifetime model unchanged and avoids making every native-backed C# object carry the `IDisposable` contract.

Tests: `cargo test -p diplomat-tool dotnet`, `cargo make test-dotnet`, clippy


feature_tests/dotnet/Generated/Unnamespaced.cs
```diff
@@ -35,11 +36,6 @@ internal unsafe Unnamespaced(Raw.Unnamespaced* handle)
         _edges = System.Array.Empty<object>();
     }
 
-    /// <remarks>
```
Manishearth, line 38: why were these comments removed?

irvingoujAtDevolution: Hi @Manishearth, still in draft, I'll address your comments when it's done

example/dotnet/Generated/RustHandle.cs
```diff
@@ -1,50 +1,384 @@
...
 namespace Somelib.Diplomat;
 
 #nullable enable
 
-/// <summary>
-/// Frees a Rust-owned <typeparamref name="T"/> by calling its native
-/// destructor. Held by an owned <see cref="RustHandle{T}"/> so the handle can
-/// release the pointer without knowing the concrete type.
-/// </summary>
 internal unsafe delegate void RustDestructor<T>(T* ptr) where T : unmanaged;
```
CBenoit, line 12: Manishearth left a similar comment elsewhere, but I see this is a repeating pattern: let’s be careful to not change other (seemingly) unrelated pieces of code. I know this is a draft, and you’ll probably go through it once before lifting draft status, just wanted to make it a clear ask

feature_tests/src/lifetimes.rs
CBenoit, line 1: This is extending by a lot the _common_ testing code mostly for testing .NET backend specific behaviors; at a glance maybe a bit too much? Not sure, maybe it’s fine. My instinct would be to keep most of that directly as tests in the dotnet backend itself (IIRC we used to have tests directly along the backend tool code).

tool/templates/dotnet/opaque.impl.cs.jinja
```diff
@@ -71,54 +26,20 @@ public partial class {{ name }}: IDisposable
...
-
-            GC.SuppressFinalize(this);
-        }
-    }
-
-    ~{{ name }}()
-    {
-        Dispose();
+        _inner.Dispose();
+        _inner = default;
```
CBenoit: Maybe worth an inline comment to explain what we are doing here.

feature_tests/dotnet/Tests/BorrowedReturnTests.cs
```diff
@@ -17,7 +17,7 @@ public class BorrowedReturnTests
     [Fact]
     public void First_ReturnsReadableBorrowedView()
     {
-        using OpaqueThinVec vec = OpaqueThinVec.CreateSingle(7, 1.5f, Utf8("hello"));
+        OpaqueThinVec vec = OpaqueThinVec.CreateSingle(7, 1.5f, Utf8("hello"));
```
CBenoit: I was confused as to why the `using` was removed here; it appears `IDisposable` is still implemented on `OpaqueThinVec`. Is it an intended change? I would suggest splitting into multiple PRs before removing the draft status, for ease of review

CBenoit (commented):
Chiming in to give a quick feedback.

The initial motivation was to fix a soundness hole where borrowed views were outliving the borrowed object.
In that context, IDisposable was seen as an easy source of problems, because it was possible to Dispose/free a resource still borrowed.

I understand that the approach here is, when calling Dispose, to not immediately free an object while it is being borrowed, and instead free it once all the borrowers are themselves freed, is my understanding correct?

This is something we can discuss before we commit to it, and I would recommend to implement that in a separate PR for reviewability anyway, but I’m still under the impression that it would be better if the user had control on which structs are implementing IDisposable, and which are not (default to not implementing it). The rationale for that is that not calling `Dispose` on an `IDisposable` object is considered bad, as such it’s _linted against_. I think _most_ classes don’t need such strict scopes once in the C# world. A Diplomat attribute that is only recognized by the dotnet backend may be ideal: `#[diplomat::attr(dotnet, idisposable)]` (not sure if the syntax is right, writing from memory). I would keep it for things such as `RustVec`, buffers and views/spans on them in general, because it’s memory we would rather free sooner than later.

Alternatively, if you still want to have an explicit way to free the native object sooner, I recommend we provide a different interface such as `RustFreeable` (feel free to come up with a better naming), that does not come with the `IDisposable` contract, and would remain optional. When `IDisposable` is implemented, simply forward `Dispose` to the function provided by `RustFreeable` (same in the deconstructor).

On a different note: moving more responsibilities to the `RustHandle` seems to achieve compression of the generated code through shared logic, this is a good thing.

Manishearth:
> The initial motivation was to fix a soundness hole where borrowed views were outliving the borrowed object.


This shouldn't be happening if edges are set up correctly.

irvingoujAtDevolution:
> > The initial motivation was to fix a soundness hole where borrowed views were outliving the borrowed object.
> 
> This shouldn't be happening if edges are set up correctly.

Yes, edges handle GC reachability correctly. The hole is explicit Dispose(): it can free the native source even while a borrowed view still holds a managed edge to it.

We considered removing IDisposable and relying entirely on finalization, but that would make releasing native resources and pinned buffers nondeterministic. This PR keeps immediate disposal when there are no borrowers, while existing views retain the source SafeHandle and delay only the physical destruction.

Still in draft, I will let you know when it's ready.

CBenoit:
I used Copilot for comparing my mental model of what the final shape should look like ultimately to obtain a sound API (hopefully I’m not missing anything). This is besides the edges being properly set up.

Here is the prompt I used:

> Can you compare this PR against my mental model?
> My expectations are that this PR is:
> 
> - Making IDisposable opt-in via a new diplomat attribute.
> - By not implementing IDisposable, in the vertue of removing the public `Dispose` function available to consumer code, the API is now _sound_, that is: it’s not possible create a memory corruption by calling `Dispose` on a borrowed object [before the borrowing objects are themselves freed].
> - When IDisposable is implemented, and thus when freeing is controllable by the consumer code, ref-counting is implemented so that memory reclaiming is delayed to the point where all borrowing views are themselves freed.

Here is the full response:

<details>
<summary>Copilot’s answer</summary>

**Partly. Your first two points match the current PR; the third does not.**

1. **`IDisposable` becomes opt-in — correct.**  
   `#[diplomat::attr(dotnet, idisposable)]` is accepted only on opaque types. Unmarked opaques no longer implement `IDisposable` and expose no public `Dispose()`; they retain a finalizer that invokes private cleanup.

2. **Default (non-`IDisposable`) opaques close the explicit-disposal soundness hole — correct, with the normal public C# API in mind.**  
   Borrowed wrappers retain their source wrapper through `_edges`, preventing ordinary GC/finalization of the source while a borrowing view remains reachable. Since the source has no public `Dispose()`, consumer code cannot explicitly destroy its native allocation early. Thus the problematic sequence—create a borrowed view, explicitly dispose its parent, then use the view—is no longer expressible for a default opaque.

3. **Opting into `IDisposable` does *not* add ref-counted/deferred native destruction — incorrect.**  
   At the PR’s current head (`0d3410089ff1ca7561de135491db25b907228a3a`), `RustHandle<T>` is just a pointer plus an optional destructor delegate. `Release()` immediately calls that destructor for owned handles; there is no `SafeHandle`, reference count, or “outstanding borrower” tracking.

   Consequently, for an opted-in opaque, this remains unsafe:

   ```csharp
   var parent = /* owned, IDisposable opaque */;
   var view = parent.GetBorrowedView();
   parent.Dispose(); // immediately destroys the native parent
   view.Use();       // may use memory borrowed from the now-freed parent
   ```

   The generated-template comments say this explicitly: edges keep the owner **GC-reachable**, but disposing an opted-in parent while a borrowing child is in use “is still a use-after-free and remains the caller’s responsibility.”

So the more precise mental model is:

| Opaque kind | Public `Dispose` | GC lifetime protection via edges | Explicit parent disposal while borrower lives |
|---|---:|---:|---|
| Default | No | Yes | Not available through the ordinary public API |
| `#[diplomat::attr(dotnet, idisposable)]` | Yes | Yes | **Still immediately frees native memory; caller must avoid it** |

One noteworthy discrepancy: a PR comment describes an approach where existing views retain a source `SafeHandle` and physical destruction is deferred. That is **not what the current PR code implements**: the current generated `RustHandle<T>` has no `SafeHandle` or reference-counting mechanism.
</details>

Here is the state of the code right now:

> | Opaque kind | Public `Dispose` | GC lifetime protection via edges | Explicit parent disposal while borrower lives |
> |---|---:|---:|---|
> | Default | No | Yes | Not available through the ordinary public API |
> | `#[diplomat::attr(dotnet, idisposable)]` | Yes | Yes | **Still immediately frees native memory; caller must avoid it** |

With the explanation:

> This remains unsafe:
> 
> ```csharp
> var parent = /* owned, IDisposable opaque */;
> var view = parent.GetBorrowedView();
> parent.Dispose(); // immediately destroys the native parent
> view.Use();       // may use memory borrowed from the now-freed parent
> ```
> 
> The generated-template comments say this explicitly: edges keep the owner **GC-reachable**, but disposing an opted-in parent while a borrowing child is in use “is still a use-after-free and remains the caller’s responsibility.”

I disagree with the decision that avoiding the use-after-free here should be the caller’s responsibility, unless the API clearly conveys that premise, which it does not right now. For instance: marking the function as unsafe.

I think that the delayed free was a satisfying answer though. We don’t need it at all unless IDisposable is implemented, and we can simplify the generated code / remove allocations for objects that do not implement IDisposable, but it would be worth keeping it otherwise. What do you think?

CBenoit:
Leaving a follow up comment based on Slack discussion: this is in itself a strict improvement over the status quo, so if @Manishearth is okay too, I’m fine with landing the subset already implemented:

> - **Making IDisposable opt-in via a new diplomat attribute.**
> - **By not implementing IDisposable, in the vertue of removing the public `Dispose` function available to consumer code, the API is now _sound_, that is: it’s not possible create a memory corruption by calling `Dispose` on a borrowed object [before the borrowing objects are themselves freed].**

Keeping the third item as a follow up:

> - When IDisposable is implemented, and thus when freeing is controllable by the consumer code, ref-counting is implemented so that memory reclaiming is delayed to the point where all borrowing views are themselves freed.

We can discuss about the exact shape of this separately.

irvingoujAtDevolution:
Hi @Manishearth

Here's the full story of why I think this PR is necessary. Two things before the technical details:

1. This is a breaking change: behavior that used to be the default is now opt-in.
2. The old default was my design choice. I focused too much on explicit resource management and performance, and made the generated API less sound as a result.

Technical:

C# has `IDisposable`, with language support through `using`. It gives callers deterministic cleanup on top of finalization and GC. For example:

```csharp
class HoldingSystemResource : IDisposable
{
    private SystemResource _resource;

    public HoldingSystemResource()
    {
        _resource = new SystemResource();
    }

    public void Dispose()
    {
        _resource.Dispose();
    }
}

// And you can use it like this:

class Program
{
    static void Main(string[] args)
    {
        using (var resource = new HoldingSystemResource())
        {
            // Use the resource here
        } // Dispose is called automatically here

        // or you can just call
        var anotherResource = new HoldingSystemResource();
        anotherResource.Dispose(); // manually call Dispose when done
    }
}
```

This is a useful primitive and gives C# finer resource control than Kotlin, Dart, and other GC languages. It is also idiomatic to use `using` whenever a type implements `IDisposable`. Humans, analyzers, and LLMs will all naturally do that when the interface is present.

This is normally fine. It becomes a problem when Rust returns something that borrows from another opaque.

## Example

```rust
#[diplomat::opaque]
pub struct Store(Vec<Item>);

#[diplomat::opaque]
pub struct Item {
    value: u32,
}

impl Store {
    pub fn create() -> Box<Self> {
        Box::new(Self(vec![Item { value: 42 }]))
    }

    pub fn first<'a>(&'a self) -> Option<&'a Item> {
        self.0.first()
    }
}
```

Rust guarantees that the returned `Item` cannot outlive `Store`. C# cannot express this lifetime, so we use `_edges` to keep the `Store` wrapper alive. This works for GC, but not for an explicit `Dispose()`:

```csharp
var store = Store.Create();
var item = store.First!;

store.Dispose();               // frees the Vec<Item>
Console.WriteLine(item.Value); // item now points into freed memory
```

The edge is still there, but it only keeps the managed `store` object alive. It cannot undo the native free that already happened inside `Dispose()`.

The same bug can hide behind completely normal C#:

```csharp
static Item GetFirstItem()
{
    using var store = Store.Create();
    return store.First!;
}

var item = GetFirstItem();
Console.WriteLine(item.Value); // use-after-free
```


Hence, in this PR, I propose that we make this specific C# feature opt-in.



## Possible alternative solutions to avoid the breaking API change

We have considered a few alternative solutions so that the breaking change is not introduced. Many solutions are not applicable here for various reasons, but there was one that I thought would be even better than making `IDisposable` opt-in. That solution would also solve a class of bugs in C# that can be reproduced in other backends.

### The non-guaranteed finalization order of the GC

In JVM, Dart, JS runtimes, and .NET, once objects become unreachable, they may be collected and their generated destructor/finalizer called without a guaranteed order of finalization. The `_edges` primitive doesn't help with this in any of these cases.

For example:

```rust
#[diplomat::opaque]
pub struct Source(Box<u8>);

#[diplomat::opaque]
pub struct Dependent<'a>(&'a u8);

impl Source {
    pub fn new() -> Box<Self> {
        Box::new(Self(Box::new(42)))
    }

    pub fn dependent<'a>(&'a self) -> Box<Dependent<'a>> {
        Box::new(Dependent(&self.0))
    }
}

impl Drop for Dependent<'_> {
    fn drop(&mut self) {
        // Valid in Rust: Source must outlive Dependent.
        assert_eq!(*self.0, 42);
    }
}
```

The generated `Dependent` has `_edges = new object[] { source }`. This keeps `source` reachable while `dependent` is reachable, but it does not guarantee the finalization order when both become unreachable:

```csharp
static void CreateAndForget()
{
    var source = Source.New();
    var dependent = source.Dependent();

    // No Dispose here. Both simply become unreachable.
}

CreateAndForget();
GC.Collect();
GC.WaitForPendingFinalizers();
```

The GC may finalize `source` first, freeing the `Box<u8>`, and then finalize `dependent`. The Rust destructor of `Dependent` now reads freed memory. We reproduced this with the generated bindings and a real Rust drop probe: it crashes with `0xC0000005` even though `_edges` is present and nobody calls `Dispose()`.

This means `_edges` gives us managed reachability, but not native destruction ordering. The dependent needs to retain the native lifetime of its source until its own destructor is done.

More importantly, this can be reproduced on [this branch](https://github.com/Devolutions/diplomat/tree/experiment/kotlin-js-finalizer-order-repro) for JS, Kotlin, and .NET.

The solution I was thinking about is some sort of reference counting. From this PR's commit history, you can see that we had something conceptually like this:
```csharp
class NativeObject
{
    private int _references = 1;

    void AddReference() => _references++;

    void Release()
    {
        if (--_references == 0)
            DestroyNativeObject();
    }
}

// Dependent borrows Source, so it holds one reference to Source.
source.AddReference();

// Source.Dispose() releases its own reference, but Source stays alive.
source.Release();

// Dependent is destroyed first, then releases the borrowed Source.
dependent.DestroyNativeObject();
source.Release();
```
This is essentially what `Rc` does in Rust. This is way beyond what we initially thought was in scope; it surfaced while I investigated the .NET issue. I think it is necessary for the team to be aware of this problem. I hope we can have this PR merged, and then we can discuss the finalization order problem further.

Best,
Irving

Manishearth:
I would like the attribute to either be named `dotnet_idisposable`, or (preferred) have a name that is generic, something like `manually_disposable`.

Rest seems fine.

irvingoujAtDevolution:
@Manishearth Updated, thanks!

Manishearth (approved):

PR #1246, opened 2026-08-06
dotnet: retain native borrow dependencies until teardown

irvingoujAtDevolution:
Keeps each native borrow source alive until its dependents have run their Rust destructors. This fixes explicit-dispose and GC finalization ordering without `SafeHandle` or per-call reference guards.

RC synchronization only happens when dependencies are created or released. The committed-ref benchmark shows ordinary calls unchanged (1.22 vs 1.21 ns); borrowed-value creation pays the expected state cost (53 vs 161 ns, 72 vs 200 B).

Includes net8/net48 coverage for explicit disposal, finalizer ordering, transitive dependencies, pinned inputs, and Dispose/finalizer races.

Depends on #1244.


irvingoujAtDevolution:
closing — superseded by [#1251](https://github.com/rust-diplomat/diplomat/pull/1251) (native borrow-dependency RC). the stress/lifetime bits here aren't landing separately.

PR #1248, opened 2026-08-07
C++ traits

### LINKED ISSUE #1234: C++ Traits Support
This has become somewhat of a medium priority for Zeromatter, mostly because we're interested in having something akin to interfaces in Python.

tyler-zeromatter:
Closes https://github.com/rust-diplomat/diplomat/issues/1234

Mostly done, with a few TODOs:

- [x] Optional return tests
- [x] Result of optional return tests
- Will probably have to update the macro for these
- [x] Rebase and clean up for review


tyler-zeromatter:
This also fixes a bug I found where callbacks don't properly unwrap optionals inside of the error struct of a result

macro/src/lib.rs
```diff
@@ -138,12 +138,24 @@ fn gen_custom_vtable(custom_trait: &ast::Trait, custom_trait_vtable_type: &Ident
         pub alignment: usize,
     ));
     for m in &custom_trait.methods {
-        // TODO check that this is the right conversion, it might be the wrong direction
         let mut param_types: Vec<syn::Type> = m.params.iter().map(|p| param_ty(&p.ty)).collect();
         let method_name = Ident::new(&format!("run_{}_callback", m.name), Span::call_site());
         let return_tokens = match &m.output_type {
             Some(ret_ty) => {
-                let conv_ret_ty = ret_ty.to_syn();
+                let conv_ret_ty = match ret_ty {
```
Manishearth: question: why does `to_syn` not handle this? The whole point of `to_syn` is to handle this, no?

tyler-zeromatter: It's the same implementation as `gen_custom_function`, which does something similar (see line 356). `to_syn` does not catch when a type is a stdlib `Result` when we require it to be FFI safe.

Manishearth: Can we make `to_syn` more useful here?

tyler-zeromatter: I'm wondering if we can just replace all instances with `.ffi_safe_version().to_syn()`. I mirrored the `gen_custom_function`'s behavior because I was worried I was missing something. But I can try to eliminate the codegen of `DiplomatResult` with `.ffi_safe_version` and see where that gets us

Manishearth: Great!

tyler-zeromatter: Didn't take all that many changes, let me know how this looks

Manishearth (approved):

PR #1251, opened 2026-08-13
dotnet: retain native borrow dependencies

irvingoujAtDevolution:
Smaller replacement for #1246.

Opaque wrappers now retain borrowed-from native handles, so disposing/finalizing a parent only releases its owner ref. Native destruction waits for borrowers and always runs child → parent. Borrowed spans use a sealed reference wrapper with the same parent retention.

Lifecycle operations remain thread-confined.

Tests: `cargo test -p diplomat-tool dotnet`, .NET feature tests on net8/net48, clippy.


Manishearth (approved):

irvingoujAtDevolution:
The [test-dart failure](https://github.com/rust-diplomat/diplomat/actions/runs/31747651159/job/94605928269) is unrelated — `dart format` on the runner's latest stable SDK reformats `test/slice_test.dart`, which is byte-identical to main (this PR has no dart changes at all). Same failure would show up on any branch right now, main's last green run was before the new SDK landed. Everything else is green.

Manishearth (approved):

PR #1256, opened 2026-08-18
Always coerce enums in Demo Gen

### LINKED ISSUE #1255: JS does not handle strings for optional enums
Steps to reproduce:

1. Open https://icu4x.unicode.org/2_2/demo/
2. Scroll to "DateFormatter.formatIso"
3. Fill in the boxes, with the enums set to Short, Column, and Auto

Expected behavior: the code should be

```javascript
icu.DateFormatter.createYmd(icu.Locale
    .fromString("en-US"), icu.DateTimeLength.Short, icu.DateTimeAlignment.Column,
    icu.YearStyle.Auto)
  .formatIso(new icu.IsoDate(2026, 8, 16))
```

and it should produce: "08/16/26"

Actual behavior: the code returned is

```javascript
icu.DateFormatter.createYmd(icu.Locale
    .fromString("en-US"), 'Short', 'Column',
    'Auto')
  .formatIso(new icu.IsoDate(2026, 8, 16))
```

and it produces: "August 16, 2026"

CC @ambiguousname

ambiguousname:
Fixes https://github.com/rust-diplomat/diplomat/issues/1255 

This just forces enums to always use the constructor in methods, since they're always strings within demo_gen.

Need to test within ICU4X, but I believe this should work.

### DIFF
--- a/example/demo_gen/demo/index.mjs
+++ b/example/demo_gen/demo/index.mjs
@@ -11,12 +11,12 @@ const displayOptionalEnum = (out) => out?.value || 'None';
 let termini = Object.assign({
     "FixedDecimalFormatter.formatWrite": {
         func: (selfLocaleName, selfOptionsGroupingStrategy, selfOptionsSomeOtherConfig, valueV) => somelib.FixedDecimalFormatter.tryNew(new somelib.Locale(selfLocaleName), somelib.DataProvider.newStatic(), somelib.FixedDecimalFormatterOptions.fromFields({
-            groupingStrategy: selfOptionsGroupingStrategy,
+            groupingStrategy: new somelib.FixedDecimalGroupingStrategy(selfOptionsGroupingStrategy),
             someOtherConfig: selfOptionsSomeOtherConfig
         })).formatWrite(new somelib.FixedDecimal(valueV)),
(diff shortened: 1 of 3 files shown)


ambiguousname:
The ICU4X build process for web has changed since I last had to work with it, but this seems to work fine in the local generation for diplomat. Someone else feel free to test with the ICU4X demo, or I can take a look some other day.

Manishearth:
Instead of doing this (which gets kinda verbose) can we just have enum _values_ generated as `lib.Enum.Variant`?

ambiguousname:
We could do that, although I opted for this since I believe @robertbastian mentioned this was a solution he didn't want in #896. Happy to go either way on this

Manishearth:
In that case can we fix the underlying JS bug and coerce everywhere?

Either strings are valid as enum values (and we do that everywhere), or we encourage typed values (and demogen produces them)

sffc:
I prefer either coerced strings or constants. 

I don't like calling the constructor with a string, because then you miss out on the ergonomics without gaining the type safety.

robertbastian (changes requested):
How is this different from https://github.com/rust-diplomat/diplomat/pull/1113?

PR #1257, opened 2026-08-18
dotnet: treat borrow leases as transferable claims

irvingoujAtDevolution:
A lease now holds both a lifetime claim and shared/exclusive mode. Returned borrows take over that lease instead of only GC-rooting the source, so overlapping mutation stays blocked until the view is released. Dispose during an active call only drops the wrapper claim.

## Why

Generated wrappers could start a Rust `&mut` while another shared or exclusive use of the same handle was still live. The old keep-alive edge kept the native allocation alive, but it did not keep the source borrowed.

## What this does

- acquire a shared or exclusive lease before every opaque native call
- transfer that lease onto a borrowed / owned-borrowing return
- fail fast with `InvalidOperationException` on overlap
- keep native destruction behind the last claim, so `Dispose()` mid-call does not free the pointer
- keep legacy `manually_disposable` accepted as an ignored compatibility no-op until [#1260](https://github.com/rust-diplomat/diplomat/issues/1260)

## Tests

- `cargo test -p diplomat-tool dotnet::test::`
- `cargo make test-dotnet-feature`


irvingoujAtDevolution:
# Runtime Borrow Safety for Diplomat’s .NET Bindings

## The problem

Diplomat erased Rust’s borrow rules when it generated the .NET API.

~~~rust
fn read(&self);
fn mutate(&mut self);
fn view(&self) -> &Item;
~~~

All three became ordinary C# calls backed by raw pointers:

~~~csharp
var view = owner.View();

Parallel.Invoke(
    () => view.Read(),
    () => owner.Mutate()
);
~~~

On main, both calls could enter Rust at the same time.

The same problem existed without a returned view:

~~~csharp
Parallel.Invoke(
    () => owner.Read(),
    () => owner.Mutate()
);
~~~

Rust would reject these overlapping borrows. C# could not see them.

This was more than a race in application data. Rust code could reallocate, replace, or destroy memory while another call still used a reference into that memory. The result could be undefined behavior instead of a managed exception.

## Attempt 1: hold every borrow until the view dies

The first model transferred the borrow into the returned wrapper:

~~~csharp
var view = owner.View();

view.Read();
owner.Mutate(); // throws while view remains alive

view.Dispose();
owner.Mutate(); // works
~~~

This was safe, but too restrictive.

Rust uses the last actual use of a reference:

~~~rust
let view = owner.view();
view.read();

owner.mutate(); // valid: view is no longer used
~~~

The .NET implementation used the garbage collector as the last-use detector:

~~~text
borrow ends ≈ wrapper finalizer runs
~~~

That means one read could block mutation until explicit disposal or an unpredictable garbage collection.

The model was safe, but it turned normal shared views into long-lived locks.

## Attempt 2: scope every borrowed return

We then considered making every borrowed return a ScopedUse<T>:

~~~csharp
using (var scope = owner.View())
{
    scope.Value.Read();
}

owner.Mutate();
~~~

This made the borrow boundary explicit, but charged every &T call for the hardest &mut T case.

A ref struct also cannot move freely through normal C# code:

~~~csharp
await UseLater(scope);           // cannot cross await
Task.Run(() => scope.Value);     // cannot capture
StoreForLater(scope);            // cannot place on the heap
~~~

That contract is reasonable for an exclusive mutable reference. It is excessive for an ordinary shared view.

## Attempt 3: epoch invalidation alone

An epoch lets mutation invalidate old views:

~~~text
view captures epoch 4
owner mutates
owner epoch becomes 5
view checks 4 != 5 and throws
~~~

This works well for shared views:

~~~csharp
var oldView = owner.View();

owner.Mutate();

oldView.Read();       // InvalidOperationException
var newView = owner.View();
newView.Read();       // valid
~~~

Mutation does not permanently prevent new views. A new view captures the new epoch.

But an epoch alone cannot represent an active &mut T. A mutable reference requires exclusivity for its whole usable lifetime:

~~~csharp
var mutableView = owner.ViewMut();

// owner must remain inaccessible while mutableView can call Rust
mutableView.Mutate();
~~~

Invalidating the mutable view after another call is too late. Both calls may already have entered Rust.

## The hybrid model

The final design treats shared and mutable borrowed returns differently.

### Every native call gets an operation lease

~~~csharp
using (BorrowLease<Raw.Owner> lease = BorrowShared())
{
    return Raw.Owner.Read(lease.Ptr);
}
~~~

Mutable calls request an exclusive lease:

~~~csharp
using (BorrowLease<Raw.Owner> lease = BorrowExclusive())
{
    return Raw.Owner.Mutate(lease.Ptr);
}
~~~

The shared state uses atomic transitions:

~~~text
shared:    n >= 0  → n + 1
exclusive: 0       → -1
~~~

Therefore:

~~~text
shared + shared    allowed
shared + exclusive rejected
exclusive + shared rejected
exclusive + exclusive rejected
~~~

The losing call throws InvalidOperationException before it reaches Rust.

### Shared views use versioned dependencies

A shared view does not keep its source borrowed forever:

~~~csharp
var view = owner.View();

view.Read();       // takes a temporary shared source lease
owner.Mutate();    // allowed after Read returns
view.Read();       // throws: source version changed

var fresh = owner.View();
fresh.Read();      // valid
~~~

Before every native call, the view:

~~~text
checks source version
→ acquires source shared lease
→ checks source version again
→ enters Rust
~~~

The second check closes the race between validation and lease acquisition.

### Mutable views use deterministic scopes

A mutable borrowed return keeps its source exclusively borrowed:

~~~csharp
using (var scope = owner.ViewMut())
{
    var view = scope.Value;

    view.Mutate(); // allowed
    owner.Read();  // InvalidOperationException
    owner.Mutate();// InvalidOperationException
}

owner.Mutate();    // allowed
~~~

ScopedUse<T>.Dispose() ends that exclusive borrow. If the caller forgets to dispose it, the wrapper finalizer remains a safety backstop.

### Disposal cannot destroy an active call

Each operation lease also holds a lifetime claim:

~~~text
call starts → retain claim
Dispose     → release wrapper claim
call ends   → release operation claim
refcount 0  → run native destructor
~~~

Calling Dispose() during a native operation invalidates the managed wrapper immediately, but native destruction waits until the active call returns.

## Result

The hybrid model preserves the two useful behaviors:

~~~text
shared view:
    easy to use
    mutation allowed after each call
    stale use throws

mutable view:
    explicit scope
    exclusive across threads
    source unlocked deterministically
~~~

For idiomatic generated APIs, conflicting mutable borrows now fail before crossing the foreign function interface. Active calls also keep native memory alive during concurrent disposal.

Concurrent shared calls remain allowed. Whether a Rust type is actually Sync is a separate cross-thread contract and belongs in its own PR.

---

*LLM Generated, Irving Audited/Edited Content*

irvingoujAtDevolution:
After thinking this through, I've decided to remove opt-in disposal.

The opt-in made sense as an API choice, but it limits our ability to enforce Rust's borrowing rules consistently at runtime. We tried several approaches above, and I genuinely think the runtime model in [a0772e3a](https://github.com/rust-diplomat/diplomat/commit/a0772e3af7aa4e5abebffcc2a63a6346623607bb) is the best we've had so far.

The follow-up that makes every opaque wrapper disposable is still local and hasn't been pushed yet.

_Written by Irving; grammar and wording edited with an LLM._

## What this means for .NET users

Before this follow-up, `using (...)` worked only when the Rust type opted into `manually_disposable`; otherwise, the caller had to wait for finalization.

With this change, every opaque wrapper supports the same pattern:

```csharp
// This did not compile before because the type did not opt in.
using (DefaultDropProbe probe = DefaultDropProbe.Create())
{
    // Use the owned value.
}
```

Callers do not need to scope shared views for safety, and mutation is still allowed after creating one:

```csharp
RcSource source = RcSource.Create(42);
RcSource view = source.View();

view.Id();
source.PingMutable(); // Allowed, but invalidates view.
view.Id();            // Throws InvalidOperationException.
```

A mutable view is different because it holds an exclusive borrow, so the source cannot be read or mutated until that scope ends:

```csharp
using (ScopedUse<RcSource> mutable = source.ViewMut())
{
    mutable.Value.Id();
    mutable.Value.PingMutable();

    // source.Id() and source.PingMutable() throw here.
}

source.Id();
source.PingMutable();
```

The same rule applies to mutable-after-mutable calls: a second mutable view fails while the first scope is active and succeeds after it ends.

Conflicting calls from different threads now fail before Rust runs:

```csharp
Task<ulong> read = Task.Run(source.Id);
Task<bool> write = Task.Run(source.PingMutable);

Task.WaitAll(read, write);
```

If those calls overlap, the conflicting call throws `InvalidOperationException` instead of violating Rust's aliasing rules.

`Dispose()` is not treated as a mutable borrow; it is safe during an active call because native destruction waits for that call to finish:

```csharp
Task<bool> call = Task.Run(source.PingMutable);
source.Dispose();

call.Wait();
```

If the call starts first, native destruction waits; if `Dispose()` starts first, the call throws `ObjectDisposedException`.

Concurrent shared calls remain allowed. Cross-thread `Sync` enforcement is separate from runtime borrow checking and remains follow-up work.

Finalization remains a fallback, but explicit scopes make mutable borrows and native lifetimes predictable.

_Generated with an LLM; reviewed and edited by Irving._

irvingoujAtDevolution:
Forgot to mention: the most important problem with the current model is that concurrent read + mutable calls, or mutable + mutable calls, can violate Rust's aliasing rules and cause undefined behavior.

For example:

```csharp
Task.WaitAll(
    Task.Run(source.Id),
    Task.Run(source.PingMutable)
);

Task.WaitAll(
    Task.Run(source.PingMutable),
    Task.Run(source.PingMutable)
);
```

If these calls overlap, the current model allows both to enter Rust. That can lead to memory corruption and concurrency bugs that are extremely hard to reproduce.

We'd much rather get a clear runtime exception. As we increasingly work with LLM-generated code, anything that can corrupt state eventually will.

_Written by Irving; grammar and technical claims audited with an LLM._

irvingoujAtDevolution:
Sorry team, I thought we'd be releasing later and only just saw that 0.16.0 is already out. I was thinking we could undo the removal of `manually_disposable`. Let me know what you think—maybe we should ignore the attribute for now instead of removing it, since the new version is already released.

*Irving Wrote, LLM Audited.*

Manishearth:
Ignore it for now and document it as unused. We can remove it in the next release, and can try to have a tracking issue to make sure we do that.

irvingoujAtDevolution:
Kept `manually_disposable` as an ignored compatibility no-op in [0ab31a1d](https://github.com/rust-diplomat/diplomat/commit/0ab31a1ddd5f45ae5b214f54f863136b68842f8a), added the docs, and opened [#1260](https://github.com/rust-diplomat/diplomat/issues/1260) to remove it in the next breaking release.

---

> **LLM Wrote, Unaudited — Read With a Grain of Salt.**

Manishearth (approved):

PR #1258, opened 2026-08-19
CONTRIBUTING.md

tyler-zeromatter:
Wrote something quickly, let me know how this looks.

PR #1254, opened 2026-08-14
ci: pin Dart SDK to 3.12.2

irvingoujAtDevolution:
## What

Pin `dart-lang/setup-dart` to **3.12.2** in the `test-dart` job.

## Why

`setup-dart` was unpinned, so CI floated to stable **3.13.0**. That formatter rewrites `feature_tests/dart/test/slice_test.dart`, and `dart format --set-exit-if-changed` fails `test-dart` on every PR (including pure .NET ones). Main has been red since [#1251](https://github.com/rust-diplomat/diplomat/pull/1251) for this reason.

3.12.2 is the last stable that leaves the tree format-clean as-is.

## Test

- local: `dart format --set-exit-if-changed feature_tests/dart` clean under 3.12.2; dirty under 3.13.0

PR #1259, opened 2026-08-20
Publish core/macro/tool 0.16.1

Manishearth:
Changes since 0.16.1:

 - 8e2b7a3a Publish core/macro/tool 0.16.1
 - 4cc68edd  CONTRIBUTING.md (#1258)
 - e75e0643 C++ traits (#1248)
 - 60cd7b91 dotnet: dispose borrowed-span edges on oversize construction (#1253)
 - 3738d61e ci: pin Dart SDK to 3.12.2 (#1254)
 - 6610ee15 dotnet: retain native borrow dependencies (#1251)
 - 48fa3f3a dotnet: make IDisposable opt-in for opaques (#1244)
 - 08e1f64d Upgrade Clippy MSRV to 1.88 (#1249)
 - e7d95fe8 Add `Error` and `Exception` to enums/classes which are errors/exceptions in Dart (#1220)
 - 242e3cb3 Validate JS arguments (#902)
 - 7b6ac1c5 Improve some diplomat-runtime docs (#1242)
 - 702b1b22 runtime should be 0.15.2 (#1237)
 - 5655be79 Switch diplomat-runtime to 0.15.1 (#1236)
 - 7de34935 Expose tuples entry in book (#1233)

The goal here is to get https://github.com/rust-diplomat/diplomat/pull/1220 and  #902 into the release. There are also dotnet fixes that are nice to get out.

This does include one larger feature: C++ trait support. I think that's okay.

Verified that there are no semver-checks issues in core/tool.

This also moves diplomat-runtime to 0.16.0 (already published) but adds a note that we plan to keep it there. As @tyler-zeromatter and I have been discussing, we're going to treat 0.16.x as the frozen stream. This is just to keep the repo state less confusing for the future.

### DIFF
--- a/runtime/Cargo.toml
+++ b/runtime/Cargo.toml
@@ -3,7 +3,7 @@ name = "diplomat-runtime"
 description = "Common runtime utilities used by diplomat codegen"
 # We are working on decoupling diplomat-runtime's version from the rest
 # https://github.com/rust-diplomat/diplomat/issues/958
-version = "0.15.2"
+version = "0.16.0"
 rust-version.workspace = true
 authors.workspace = true
 categories.workspace = true
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -365,7 +365,7 @@ checksum = "7a81dae078cea95a014a339291cec439d2f232ebe854a9d672b796c6afafa9b7"
 
 [[package]]
 name = "diplomat"
-version = "0.16.0"
+version = "0.16.1"
 dependencies = [
  "diplomat_core",
  "insta",
@@ -377,7 +377,7 @@ dependencies = [
 
 [[package]]
 name = "diplomat-example"
-version = "0.16.0"
+version = "0.16.1"
 dependencies = [
  "criterion",
  "diplomat",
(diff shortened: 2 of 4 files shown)


tyler-zeromatter:
I think the `0.16.1` commits got dropped on push, I can only see the clippy fixes right now in the diff

Manishearth:
Oh it's because I reordered them and the bookmark moved. I'll fix when I'm back at a computer (in an hour unfortunately)

tyler-zeromatter:
No rush on my end!

runtime/Cargo.toml
```diff
@@ -3,7 +3,7 @@ name = "diplomat-runtime"
 description = "Common runtime utilities used by diplomat codegen"
 # We are working on decoupling diplomat-runtime's version from the rest
 # https://github.com/rust-diplomat/diplomat/issues/958
-version = "0.15.2"
+version = "0.16.0"
```
tyler-zeromatter, line 6: I think there's already a 0.16.0 on crates.io: https://crates.io/crates/diplomat-runtime/0.16.0

This should also be 0.16.1

tyler-zeromatter (commented):
LGTM with one small issue.

runtime/Cargo.toml
Manishearth: No, this isn't a publish. I documented this in the PR body.

PR #1261, opened 2026-08-24
Nanobind traits

tyler-zeromatter:
TODOs:
- [x] Do not use std::exception on failure to cast (the panic is not revealed this way)
	- Maybe somehow set Python exceptions?
	- Tests for returning the wrong value in a trait
- [x] Fix memory leak issues
- [x] Trait python stubs
- [x] Expand tests
- [x] Rebase and clean up
- [x] C++ and Nanobind docs


tyler-zeromatter:
Still need to write the C++/Nanobind trait docs, but this is ready for review on everything else.

Manishearth (approved):
Reviewing hir and cpp code, you can decide what you want to do for the nanobind code (if you want Walter's approval, etc).

Is the ffi_error stuff already being enforced?

tyler-zeromatter:
Yes, FFI error stuff is already enforced, that was added to the HIR validation pass in the previous FFI error PR. This just also updates enforcement to disallow unit types on fallible traits

tool/src/nanobind/formatter.rs
```diff
@@ -131,4 +142,122 @@ impl<'tcx> PyFormatter<'tcx> {
...
+                    format!("list[{}]", self.symbol_to_python_type(st.id().into())).into()
+                }
+                _ => unreachable!("Unknown AST/HIR variant: {sl:?}"),
+            },
+            crate::hir::Type::Callback(cb) => {
+                let ret =
+                    diplomat_core::hir::CallbackInstantiationFunctionality::get_output_type(cb)
+                        .expect("Could not get output type.");
+
+                let mut params = Vec::new();
```
walter-zeromatter: generally speaking I'd encourage the use of the iterator transformers pattern over the `mut vec + for loop` pattern, just because it allows the correct number of elements to be reserved.

tool/templates/cpp/trait_decl.h.jinja
```diff
@@ -14,7 +14,10 @@ namespace {{lib_name}} {
 {% endif -%}
 {% if let Some(deprecated) = deprecated %}[[deprecated("{{ deprecated }}")]] {% endif -%} class {{trait_name_unnamespaced}} {
     private:
-    static void Destroy(const void* data);
+    static void Destroy(const void* data) {
```
walter-zeromatter, line 17: just curios, why the move to inline?

walter-zeromatter (approved):
a few minor style/micro perf comments but structurally this seems pretty sound to me.

tool/templates/cpp/trait_decl.h.jinja
tyler-zeromatter: I was getting compiler errors about double definitions

tyler-zeromatter:
Will merge once this passes CI

PR #1264, opened 2026-08-26
HIR Location Errors

tyler-zeromatter:
Handles https://github.com/rust-diplomat/diplomat/issues/111 in HIR lowering errors.

This adds a new LoweringError type: `InvalidField`, for testing the new printer. It also adds new `LoweringReport` and `hir::defs::Ident` structs for including location information when printing errors (but it functions without location information, as seen in the updated snapshots)

TODOs:
- [x] Document breaking API changes to core on approval (for the relnotes if this gets approved/merged)

### DIFF
--- a/core/src/ast/logging.rs
+++ b/core/src/ast/logging.rs
@@ -87,11 +87,33 @@ fn evaluate_bytes(sp: &super::idents::Span, src: &str) -> std::ops::Range<usize>
     }
 }
 
-pub(crate) fn create_report(report: AstReport) -> ! {
-    use std::io::Write;
-    let mut out = WRITER.read().unwrap()();
+pub trait WriteReport {
+    fn write_fmt(&mut self, args: std::fmt::Arguments<'_>) -> Result<(), String>;
+    fn flush(&mut self) -> Result<(), String>;
+}
 
-    let span = report.primary_loc;
+impl WriteReport for Box<dyn std::io::Write> {
+    fn write_fmt(&mut self, args: std::fmt::Arguments<'_>) -> Result<(), String> {
+        std::io::Write::write_fmt(self, args).map_err(|e| e.to_string())
+    }
+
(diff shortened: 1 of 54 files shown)


tyler-zeromatter:
The span based test results are located in https://github.com/rust-diplomat/diplomat/pull/1264/changes#diff-bcdd02c41aeee44274091a041205d0fc31a6323e5653f661007b092fe4b5b700

core/src/hir/lowering.rs
```diff
@@ -20,6 +20,9 @@ use strck::IntoCk;
 #[derive(Debug)]
 #[non_exhaustive]
 pub enum LoweringError {
+    /// A field type is invalid in some way. The contained [`LoweringError`] has more specifics,
+    /// this is to just demarcate where the field is for the error.
+    InvalidField(super::defs::Ident, Box<LoweringError>),
```
Manishearth: issue: I don't think we should nest LoweringErrors? Fine to have an internal string type there but let's not nest

core/src/ast/logging.rs
```diff
@@ -112,7 +112,19 @@ impl WriteReport for &mut std::fmt::Formatter<'_> {
...
+    }
+    fn flush(&mut self) -> Result<(), String> {
+        Ok(())
+    }
+}
+
+pub fn write_report(report : &AstReport, mut out : impl WriteReport, 
+    // Only used when pretty-print feature is enabled:
+    #[allow(unused)]
+    force_ugly : bool) -> Result<(), String> {
```
Manishearth: nit: can we use a  PrettyPrint::ForceUgly enum here to avoid lots of booleans everywhere

core/src/hir/lowering.rs
tyler-zeromatter: Let me know how this revised structure works

Manishearth (approved):

PR #1265, opened 2026-08-28
Tool Location Errors

### LINKED ISSUE #111: Include file/line info in Diplomat errors

tyler-zeromatter:
Closes #111 

TODO:
- [x] Document breaking API changes


core/src/hir/defs.rs
```diff
@@ -318,6 +328,15 @@ impl<'tcx> TypeDef<'tcx> {
         }
     }
 
+    pub fn loc_name(&self) -> &'tcx Ident {
```
Manishearth: not sure of the name of this. Can this be a bit more self-documenting (and have docs)?

Manishearth (approved):

PR #1267, opened 2026-09-01
dotnet: enforce deterministic disposal for retained borrows

irvingoujAtDevolution:
Makes `IDisposable` opt-in again for .NET opaques, while requiring `#[diplomat::attr(dotnet, manually_disposable)]` where deterministic cleanup is needed to preserve Rust borrow semantics.

- drops `ScopedUse`; `&mut T` / `Option<&mut T>` return ordinary wrappers
- checks mutable views, owned borrowing returns, pinned managed inputs, and transitive retention
- keeps finalizers as fallback cleanup
- reports the Rust type, method, source, borrow mode, and blocked calls when the attribute is missing

This also keeps the native lifetime counter and borrow ledger together in `RustHandle`, so destructors run before dependency edges and pins are released.

Validation:
- `cargo make gen-dotnet`
- `cargo fmt --all -- --check`
- `cargo check -p diplomat-tool --tests`


irvingoujAtDevolution:
# Dotnet's Dispose Problem


## Background

In Rust, the compiler enforces a few rules on how to use pointers and references safely. The compiler's guard cannot help us when we share data across the FFI. In particular, we are trying to solve and are able to solve the following issues today:
    a. The single mutability guarantee
    b. The ownership and lifetime rules

## Rust's Guarantees (You can skip this section if you are familiar with Rust)

1. Exclusive Mutability Guarantee: Safe Rust allows either any number of active shared references or one active mutable reference to the same value, but not both at the same time.

Example:
```rust
fn main() {
    let mut s = String::from("hello");

    let r1 = &s; // no problem
    let r2 = &s; // no problem
    println!("{r1} and {r2}");
    // Variables r1 and r2 will not be used after this point.

    let r3 = &mut s; // no problem
    println!("{r3}");
}
```

2. Ownership and Lifetime Guarantees: Safe Rust ensures that a reference cannot outlive the value it borrows. If a child contains a reference to its parent, the parent cannot be dropped while the child remains usable.

Example:  
```rust
struct Child<'a> {
    parent: &'a str,
}

fn main() {
    let child;

    {
        let parent = String::from("parent");
        child = Child { parent: &parent };
    } // Error: parent does not live long enough

    println!("{}", child.parent);
}
```


## These guarantees DO NOT exist when values cross the FFI boundary

Diplomat lets you write regular Rust functions for FFI and generate bindings for other languages (the alternative is to write `extern "C"` functions manually).
However, safety is NOT guaranteed. As for Dotnet, we try our best to make it as performant and as safe as possible, allowing Dotnet developers or AI-generated code to be free from undefined behavior, corrupted memory, and concurrency bugs. This is subject to Dotnet's language design and runtime capabilities. Self-evidently, we cannot prevent everything at compile time, and the goal is to help developers be aware of the issues as early as possible at runtime.


## The Ownership and Lifetime Guarantees Across FFI

A value (referred to as the child) can borrow from another value (referred to as the parent) across the FFI boundary. For example:
```rust

pub struct Parent {
    value: String,
}

pub struct Child<'a> {
    parent: &'a Parent,
}


impl Parent {
    pub fn new() -> Self {
        Self { value: "parent".to_owned() }
    }

    pub fn create_child<'a>(&'a self) -> Child<'a> {
        Child { parent: self }
    }
}
``` 

On the Dotnet side, the following usage pattern is very common:
```csharp

public class Program
{

    public static Child CreateChild()
    {
        var parent = Parent.New();
        return parent.CreateChild();
    }

    public static void Main()
    {
        var child = CreateChild();
        Console.WriteLine(child);
    }
}
```

This pattern will not compile in Rust, but is acceptable in Dotnet. We do not want to regulate, nor are we able to enforce, how Dotnet uses values. The goal here is to glue them together so the Dotnet side does not need to worry about Rust's strict ownership and lifetime rules, and the Rust side can be confident that the lifetime rules are respected.

### Solution

### Reference Counting 

We introduce an RC-like mechanism. See RustHandle.cs.
When the above example is executed, the underlying Rust handle will perform reference counting on the parent and create a `BorrowToken` for the child to hold, and the parent's reference count will be increased automatically.

```csharp
public Bar Bar
{
    get
    {
        unsafe
        {
            using (BorrowLease<Raw.Foo> selfLease = Lease(BorrowKind.Shared))
            {
                Raw.Bar* result = Raw.Foo.GetBar(selfLease.Ptr);
                GC.KeepAlive(this);
                return new Bar(result, selfLease);
            }
        }
    }
}
```

There are a couple of things going on here. `Lease` will internally increase the reference counter by 1, and selfLease is essentially an object containing the `RustHandle<Foo>`. It will be passed to the `Bar` constructor and stored internally. This way, the child will keep the parent's underlying Rust handle alive as long as the child is alive. 

In some cases where `manually_disposable` is required (covered later), it is also safe to call `Foo.Dispose()` or to execute `using (var foo = new Foo()) { ... }; useBar(bar)`, because internally, the destructor (which is a Rust function that takes Foo's handle) of `Foo` will ONLY be executed if releasing the reference makes the count zero.


### Mutability

We try to prevent this from happening:

```rust
struct Source {
    value: u64,
}

impl Source {
    fn id(&self) -> u64 {
        self.value
    }

    fn view_mut(&mut self) -> &mut Self {
        self
    }
}

fn main() {
    let mut source = Source { value: 7 };
    let mutable_view = source.view_mut();

    source.id(); // Error: source is still mutably borrowed. This is perfectly blocked by Rust's compiler.

    mutable_view.value = 8;
}
```
Dotnet has nothing to prevent this from happening. The code below is perfectly executable Dotnet code. It is fine in this example, but there are a million ways this could lead to undefined behavior if not handled carefully.

```csharp
using RcSource source = RcSource.Create(7);

RcSource mutableView = source.ViewMut();

source.Id(); // <--- We hope this gets banned
```
We could easily block any read after a mutable borrow is created, but this way, we could never use the value of `source` ever again until `mutableView` gets disposed or collected.

Hence, we will enforce that the type returned as a mutable view uses the `manually_disposable` attribute, which will later allow us to drop it explicitly after use:
```csharp
using RcSource source = RcSource.Create(7);

using (RcSource mutableView = source.ViewMut())
{
    source.Id(); // <--- This is now banned at runtime
}

source.Id(); // <--- But this is perfectly fine
```


## Manually Disposable Attribute

I previously temporarily removed the `manually_disposable` attribute, and then I learned that this is a bad idea.

As per Benoit:
    Implementing IDisposable on every opaque is not ergonomically neutral: it strongly communicates that every returned instance should normally be deterministically disposed, and we have lints/analyzers to enforce that expectation. A lightweight opaque should not impose deterministic-disposal semantics on its C# users just because its implementation happens to be boxed across the FFI.


However, as previously mentioned, since we now use IDisposable to express Rust's lifetime semantics at runtime in Dotnet, there are a few other cases where this should hold.

1. Read a value from the source when there are active mutable borrows. See the "Mutability" section above.

2. Mutate the source when an owned return shares a reference from the source.

In Rust, the following code cannot compile:
```rust
struct Source {
    value: u64,
}

struct Dependent<'a> {
    source: &'a Source,
}

impl Source {
    fn dependent(&self) -> Box<Dependent<'_>> {
        Box::new(Dependent { source: self })
    }

    fn set_value(&mut self, value: u64) {
        self.value = value;
    }
}

impl Dependent<'_> {
    fn source_value(&self) -> u64 {
        self.source.value
    }
}

fn main() {
    let mut source = Source { value: 7 };
    let dependent = source.dependent();

    // Error: dependent still holds a shared borrow of source.
    source.set_value(8);

    println!("{}", dependent.source_value());
}
```

The `Box<Dependent<'_>>` is different from `&Source` because `Box<Dependent<'_>>` also owns memory that is more than just a reference to `Source`. For `&Source`, we could easily invalidate it because we do not need to run any destructor on it, whereas `Box<Dependent<'_>>` needs to be explicitly freed/dropped here.

```csharp
using RcSource source = RcSource.Create(7);

using (RcDependent dependent = source.MakeDependent())
{
    useDependent(dependent);

    // Error: dependent still holds a shared borrow of source.
    source.PingMutable();
}

// Now we can safely use source mutably again.
source.PingMutable();

```

3. A reference to Dotnet-pinned memory.  

#### Background
We allow directly have Rust struct borrows a reference from Dotnet pinned memory, a real example here:
```csharp
public static PinnedRcSource Create(ReadOnlyMemory<byte> data)
{
    unsafe
    {
        DiplomatPinnedMemory? dataPin = null;
        try
        {
            dataPin = DiplomatPinnedMemory.Pin(data);
            Raw.PinnedRcSource* result = Raw.PinnedRcSource.Create(
                new DiplomatSliceU8
                {
                    Ptr = (byte*)dataPin.Pointer,
                    Len = (nuint)data.Length,
                }
            );
            return new PinnedRcSource(result, dataPin);
        }
        catch
        {
            dataPin?.Dispose();
            throw;
        }
    }
}
```

and on dispose, we have
```csharp
parser.Dispose()
    -> Drop the Rust-owned Parser
    -> dataPin.Dispose()
```

for more, you can see `DiplomatPinnedMemory` class.



Dotnet allows memory to be pinned, which is a really good feature. When we pass pinned memory from Dotnet to a Rust struct, the glue layer should be aware that this piece of memory must be unpinned when the Rust-owned struct gets dropped. 
Example:
```rust
struct Parser<'a> {
    data: &'a [u8],
}

impl Parser<'_> {
    fn parse<'a>(data: &'a [u8]) -> Box<Parser<'a>> {
        Box::new(Parser { data })
    }

    fn first(&self) -> u8 {
        self.data[0]
    }
}
```

On the Dotnet side, the generated wrapper will pin the managed memory and store the pin together with the Rust-owned `Parser`:

```csharp
byte[] data = { 1, 2, 3 };

using (Parser parser = Parser.Parse(data))
{
    Console.WriteLine(parser.First());
    // data stays pinned while parser is alive.
}

// Parser's Rust destructor has run, so the managed memory is now unpinned.
```

4. For all the cases above, plus any case where `manually_disposable` is explicitly declared, every opaque type that transitively retains a type requiring `Dispose()` must also implement `IDisposable`.
Example: 
```rust
#[manually_disposable]
struct Source(Middle);

#[manually_disposable] // Required because the returned &Middle retains Source.
struct Middle(Outer);

#[manually_disposable] // Required because the returned &Outer retains Middle.
struct Outer;

Source::middle(&self) -> &Middle;
Middle::outer(&self) -> &Outer;
```


## Notes
### Why RC is still needed here? .NET already has a garbage collector
The earlier approach of keeping parent objects alive with edge references was fine before we introduced `IDisposable` to model RustHandle lifetimes. It is no longer sufficient once `IDisposable` is part of the runtime contract, because a parent can be disposed while a dependent still needs to access the underlying Rust value.

Example:
```csharp
RcSource source = RcSource.Create(1);
RcDependent dependent = source.MakeDependent();
source.Dispose();
dependent.SourceId(); // We still want this to remain valid, so the source must stay alive.
```

Second, the .NET finalizer does not guarantee cleanup order. If both the parent and the child become unreachable at the same time, the GC may finalize them in either order, regardless of their dependency relationship.

Example:
```csharp
public class Program
{
    public static void Main()
    {
        RcSource source = RcSource.Create(1);
        RcDependent dependent = source.MakeDependent();
    }
}
```

The .NET garbage collector could invoke the source destructor first. If the dependent has a custom `Drop` implementation like this:
```rust
impl Drop for RcDependent {
    fn drop(&mut self) {
        self.source.id(); // use-after-free
    }
}
```
then we would have a use-after-free bug. Rust guarantees that this cannot happen, so we must enforce the same lifetime relationship across the FFI boundary.

CBenoit:
Thank you for the write-up @irvingoujAtDevolution, I'm on board with the approach!

PR #1252, opened 2026-08-13
nanobind: support None for char32_t

floze-the-genius:
Closes #1126.

## Summary

- honor nanobind's `accepts_none` flag in the custom `char32_t` caster
- convert an allowed `None` value to the default null character (`U+0000`)
- reject non-string and multi-code-point inputs before reading the character
- cover the struct-constructor path with a regression test

## Tests

- `uvx --with ./feature_tests/nanobind --refresh-package somelib pytest feature_tests/nanobind/test/ -q` (36 passed, clean wheel rebuild)
- `cargo test -p diplomat-tool nanobind` (4 passed)
- `cargo fmt --all -- --check`

PR #1269, opened 2026-09-02
fix: reject None in the nanobind char32_t caster instead of crashing

### LINKED ISSUE #1126: Nanobind `char32_t` conversion does not support `None`
Currently if you pass in `None` where Nanobind expects a `char32_t` (I've encountered this bug happening calling a struct constructor), our custom conversion code isn't able to handle it.

I don't believe we use `DiplomatChar` anywhere in Zeromatter, but this would certainly be an issue for ICU4X python support.

Jorge-Polanco-Roque:
Fixes #1126

## Problem

Passing `None` where nanobind expects a `char32_t` crashes the interpreter. `type_caster<char32_t>::from_python` does not handle `None`: `PyUnicode_ReadChar` fails and leaves a dangling Python error, but the `if (!value)` guard misses it (the failure sentinel is `(Py_UCS4)-1`, not `0`), so `from_python` returns `true` with a garbage char. That garbage reaches the callee and — as the feature test shows — panics across the C ABI, aborting the process (SIGABRT).

## Fix

Reject `None` up front (`if (src.is_none()) return false;`) so nanobind raises a clean `TypeError` instead. Applied to the template (`tool/templates/nanobind/common.h.jinja`) and its checked-in generated copy (`feature_tests/nanobind/src/include/diplomat_nanobind_common.hpp`).

## Testing

Added `test_char_rejects_none`: without the fix the process aborts, with it the call raises `TypeError` and the interpreter stays healthy. Full nanobind feature-test suite passes (37 passed).

### DIFF
--- a/feature_tests/nanobind/src/include/diplomat_nanobind_common.hpp
+++ b/feature_tests/nanobind/src/include/diplomat_nanobind_common.hpp
@@ -34,6 +34,12 @@ namespace nanobind::detail
 
         bool from_python(handle src, uint8_t, cleanup_list *) noexcept
         {
+            // None is not a str: reject it cleanly so an optional char (or the next
+            // overload) is handled, instead of leaving a dangling Python error.
+            if (src.is_none())
+            {
+                return false;
+            }
             value = PyUnicode_ReadChar(src.ptr(), 0);
             if (!value)
             {
--- a/feature_tests/nanobind/src/include/diplomat_nanobind_common.hpp
+++ b/feature_tests/nanobind/src/include/diplomat_nanobind_common.hpp
@@ -34,11 +34,13 @@ namespace nanobind::detail
 
         bool from_python(handle src, uint8_t, cleanup_list *) noexcept
         {
-            // None is not a str: reject it cleanly so an optional char (or the next
-            // overload) is handled, instead of leaving a dangling Python error.
+            // None maps to the default-initialized char '\0', mirroring C struct
+            // zero-initialization, instead of leaving a dangling Python error.
             if (src.is_none())
             {
-                return false;
+                value = U'\0';
+                size = 1;
+                return true;
             }
             value = PyUnicode_ReadChar(src.ptr(), 0);
             if (!value)
@@ -77,7 +79,9 @@ namespace nanobind::detail
         template <typename T_>
         NB_INLINE bool can_cast() const noexcept
         {
-            return (value && size == 1);
+            // size == 1 means exactly one code point was read (incl. '\0' from None);
+            // the read-failure sentinel leaves size == 0, so it is rejected here.
+            return (size == 1);
         }
 
         explicit operator char32_t()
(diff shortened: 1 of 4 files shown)


tyler-zeromatter:
Taking a look now

tyler-zeromatter (changes requested):
Generally looks good, but with one issue.

I'm worried about the case where we don't have a `#[diplomat::attr(auto, constructor)]` for a struct that contains `char32_t`, since the Nanobind backend will create a struct that automatically casts all struct members to none (see `BigStructWithStuff_binding.cpp` for an example). The constructor then, would fail. I'd like to see a test case that handles that scenario (maybe by creating a new struct that holds `DiplomatChar`).

tyler-zeromatter:
Actually, I think https://github.com/rust-diplomat/diplomat/pull/1252 has the test cases I have in mind, so feel free to use those as an example

Jorge-Polanco-Roque:
Thanks! Added a test for exactly that scenario in `feature_tests/nanobind/test/test_structs.py` (`test_struct_char_constructor_rejects_none`).

It turned out the case you described already exists in the fixtures: `PrimitiveStruct` has a `DiplomatChar` field (`b`) and no `#[diplomat::attr(auto, constructor)]`, so its generated binding is precisely the situation you flagged — an auto-generated positional constructor that marks every member `.none()`, routing `None` straight into the `char32_t` caster:

```cpp
// PrimitiveStruct_binding.cpp
.def(nb::init<float, bool, char32_t, int64_t, intptr_t, uint8_t>(), "x"_a.none(), "a"_a.none(), "b"_a.none(), ...)
```

The test constructs `PrimitiveStruct` through that `.none()` constructor with a valid char (asserts it round-trips) and with `None` in the char position (asserts `TypeError`). I reused the existing struct rather than adding a new one so there's no cross-backend regen churn — but happy to add a dedicated struct instead if you'd prefer it spelled out explicitly. Verified locally: `test_structs.py` + `test_option.py` pass (8/8).

tyler-zeromatter (changes requested):
The test looks good, but I would like to see `somelib.PrimitiveStruct(1.0, True, None, 0, 0, 0)` pass without raising an exception.

Nanobind's behavior roughly mirrors how C structs work: if you initialize a struct with no values in C, then the struct is initialized to its "default" values. The same should apply to `char32_t` fields: it should store the `\0` character if the value is `None`.

Jorge-Polanco-Roque:
Thanks again for the review. I reworked it per your feedback: instead of rejecting `None`, a `None` for a `char32_t` field now default-initializes to `U'\0'` (mirroring how C structs zero-init), so `somelib.PrimitiveStruct(1.0, True, None, 0, 0, 0)` passes without raising, which is exactly the case you asked for. I also added a struct-member test covering it, and the nanobind suite is green. Whenever you have a moment for another look, much appreciated!

tyler-zeromatter:
Taking a look again, thanks for your patience

tyler-zeromatter (approved):
LGTM, if tests pass I will merge.

PR #1272, opened 2026-09-08
Reproduce unstable annotation in bindings

robertbastian:
Avoids things like https://github.com/unicode-org/icu4x/pull/8461 in the future.

### DIFF
--- a/core/src/hir/attrs.rs
+++ b/core/src/hir/attrs.rs
@@ -1734,6 +1734,7 @@ impl AttributeValidator for BasicAttributeValidator {
     }
     fn is_unstable(&self, cfg: &syn::Attribute) -> bool {
         self.semver_unstable_features.iter().any(|f| {
+            // TODO(#1274): Improve this heuristic
             cfg.to_token_stream()
                 .to_string()
                 .contains(&format!("feature = {f:?}"))
--- a/core/src/hir/attrs.rs
+++ b/core/src/hir/attrs.rs
@@ -463,11 +463,7 @@ impl Attrs {
 
         this.deprecated = ast.deprecated.clone();
 
-        this.unstable |= ast.cfg.iter().any(|x| {
-            x.to_token_stream()
-                .to_string()
-                .contains("feature = \"unstable\"")
-        });
+        this.unstable |= ast.cfg.iter().any(|x| validator.is_unstable(x));
 
         let support = validator.attrs_supported();
         let backend = validator.primary_name();
@@ -1591,6 +1587,8 @@ pub trait AttributeValidator {
     fn validate(&self, attrs: &Attrs, context: AttributeContext, errors: &mut ErrorStore) {
         attrs.validate(self, context, errors)
     }
+
+    fn is_unstable(&self, feature: &syn::Attribute) -> bool;
 }
 
 /// A basic attribute validator
@@ -1608,6 +1606,8 @@ pub struct BasicAttributeValidator {
     pub is_name_value: Option<Box<dyn Fn(&str, &str) -> bool>>,
     /// The features enabled.
     pub features_enabled: HashSet<String>,
+    /// The unstable features.
+    pub semver_unstable_features: HashSet<String>,
 }
 
 impl BasicAttributeValidator {
@@ -1731,6 +1731,13 @@ impl AttributeValidator for BasicAttributeValidator {
     }
     fn attrs_supported(&self) -> BackendAttrSupport {
         self.support
+    }
+    fn is_unstable(&self, cfg: &syn::Attribute) -> bool {
+        self.semver_unstable_features.iter().any(|f| {
+            cfg.to_token_stream()
+                .to_string()
+                .contains(&format!("feature = {f:?}"))
+        })
     }
 }
 
--- a/core/src/hir/attrs.rs
+++ b/core/src/hir/attrs.rs
@@ -27,6 +27,8 @@ pub struct Attrs {
     ///
     /// This attribute is always inherited except to variants
     pub disable: bool,
+    /// Whether this item is `cfg(feature = "unstable")`
+    pub unstable: bool,
     /// Mark this item deprecated in FFI.
     pub deprecated: Option<String>,
     /// An optional namespace. None is equivalent to the root namespace.
@@ -451,15 +453,21 @@ impl Attrs {
     pub fn from_ast(
         ast: &ast::Attrs,
         validator: &(impl AttributeValidator + ?Sized),
-        parent_attrs: &Attrs,
+        parent_attrs: Attrs,
         errors: &mut ErrorStore,
     ) -> Self {
-        let mut this = parent_attrs.clone();
+        let mut this = parent_attrs;
         // Backends must support this since it applies to the macro/C code.
         // No special inheritance, was already appropriately inherited in AST
         this.abi_rename = ast.abi_rename.clone();
 
         this.deprecated = ast.deprecated.clone();
+
+        this.unstable |= ast.cfg.iter().any(|x| {
+            x.to_token_stream()
+                .to_string()
+                .contains("feature = \"unstable\"")
+        });
 
         let support = validator.attrs_supported();
         let backend = validator.primary_name();
@@ -772,6 +780,7 @@ impl Attrs {
         // use an exhaustive destructure so new attributes are handled
         let Attrs {
             disable,
+            unstable: _unstable,
             deprecated: _deprecated,
             namespace,
             rename,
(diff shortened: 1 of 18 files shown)


Manishearth (changes requested):
This is currently done in a very ICU4X-specific way.

I think the general feature here is having a general config option for features that we care about, mapping feature names to the doc string, and then reading from the config during HIR lowering.

This would probably be stuffed into [LoweringConfig](https://docs.rs/diplomat_core/latest/diplomat_core/hir/struct.LoweringConfig.html), populated from the tool's [SharedConfig](https://docs.rs/diplomat-tool/latest/diplomat_tool/config/struct.SharedConfig.html)

robertbastian:
`unstable` is not ICU4X specific. It's the one and only feature name considered unstable by `cargo-semver-checks`

Manishearth:
CSC has a whole bunch of heuristics for unstableness, this is one of them. This is a case of CSC following what it has seen in the ecosystem, that does not make it an ecosystem norm.

In fact a lot of crates just have a feature that is documented as unstable (proc-macro2 used to do this), and CSC's recommendation there is to explicitly test the right feature set.

robertbastian:
What you're suggesting is a lot more work, which I will not do. This is the solution for ICU4X, which is what I'm being paid for.

robertbastian:
This is as far as I will take this

Manishearth:
Tyler, what do you think of this approach?

Manishearth:
(Doing this the more generic way is also something I might just do myself, I don't think it's that hard for someone who knows this code)

robertbastian:
I don't understand what you're planning to make more generic. This is not trying to document arbitrary features, but to explicitly handle unstable APIs. For example for Dart it generates an `@experimental` annotation.

tyler-zeromatter:
Will take a look this morning

Manishearth:
> I don't understand what you're planning to make more generic. This is not trying to document arbitrary features, but to explicitly handle unstable APIs. For example for Dart it generates an `@experimental` annotation.

Okay, I hadn't picked up on that. That's a compelling argument.

In general I would prefer to have far more context on changes like this so that I don't need to review code to understand the intent.

We _generally_ try to make things generic to be more broadly useful: I think this has paid off good dividends as we've had backends benefit from each other even if the original benefit was a smaller scope.

robertbastian:
https://github.com/unicode-org/icu4x/pull/8462/changes/5e6940bce0a3bae435baaa8df7ee2cdee8cc2c59

core/src/hir/attrs.rs
```diff
@@ -1722,6 +1732,13 @@ impl AttributeValidator for BasicAttributeValidator {
     fn attrs_supported(&self) -> BackendAttrSupport {
         self.support
     }
+    fn is_unstable(&self, cfg: &syn::Attribute) -> bool {
+        self.semver_unstable_features.iter().any(|f| {
+            cfg.to_token_stream()
+                .to_string()
+                .contains(&format!("feature = {f:?}"))
```
Manishearth, line 1740: please file a followup to improve this heuristic, this is very brittle (relies on code being formatted a certain way)

core/src/hir/attrs.rs
```diff
@@ -1722,6 +1732,13 @@ impl AttributeValidator for BasicAttributeValidator {
     fn attrs_supported(&self) -> BackendAttrSupport {
         self.support
     }
+    fn is_unstable(&self, cfg: &syn::Attribute) -> bool {
+        self.semver_unstable_features.iter().any(|f| {
```
Manishearth, line 1736: question: should we default to `"unstable"` if this is empty?

Manishearth (commented):
Yeah, I think we should do this now that it's a backend-specific behavior thing.

core/src/hir/attrs.rs
tyler-zeromatter: Seconded, this shouldn't be too hard to write as a syn parser

core/src/hir/attrs.rs
```diff
@@ -451,16 +453,18 @@ impl Attrs {
     pub fn from_ast(
         ast: &ast::Attrs,
         validator: &(impl AttributeValidator + ?Sized),
-        parent_attrs: &Attrs,
+        parent_attrs: Attrs,
```
tyler-zeromatter, line 456: Just out of curiosity: why was this changed to take ownership of the parent's attributes instead of a reference?

tool/src/config.rs
```diff
@@ -27,6 +27,8 @@ pub struct SharedConfig {
     pub custom_extra_code_location: PathBuf,
     /// List of features to enable/disable generation for.
     pub features_enabled: HashSet<String>,
+    /// List of unstable features.
+    pub semver_unstable_features: HashSet<String>,
```
tyler-zeromatter, line 31: Question: why is this called `semver_unstable_features`?

Can a user mark a feature from `features_enabled` as `unstable` for any reason other than that it might change across semantic versions?

tyler-zeromatter (commented):
This mostly looks good to me, with some small comments.

I'm mostly having to pick up what this PR is trying to accomplish through context clues, so a lot of my comments are merely clarifying questions.

Additionally, I'd love to see an example of this working as intended through `feature_tests`, just to test unstable feature configuration and codegen.

core/src/hir/attrs.rs
robertbastian: because it's cloned in the first line anyway, and the caller in `type_context.rs` already needs to clone in order to do `.unwrap_or_default()`, so it shouldn't borrow and clone again.

core/src/hir/attrs.rs
robertbastian: this does not rely on code formatting. `TokenStream::to_string` returns `feature = "foo"`, no matter if the input is `feature ="foo"`, `feature="foo"`, or `feature      = "foo"`.

an actual syn parser would be complex, because it would have to parse a whole _tree_ of `all/any/not` statements and then build and reduce a truth table over those (and the parent's) to figure out whether !unstable actually disables the item. checking whether an unstable feature is mentioned at all is a good enough heuristic.

tool/src/config.rs
robertbastian: > Question: why is this called `semver_unstable_features`?

it's the list of Cargo features that are semver-exempt. that's fairly common in Rust crates, including ICU4X. Any API that requires such a feature will be marked unstable in the generated bindings.

> Can a user mark a feature from `features_enabled` as `unstable` for any reason other than that it might change across semantic versions?

I don't really understand this question. Can this be used for a reason other than the reason it's for?

core/src/hir/attrs.rs
robertbastian: I thought you were arguing that `unstable` is not an ecosystem norm?

robertbastian:
> Additionally, I'd love to see an example of this working as intended through feature_tests, just to test unstable feature configuration and codegen.

I did not want to go through the hassle of actually introducing features to the test crates. See https://github.com/unicode-org/icu4x/commit/5e6940bce0a3bae435baaa8df7ee2cdee8cc2c59 for results.

tool/src/config.rs
tyler-zeromatter: I was mostly asking if "unstable" could mean something else other than the context defined here. It sounds like it's not, so I'll mark this as resolved.

tyler-zeromatter:
We do already have some features for testing: `feature=some_feature` is introduced in `feature_tests/src/lib.rs:35`. I'm a little worried about feature tests existing outside of the Diplomat crate, because that usually means it takes a little longer to catch any issues with codegen.

tyler-zeromatter (approved):
Generally LGTM, I think this is a good feature to add. I only have small nits in the existing comments above, but those don't need to be addressed for this PR to merge

robertbastian:
> We do already have some features for testing: `feature=some_feature` is introduced in `feature_tests/src/lib.rs:35`.

But those are not cargo features, they're diplomat features that need to be kept in sync with cargo features (seems error prone).

> I'm a little worried about feature tests existing outside of the Diplomat crate, because that usually means it takes a little longer to catch any issues with codegen.

This is an opt-in feature which I've implemented for Kotlin, JS, Dart, C++, because we use those languages in ICU4X.

core/src/hir/attrs.rs
Manishearth: I think having a default is fine, but on second thought yeah we shouldn't.

core/src/hir/attrs.rs
Manishearth: Right, can you file a low priority followup about that? I don't actually think we need to parse the whole tree but we can parse a single layer of `all` and not accidentally parse `not`.

robertbastian: ok

PR #1278, opened 2026-09-16
dotnet: replace lifetime reference counting with managed edges and borrow rules

irvingoujAtDevolution:
Follow-up to #1267 after the Zulip discussion with @Manishearth: this drops the .NET-specific reference counting and uses the same borrowing model as the other GC backends, with .NET-specific tweaks on top.

What changes:

- A dependent holds an ordinary managed reference to its source handle, like Kotlin/JS/Dart. No native refcount, no `SafeHandle` claims.
- Anything borrowed from a shared source (`&'a T`, `&'a [u8]`, or an owned `Box<T<'a>>`) is a read view: it records the source's mutation version, the source can still be mutated, and the stale view throws on next use. Values born from `&mut` keep the borrow and must be `manually_disposable` (generator error otherwise), so `using` ends it.
- Borrowing from a `manually_disposable` type is a generator error, so `Dispose()` never has to invalidate children.
- Methods that pin a managed slice for an owned `T<'a>` say so in their generated doc comment.
- Duplicate disposed checks removed, `GcRaceTests` GC-pressure repro restored, new fixtures/tests for the above.

Still on this branch as a known limit: destructors that read a borrowed parent are a binding-author rule, not enforced. Cross-thread call/dispose races are the caller's problem, same as the other backends.

Draft because it's based on `6610ee15` and needs a rebase onto current `main` (RustHandle was rewritten in #1257 since). Opening it now so the design can be looked at first.

Tests: `cargo test -p diplomat-tool` (105), `dotnet test` feature tests 120/120 on net8.0 and net48.


irvingoujAtDevolution:
Hello @Manishearth and @CBenoit


This PR should be mutually exclusive from https://github.com/rust-diplomat/diplomat/pull/1267
The approach is different and the pholosophy + trade offs should be discussed.

After careful consideration, and thinking + experimenting it in our codebase that depends on Diplomat. I'll take what Manish suggested on Zulip with a few future improvements required.

## The decisions and why:

Manish was right, it is too much to ask an FFI library to keep a child usable after its parent is disposed. Instead of delaying the parent's cleanup to support this, we reject APIs that return values borrowing from a `manually_disposable` opaque.


The behavior:
Previously I tried to support the following behavior:
See https://github.com/rust-diplomat/diplomat/pull/1267#issuecomment-5497024062

```csharp
public class Program
{

    public static Child CreateChild()
    {
        using var parent = Parent.New();
        return parent.CreateChild();
    }

    public static void Main()
    {
        var child = CreateChild();
        Console.WriteLine(child);
    }
}
```

Now, instead of supporting it, we will ban all `manually_disposable` opaque types from creating a child reference to itself. That is, anything in the following shape cannot be created:

```rust
#[diplomat::opaque]
#[diplomat::attr(dotnet, manually_disposable)]
pub struct Parent {
    // ...
}

pub struct Child<'a> {
    parent: &'a Parent,
    // ...
}

impl Parent {
    pub fn create_child<'a>(&'a self) -> Child<'a> {
        Child { parent: self }
    }
}
```

in a nutshell, if we want deterministic disposal, then we cannot return a child that borrows from it.

Side Note: Mutability garentees stays the same as in the other PR.

## The non-deterministic finalization problem

As previously mentioned in the comment in the other PR, we cannot guarantee that the child is dropped before the parent when both become unreachable. This is a existing problem that was not discussed before for other backend as well.
As @Manishearth mentioned in person, I think we should open a follow up PR for non-deterministic GC languages to ban custom `Drop` implementations for any opaque types, or at least any opaque types that borrows from other opaque types. I would love to hear more opinions on this. For now, destructors must not read borrowed parent memory. This includes field destructors, so avoiding a custom `Drop` on the opaque itself is not enough.

Manishearth (approved):

Manishearth:
As usual I have skimmed the approach and it seems to mostly be in line with what I was thinking. Thanks!

PR #1279, opened 2026-09-17
dotnet: invalidate readers during exclusive borrows (Follow up to #1278)

irvingoujAtDevolution:
Follow-up to [#1278](https://github.com/rust-diplomat/diplomat/pull/1278). A versioned read view now fails immediately while its source has an active exclusive borrow. Before this, it stayed usable until the writer was disposed and advanced the mutation version.

Tests: `cargo test -p diplomat-tool`; focused .NET regression on net8.0 and net48.

---

> **LLM Wrote, Unaudited — Read With a Grain of Salt.**


Manishearth (approved):

irvingoujAtDevolution:
@Manishearth need review, just rebased

